PR #324, opened 2024-04-15
Patch SqlDictCache to on-the-fly update cache to new format.

brianwgoldman:
I did some really basic testing and it seems to work.


dhosterman:
Can we go ahead and close this now that we have a working cache file?

wpietri:
Closing, but I'm going to keep the branch for a while, just in case.

PR #330, opened 2024-04-15
Change package name. Also upgrade to modelgauge 0.5.

wpietri:
All the tests pass, it runs, and it look right to me, but I'd appreciate a careful inspection here.


dhosterman (approved):
👍

PR #331, opened 2024-04-16
Update reference SUTs

dhosterman:
With the availability of updated cache files, we can bring the new reference SUTs used in our v0.5 demo into the code that uses Modelgauge instead of NewHELM.


wpietri (approved):
Ah, excellent. I was just thinking about pulling this from the dead branch.

PR #332, opened 2024-04-16
Update to new letter grading scheme

wpietri:


dhosterman (approved):
👍 

In the future, I'd like to see if we can figure a way to keep these letter grade mappings together (there's some in .toml files for the web output too).

PR #333, opened 2024-04-16
Trying other Python versions

wpietri:
Making the PR mainly to trigger workflows.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -18,10 +18,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.10
+    - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v3
       with:
-        python-version: "3.10"
+        python-version: ${{ matrix.python-version }}
         
     - name: cache poetry install
       uses: actions/cache@v3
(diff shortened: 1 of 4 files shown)


wpietri:
Ok, now this actually works. Tests are run in the correct python versions.

.github/workflows/python-app.yml
```diff
@@ -12,20 +12,24 @@ jobs:
...
-        python-version: "3.10"
+        python-version: ${{ matrix.python-version }}
         
     - name: cache poetry install
       uses: actions/cache@v3
       id: cache-poetry
       with:
         path: ~/.local
-        key: poetry-1.8.2-0
+        key: poetry-1.8.2-${{ steps.setup-python.outputs.python-version }}-0
```
yifanmai, line 32: I'd suggest using the [built-in Poetry cache support in `actions/setup-python`](https://github.com/actions/setup-python?tab=readme-ov-file#caching-packages-dependencies) instead, which should be less brittle. That should allow you to drop the `actions/cache` steps entirely.

wpietri: Ooh, neat! Thanks for the tip.

wpietri:
Ok @dhosterman, I think this is at last ready to review. Thanks @yifanmai for the simplification tip.

dhosterman (approved):
👍

PR #334, opened 2024-04-16
Replace Hazard's deprecated config class

bkorycki:
This addresses the pydantic deprecation warning that was appearing when running tests:
`PydanticDeprecatedSince20: Support for class-based config is deprecated, use ConfigDict instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.7/migration/`


wpietri (approved):
Oh, excellent! Thanks for running this down.

dhosterman (approved):
Great, thanks!

PR #338, opened 2024-04-17
Add ability to benchmark dynamic SUT

wpietri:
Now users should be able to run a benchmark for any registered ModelGauge SUT via the `-s` option. Also add some very basic defaulting for SUT static content so that reports will render.

### DIFF
--- a/src/modelbench/static_site_generator.py
+++ b/src/modelbench/static_site_generator.py
@@ -121,7 +121,12 @@ class StaticSiteGenerator:
 
     @content.register
     def content_sut(self, item: SutDescription, key: str):
-        return self._content[item.key][key]
+        if item.key in self._content:
+            return self._content[item.key][key]
+        elif key=="name":
+                return item.display_name
+        else:
+            raise ValueError(f"Unknown SUT key for {item} and {key}")
 
     @content.register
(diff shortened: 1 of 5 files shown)


src/modelbench/static_site_generator.py
```diff
@@ -121,7 +121,12 @@ def content_string(self, item: str, key: str):
 
     @content.register
     def content_sut(self, item: SutDescription, key: str):
-        return self._content[item.key][key]
+        if item.key in self._content:
+            return self._content[item.key][key]
+        elif key == "name":
+            return item.display_name
```
dhosterman, line 127: Since this is all we care about, should we remove the `tagline` from the .toml file now?

dhosterman (approved):
Looks solid to me! We'll need to reevaluate if the SUT .toml file ever becomes more complicated than it is now, but maybe that'll never happen?

src/modelbench/static_site_generator.py
wpietri: Removed!

wpietri:
Yeah, honestly, I think we could drop `suts.toml` entirely, in that the SutDescription has a name field that nobody is messing with. But let's see where things go.

PR #347, opened 2024-04-27
Potential v0.5.1 PyPi release

dhosterman:
A candidate for the v0.5.1 pypi release; works with pythons from 3.10 to 3.12.


wpietri (approved):
This all looks great, but I'm wondering about the intent behind the keywords. Do we want people to find this via pypi?

dhosterman:
I don't see why not, but I guess we can always remove them if there's some reason for folks not to be able to discover it.

PR #354, opened 2024-04-30
Remove unnecessary plugins

dhosterman:
This dramatically reduces the amount of time it takes to run `modelbench --help`. It's about 2x as fast on the first run (from 5.5ish seconds to 2.5ish seconds) and about 3x as fast on subsequent runs (from 1ish second to .33ish seconds).

I suspect installs will be sped up as well.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,87 +2,87 @@
 
 [[package]]
 name = "aiohttp"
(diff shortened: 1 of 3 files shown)


wpietri (approved):
:partying_face:

PR #360, opened 2024-05-07
Output JSON results if --output-json-dir is specified

yifanmai:
This outputs the test JSON if `--output-json-dir` is specified. The output files follow a similar format as modelgauge:

https://github.com/mlcommons/modelgauge/blob/79283fdd83327ee9def045fda0897e34a9cef72b/modelgauge/main.py#L203-L217

Addresses #237


wpietri (commented):
Sorry I missed this until now. It seems odd to me to have a benchmark command output JSON that's not the benchmark. Could you say a bit more about your intent here?

yifanmai:
The result of the offline discussion was that we didn't want to expose the internal JSON schema publicly. So I'll close this in favor of #362.

PR #361, opened 2024-05-10
allow-local-plugin-loading

dhosterman:


wpietri (approved):
This looks great, but I would have expected this option to go on the `cli()` function so that this is available to every command in the group. Is there a reason to restrict it to just some of the commands?

dhosterman:
2 reasons, I guess:

1) I personally hate CLIs that require you to put some options before the command and some after;
2) It's not relevant for all the CLI commands, is it?

bkorycki (approved):

PR #362, opened 2024-05-10
Add --sut flag to grid and responses commands

yifanmai:
This allows running the `grid` and `responses` commands on user-specified SUTs.

Addresses #237


wpietri (approved):
Great idea. Thanks!

PR #370, opened 2024-05-16
new Together client, move SafeTest and LlamaGuardAnnotator from modelbench

wpietri:
This is what it would look like to move the core parts of the benchmark in while using the new Together API. It all works, but I did this cowboy style, so I'd appreciate a careful review to see if this is really the direction we want to go in.

### DIFF
--- a/src/modelbench/modelgauge_runner.py
+++ b/src/modelbench/modelgauge_runner.py
@@ -36,9 +36,17 @@ class ModelGaugeSut(SutDescription, Enum):
     GEMMA_7B = "gemma-7b", "Gemma Instruct (7B)", TogetherChatSUT, "google/gemma-7b-it"
     LLAMA_2_7B = ("llama-2-7b-chat", "Meta Llama 2, 7b parameters", TogetherChatSUT, "meta-llama/Llama-2-7b-chat-hf")
     LLAMA_2_13B = (
-        "llama-2-13b-chat", "Meta Llama 2, 13b parameters", TogetherChatSUT, "meta-llama/Llama-2-13b-chat-hf")
+        "llama-2-13b-chat",
+        "Meta Llama 2, 13b parameters",
+        TogetherChatSUT,
+        "meta-llama/Llama-2-13b-chat-hf",
+    )
     LLAMA_2_70B = (
-        "llama-2-70b-chat", "Meta Llama 2, 70b parameters", TogetherChatSUT, "meta-llama/Llama-2-70b-chat-hf")
+        "llama-2-70b-chat",
+        "Meta Llama 2, 70b parameters",
+        TogetherChatSUT,
+        "meta-llama/Llama-2-70b-chat-hf",
+    )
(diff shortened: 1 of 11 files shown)


dhosterman:
Is there a reason you're choosing not to use ModelGauge plugins here?

src/modelbench/modelgauge_runner.py
```diff
@@ -53,12 +64,11 @@ class ModelGaugeSut(SutDescription, Enum):
     )
     VICUNA_13B = "vicuna-13b", "LM Sys Vicuna v1.5 (13B)", TogetherChatSUT, "lmsys/vicuna-13b-v1.5"
     WIZARDLM_13B = "wizardlm-13b", "WizardLM v1.2 (13B)", TogetherChatSUT, "WizardLM/WizardLM-13B-V1.2"
-    # YI_34B_CHAT = "yi-34b", "01-ai Yi Chat (34B)", TogetherChatSUT, "zero-one-ai/Yi-34B-Chat"
+    # YI_34B_CHAT = "yi-34b", "01-ai Yi Chat (34B)", IntegratedTogetherChatSUT, "zero-one-ai/Yi-34B-Chat"
```
bkorycki, line 67: What/where is `IntegratedTogetherChatSUT`?

bkorycki:
I see the reason to move SafeTest out of ModelGauge and into ModelBench, but I'm not sure I understand why we would move the together client over as well.
Also, ModelGauge is a library with a plugin architecture, so I think it might be easier and make more sense to just make SafeTest a plugin here.

src/modelbench/modelgauge_runner.py
wpietri: Ah, sorry. That is now TogetherChatSUT again. I first tried this under a separate name, as I hadn't yet kicked out the existing plugins from pyproject.toml. I'll fix that.

wpietri:
You both make good points. My real struggle here is that I'm not sure we need modelgauge at all anymore now that it's not a potential HELM replacement. But we should at least approach it piecemeal. So I'll try doing this as a PR for the together stuff in modelgauge, and I'll move SafeTest here.

For the together stuff, you'll note that I've thrown out a fair bit of the existing code. Do we care? This is all modelbench needs to run, but if there are other use cases I'm happy to put it back.

bkorycki:
I definitely have some thoughts on throwing out so much from Together. For instance, I think we should still keep `TogetherInferenceSUT` for models that aren't geared towards chat. Also the changes to the Request classs make for a less robust cache.
But overall, I think we might want to hold off on making these kinds of decisions until we have another higher-level discussion about merging modelgauge/bench that answers the following questions:
1. Exactly what use cases we are going to support/drop?
    E.g. How much do we cater to test creators vs. SUT developers?
2. What parts of the functionality in ModelGauge need to be modified to support our new goals? 
     OR are too many of the foundational assumptions that drove the design of ModelGauge no longer relevant? Which would warrant a complete overhaul of the design.

And then we will have more clarity into these kinds of lower-level decisions that I believe we may be trying to tackle here.

wpietri:
Stale, let's take a new direction next time

wpietri:
This PR is dead to me. DEAD I SAY!

PR #377, opened 2024-05-24
Generic reports

bkorycki:
A possibly wild stab at generating generic reports. content/ now contains all of the generic/default .toml files, and values in the new content_mlc/ directory will override the defaults if `--generic-report` is not provided.
Just want to confirm that this is generally how we want generic reports to work, and then I’ll add the CSS stuff.


src/modelbench/run.py
```diff
@@ -76,26 +76,33 @@ def cli() -> None:
     multiple=True,
 )
 @click.option("--view-embed", default=False, is_flag=True, help="Render the HTML to be embedded in another view")
-@click.option("--anonymize", type=int, help="Random number seed for consistent anonymization of SUTs")
-@click.option("--parallel", default=False, help="experimentally run SUTs in parallel")
+@click.option(
+    "--custom-branding",
+    type=click.Path(file_okay=False, dir_okay=True, path_type=pathlib.Path),
```
dhosterman, line 81: You might want `exists=True` here.

src/modelbench/run.py
```diff
@@ -76,26 +76,33 @@ def cli() -> None:
...
+    help="Path to custom branding. Implicitly sets --generic-report to remove MLCommons copy.",
+)
 @click.option(
```
dhosterman, line 84: Would it make more sense here to just have one CLI option that is either 1) `--custom-branding` which includes MLC branded stuff or 2) no `--custom-branding` which is always generic?

wpietri (commented):
Yes, I like where you're going with this, but agree with Daniel. The default should be a generic report. Then something like a --branding option lets us point to something like a folder that contains our flavor packet. In a future pull request, we'll move that folder-or-whatever to ModelLab, but this PR just needs to start extracting it.

src/modelbench/run.py
bkorycki: Wouldn't users need to know the path of MLC branding in that case?

wpietri: I think for now the only users running reports with MLC branding will be the three of us. And we'll eventually put that branding directory in `modellab`, where we'll be running the official reports from.

bkorycki:
Okay, updated to remove `--generic` and make generic default. To get a report with MLC branding, you now need to use `--custom-branding src/modelbench/templates/content_mlc` (if you're at the root level)

dhosterman (approved):
Looks good to me! Sorry it took so long for me to review!

PR #383, opened 2024-06-12
better observability tooling for big runs (issue #343)

FelixMo42:
This PR uses rich to add progress bars and manage logging for runs.

Debug and error messages are logged behind the bars with a timestamp. I removed the other non debug logs since I'm not sure if they are still wanted if there progress bars are there, but they could easily be added back in.

## Limitations / Follow ups

- Progress bars are only at a per test resolution. To get a per test item resolution some way to get that info back from modelgauge would be needed.
- Errors along with their tracebacks are printed, buts it's a little verbose and it dosen't advise any way of solving them. I don't know enough about common errors and their solutions to implement this at the moment, but it should be pretty easy.
- The ways it's coded doesn't make it very easy to re use if we want similar, but not the same, progress bars somewhere else.
- ~~When a sut errors it puts the bar at 100%, but it remains green and doesn't indicate that it failed other than logging the error message. I couldn't find a default way to do this in rich, but it shouldn't be too hard to add, the library seems super extensible.~~
- In parallel mode there was some flickering I couldn't figure out how to get ride of.

## Example

Screenshot 2024-06-12 at 9 10 25 AM
Screenshot 2024-06-12 at 9 16 36 AM


bkorycki:
Overall this looks good! My only suggestion is to put `ProgressBars` somewhere else (maybe utilities?) to avoid cluttering up run.py. Also I definitely would like to see a fix to the progress bar displaying 100% in green when a SUT errors. But other than that, good job!

FelixMo42:
I moved it into utilities.py

FelixMo42:
Screenshot 2024-06-12 at 12 40 46 PM
(generated using a forced error)

dhosterman:
I like the addition of the better error handling in the output! Very nice!

PR #386, opened 2024-06-27
Add HasUid and apply it to Benchmark and Hazard.

wpietri:
Machinery for consistent UIDs. See the uids.py file and associated tests for documentation on how it works.

### DIFF
--- a/src/modelbench/benchmarks.py
+++ b/src/modelbench/benchmarks.py
@@ -46,7 +46,7 @@ class BenchmarkScore(ABC, LetterGradeMixin):
 
 class BenchmarkDefinition(ABC, HasUid):
     _hazards: list[HazardDefinition]
-    _uid_definition = {"class":"self", "version": "0.5"}
+    _uid_definition = {"class": "self", "version": "0.5"}
 
     def __init__(self, hazards: List[HazardDefinition]):
         super().__init__()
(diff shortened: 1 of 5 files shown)


dhosterman (approved):
Looks great!

bkorycki:
Nice! A couple of thoughts:

- I think there needs to be some mechanism that checks that a UID is not used by any other object. 
- Also, it might be worth considering using something more structured than a dictionary to specify the UID. It feels unintuitive to provide keys that are unused for extra values. Also the ordering of keys affect the UID string.

wpietri:
Sorry, somehow forgot to merge this. To Barbara's points:


> * I think there needs to be some mechanism that checks that a UID is not used by any other object.

I think that would be cool, but for a given runtime, we can only know about instantiated objects, which is a small fraction of all the objects. Long term, we could put something in ModelLab that would guarantee uniqueness, although at first blush I'm not sure how we'd distinguish between true uses and false ones.


> * Also, it might be worth considering using something more structured than a dictionary to specify the UID. It feels unintuitive to provide keys that are unused for extra values. Also the ordering of keys affect the UID string.

I was thinking of the keys as having documentary value, and was intending that the ordering of the keys define the ordering of the value. I was also thinking that the structure might  later on in building a parser. Happy to look at something besides a dict as we work on this, though.

PR #392, opened 2024-07-16
Benchmark outcomes record

wpietri:
Produces a JSON version of the benchmark alongside the HTML files. Not sure this is totally right; Looking forward to feedback on the format.


wpietri:
Ok, @bkorycki and @dhosterman, this is actually ready for final review now.

src/modelbench/hazards.py
```diff
@@ -12,9 +12,14 @@
 from pydantic import BaseModel, ConfigDict
 
 from modelbench.scoring import BAND_CONSTANTS, LetterGradeMixin, NumericGradeMixin, ValueEstimate
+from modelbench.uid import HasUid
 
 
-class HazardDefinition(ABC):
+class HazardDefinition(ABC, HasUid):
+
+    def __init__(self):
```
bkorycki, line 20: Given the structural similarities between hazards and benchmarks, would it make sense to be somewhat consistent in their class definitions? E.g. Make `_tests` a class variable which is set using a `tests` parameter in `__init__`?   Or conform `BenchmarkDefinition` to the current `HazardDefinition`.

src/modelbench/run.py
```diff
@@ -101,10 +103,13 @@ def benchmark(
     anonymize=None,
     parallel=False,
 ) -> None:
+    start_time = datetime.now(timezone.utc)
     suts = find_suts_for_sut_argument(sut)
-    benchmarks = [b() for b in BenchmarkDefinition.__subclasses__() if b.__name__ in benchmark]
-    benchmark_scores = score_benchmarks(benchmarks, suts, max_instances, debug, parallel)
+    benchmark = [b() for b in BenchmarkDefinition.__subclasses__() if b.__name__ == benchmark_name][0]
+    benchmark_scores = score_benchmarks([benchmark], suts, max_instances, debug, parallel)
```
bkorycki, line 109: If we anticipate only one benchmark to be run at a time, maybe we can also simplify `score_benchmarks` and `score_a_sut` as well.

src/modelbench/run.py
```diff
@@ -101,10 +103,13 @@ def benchmark(
...
     generate_content(benchmark_scores, output_dir, anonymize, view_embed, custom_branding)
+    json_path = output_dir / f"benchmark_record-{benchmark.uid}.json"
```
bkorycki, line 111: Might be useful to provide an optional `--output-file` command option. In case a user wants to re-run a benchmark  without overwriting the last file.

src/modelbench/run.py
```diff
@@ -101,10 +103,13 @@ def benchmark(
...
+    dump_json(json_path, start_time, benchmark, benchmark_scores)
```
bkorycki, line 112: I am perplexed! How does this work without importing the method from `modelbench.record`?

src/modelbench/modelgauge_runner.py
```diff
@@ -55,6 +55,18 @@ class ModelGaugeSut(SutDescription, Enum):
     WIZARDLM_13B = "wizardlm-13b", "WizardLM v1.2 (13B)", TogetherChatSUT, "WizardLM/WizardLM-13B-V1.2"
     # YI_34B_CHAT = "yi-34b", "01-ai Yi Chat (34B)", TogetherChatSUT, "zero-one-ai/Yi-34B-Chat"
 
+    def instance(self, secrets):
```
bkorycki, line 58: Why do we need these methods?

src/modelbench/uid.py
```diff
@@ -0,0 +1,64 @@
+import re
+
+import casefy
+
+
+class HasUid:
```
bkorycki, line 6: I left a few comments about this in [the other PR](https://github.com/mlcommons/modelbench/pull/386#issuecomment-2197726110)!

src/modelbench/record.py
```diff
@@ -0,0 +1,102 @@
...
+            return {"uid": o.uid, "tests": o._tests}
+        elif isinstance(o, SafeTest):
+            return o.uid
+        elif isinstance(o, ModelGaugeSut):
+            result = {"uid": o.key}
+            if o.instance_initialization():
+                result["initialization"] = o.instance_initialization()
+            return result
+        elif isinstance(o, pydantic.BaseModel):
+            return o.__dict__
```
bkorycki, line 98: I think `o.model_dump()` is the preferred method for converting to a dictionary

bkorycki (commented):
Nice work!

src/modelbench/run.py
wpietri: I did look at that, but it involved changing a lot of code, so I thought I'd leave that for a future cleanup.

src/modelbench/run.py
wpietri: Interesting. I wanted it to be in a well-known location, so that other tools, like modellab, could pick it up. I was thinking that if they wanted to preserve an old run, they'd use the `--output-dir` argument to preserve the whole thing.

src/modelbench/run.py
wpietri: Hah! That's because it doesn't work. I must have messed up the imports at the last minute, and that is not tested. I'll fix it. Thanks!

src/modelbench/run.py
bkorycki: Ah I forgot about `--output-dir`. Maybe that's good enough.

src/modelbench/run.py
bkorycki: Fair enough

src/modelbench/run.py
wpietri: Fixed, and test added.

src/modelbench/modelgauge_runner.py
wpietri: Moving instance creation here will let me unify duplicate code, and it gives me a place to cache the instance actually used for the run, which is needed to dump out the outcome JSON.

src/modelbench/record.py
wpietri: Ooh, thanks! I'll switch it.

src/modelbench/hazards.py
wpietri: It would, but the way I'd make them parallel is by making it so that you don't need to know a Secret just to instantiate a Test. We really only need the Secret if a) we're doing a run, and b) the run isn't already fully cached. So I think the secrets are perhaps better passed in not at instantiation, but as part of the run. But I didn't want to redo all that just for this, so I went for the minimum higher distortion from the issues lower down.

wpietri:
Ok @dhosterman and @bkorycki, I think I have resolved all the outstanding issues and requests on this one.

src/modelbench/hazards.py
bkorycki: That makes sense!

bkorycki (approved):

dhosterman:
This works great so far, but it fails when attempting to use --anonymize.

dhosterman:
I also notice that in the `content` data, we have a `uid` for the benchmark that is different than the `uid` in the `benchmark` data. We might want to go through and make sure that those things are aligned, as well as the versions.

dhosterman (approved):
Looks great and I'm already using it! Thanks, William!

Issue #398, opened 2024-07-31
Add skeleton 1.0 benchmark


bkorycki:
The skeleton should be a basic minimum structure that lets us run something. It should incorporate whatever we can easily get from current workstream outputs, and provide a solid structure for adding things on. Along the way, let's make a list of what we'll need to get to a proper 1.0. 

Here's about where we'll start:
- at least 3 prompts
  - synthetic prompts from workstream 3?
  - not ground truth prompts
- at least one hazard from workstream 1's definitions
- 1 test per hazard
- llama guard 2 to start (or whatever's coming from workstream 2)
- hazard score is fraction unsafe
- personas are all combined
- benchmark scoring: use same reference models and approach as in 0.5, but in separate code
- benchmarks are separated by language and persona? Start with english and normal-ish persona

This from WS1 looks helpful: https://docs.google.com/document/d/1JUQyd4UvpXlduqaMyIdCnyGF9uleu8ZyB07PX9ZUB-k/edit#heading=h.2ua223rrefyo

PR #399, opened 2024-08-01
Skeleton benchmark 1.0

bkorycki:
The primary difference between 0.5 and 1.0 seems to be the inclusion of additional languages. WG1 says scores from different languages should not be aggregated, so I envision each language to be it's own benchmark. This will require some re-factoring of modelgauge hazards as well.


wpietri (changes requested):
As we discussed in the standup, let's go for more with this. In particular, I'm hoping to see most or all of the bullet points in the issue: https://github.com/mlcommons/modelbench/issues/398

wpietri:
Oh, sorry, I didn't realize you had already bumped the modelgauge version in here when I started in on a PR for that. Let's get this merged and then maybe drop my PR if it's duplicative.

wpietri:
Before I dive in to review, could you say how much of #398's bullet points are in this PR?

bkorycki:
@wpietri
> Before I dive in to review, could you say how much of #398's bullet points are in this PR?

- at least 3 prompts ✅
  - synthetic prompts from workstream 3? ❌ (they are fully fake)
  - not ground truth prompts ✅
- at least one hazard from workstream 1's definitions ✅ (dfm-- defamation)
- 1 test per hazard ✅
- llama guard 2 to start ✅
- hazard score is fraction unsafe  ✅
- personas are all combined  ✅
- benchmark scoring: use same reference models and approach as in 0.5, but in separate code  ✅
- benchmarks are separated by language✅ and persona❌? Start with english and normal-ish persona
  - Modelgauge separates tests by language. So far there is just one for English.
  - The personas are grouped together right now because there's some conflicting info regarding this point

wpietri (approved):
Looks great! Thanks for going the distance.

PR #403, opened 2024-08-08
Update to latest dependencies

wpietri:


wpietri:
> I think the together extra for modelgauge also needs to be removed from pyproject.toml
Great point. Removed. Are the other extras perhaps outdated as well?

PR #420, opened 2024-08-27
Always print progress bar

bkorycki:
Output with `--machine-reports --parallel true` flags from before:
Screenshot 2024-08-27 at 12 23 27 PM

After this change:
Screenshot 2024-08-27 at 12 24 20 PM

Let me know if this is what we want.

### DIFF
--- a/src/modelbench/run.py
+++ b/src/modelbench/run.py
@@ -181,7 +181,7 @@ def score_a_sut(benchmarks, max_instances, secrets, progress, debug, sut):
                     sut=sut_instance,
                     data_dir="./run",
                     max_test_items=items,
-                    disable_progress_bar=progress.print_updates,   # Proxy for machine-readable logging.
+                    disable_progress_bar=progress.print_updates,  # Proxy for machine-readable logging.
                 )
                 progress.increment()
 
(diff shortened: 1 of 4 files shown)


bkorycki:
Okay I made quite a few changes @wpietri @dhosterman 
- Renamed `--machine-reports` to `--json-logs`
- Regular logging is always printed
- CLI progress bars are disabled with `--json-logs`
- Disable modelgauge plugin loading progress bar in cli
- Print initial progress json at beginning of execution.
- Reverted to using `echo` instead of `logging.info`. I originally switching over to logging in the first progress reports PR in order to handle different log levels for machine vs. human callers. This is not relevant now that log statements always need to be printed. Also `logging.info` did not print to stdout in parallel mode.

wpietri (approved):
Sounds great! Thanks for all the changes.

dhosterman:
I just tested this against the latest changes in modellab and it works great! It's still showing progress bars for the dependency downloads, but that'll require downstream work in modelgauge to fix and doesn't impede using the progress data provided.

rogthefrog (approved):
👍🏻

PR #421, opened 2024-08-27
output dict in json format for consumption

rogthefrog:
The progress tracker optionally sends its progress as a dict to stdout.  Sugar consumes that and does string manipulation to change the quotes to make it parseable as json. This ensures the dict is printed out in a format that is json parseable as-is, without string manipulations. 

Addresses [this comment](https://github.com/mlcommons/sugar/pull/32/files#r1733406407)

### DIFF
--- a/src/modelbench/utilities.py
+++ b/src/modelbench/utilities.py
@@ -1,4 +1,5 @@
 import ctypes
+import json
 from collections import defaultdict
 from multiprocessing import Value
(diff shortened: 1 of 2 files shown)


wpietri (approved):
💯

bkorycki (approved):

dhosterman (approved):
👍

PR #435, opened 2024-09-05
public-and-private-images

dhosterman:


wpietri (approved):
Looks good to me. And if this is the thing you think was spamming, I didn't get any notifications.

dhosterman:
@wpietri , oh strange! I just assumed everyone would get the failing job messages. That's great to know, thanks!

PR #436, opened 2024-09-05
Use modelgauge v0.6.1

wpietri:


wpietri:
Oops, I must have branched from a branch, not from main. I'll redo.

PR #439, opened 2024-09-06
Smoke test includes v1 benchmark

wpietri:


dhosterman:
@wpietri , usually I'll just add `push:` to the `on:` section of the configuration while I'm testing it and then remove it when I'm done. Would that work here?

.github/workflows/scheduled-smoke-test.yml
```diff
@@ -4,6 +4,7 @@ on:
   schedule:
     # Every day at 08:00 UTC
     - cron:  '0 8 * * *'
+  workflow_dispatch:
```
rogthefrog, line 7: 👍🏻

rogthefrog (approved):

dhosterman (approved):
👍

PR #441, opened 2024-09-06
fully parallel benchmark runner

wpietri:
The code in benchmark_runner should be a full replacement for the simple_test_runner in modelgauge. If folks like this PR, then my next step is to put it to use in run.py.

Not sure how I messed up the rebase but many of the commits below are duplicates. Sorry about that!


src/modelbench/benchmark_runner.py
```diff
@@ -0,0 +1,397 @@
...
+        self.tests.append(test)
+
+    def _check_ready_to_run(self):
+        super()._check_ready_to_run()
+        if not self.tests:
+            raise ValueError("must call add_test() at least once")
+
+    def run(self, run_object=None) -> TestRun:
+        self._check_ready_to_run()
+        run = run_object or TestRun(self)
```
rogthefrog, line 328: This is just a personal preference: usually I try to avoid naming a variable the same as an enclosing function, especially in python or where the symbol is used in different places with different type (e.g. `run()` the bound method on `pipeline`, `run` the method of this object, `run` the run object, etc.

tests/test_benchmark_runner.py
```diff
@@ -0,0 +1,320 @@
...
+class ATest(PromptResponseTest):
+
+    def __init__(self, uid: str, items, secrets, annotator=None):
+        super().__init__(uid)
+        self.items = items
+        self.secrets = secrets
+        self.initialization_record = InitializationRecord(module="testing", class_name="a_test", args=[], kwargs={})
+        self.annotator = annotator or ANNOTATORS.make_instance("demo_annotator", secrets=self.secrets)
+
+    def get_dependencies(self) -> Mapping[str, ExternalData]:
```
rogthefrog, line 47: This function claims to return something, but doesn't. Is it overridden in a subclass?

tests/test_benchmark_runner.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]:
+        return self.items
+
+    def get_annotators(self) -> Mapping[str, Annotator]:
+        return {"demo_annotator": self.annotator}
+
+    def measure_quality(self, item: TestItemAnnotations) -> Dict[str, float]:
+        ann = item.interactions[0].response.completions[0].get_annotation("demo_annotator", DemoYBadAnnotation)
+        print(f"ann: {ann}")
```
rogthefrog, line 58: Do we want `print` or logging here? (I don't know)

tests/test_benchmark_runner.py
```diff
@@ -0,0 +1,320 @@
...
+        return ATest("a_test", [item_from_test], fake_secrets)
+
+    @pytest.fixture()
+    def a_sut(self):
+        return ModelGaugeSut("demo_yes_no")
+
+    @pytest.fixture()
+    def exploding_sut(self, a_sut):
+        real_sut = MagicMock()
+        real_sut.evaluate.side_effect = ValueError("sut done broke")
```
rogthefrog, line 90: 🥇

tests/test_benchmark_runner.py
```diff
@@ -0,0 +1,320 @@
...
+
+            def score(self, sut_scores: Mapping[str, TestRecord]) -> HazardScore:
+                count = 0
+                total = 0
+                for key, value in sut_scores.items():
+                    d = value.result.data
+                    count += d["badness_count"]
+                    total += d["total_badness"]
+
+                score = ValueEstimate.make(total / count, count)
```
rogthefrog, line 119: This will barf if sut_scores is empty. Should we explicitly guard against `count==0` and throw an exception instead of a runtime barfage? It'd be a degenerate result if `sut_scores` is empty, of course, but should we consider handling it less violently?

tests/test_benchmark_runner.py
```diff
@@ -0,0 +1,320 @@
...
+
+    @pytest.fixture()
+    def hazard(self):
+        pass
+
+    def test_benchmark_source(self, fake_secrets, tmp_path, benchmark):
+        bsa = TestRunItemSource(self.a_run(tmp_path, secrets=fake_secrets, max_items=1, benchmarks=[benchmark]))
+        iterator = iter(bsa.new_item_iterable())
+        first_item = next(iterator)
+        print(first_item)
```
rogthefrog, line 150: `pytest` will eat anything that gets `print`ed. Should we use logging here instead?

tests/test_benchmark_runner.py
```diff
@@ -0,0 +1,320 @@
...
+        runner.add_benchmark(benchmark)
+        sut = ModelGaugeSut("demo_yes_no")
+        runner.add_sut(sut)
+        runner.max_items = 1
+        run_result = runner.run()
+
+        assert run_result.benchmark_scores
+        assert run_result.benchmark_scores[benchmark][sut]
+
+    def test_test_runner_has_standards(self, tmp_path, a_test, fake_secrets):
```
rogthefrog, line 264: One of the best function names I've ever read.

rogthefrog (commented):
This is nice and clear. I don't have enough domain knowledge yet to say anything of substance beyond a few comments, unfortunately, but generally this looks solid to me.

Related: https://github.com/python/cpython/pull/116338

Have we tried running the pipeline runner with the GIL disabled?

src/modelbench/benchmark_runner.py
```diff
@@ -0,0 +1,397 @@
...
+        return self.actual_test.initialization_record
+
+
+@dataclass
+class TestRunItem:
+    """The data related to running a single test item"""
+
+    test: ModelgaugeTestWrapper
+    test_item: TestItem
+    sut: ModelGaugeSut = None
```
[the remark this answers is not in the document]
wpietri: It's a good question. My general rule for acronyms in class names is to case them as words. ModelGauge and ModelBench were kinda separate efforts, so they ended up with somewhat different standards. This sounds like a good thing to discuss in a retro.

src/modelbench/benchmark_runner.py
```diff
@@ -0,0 +1,397 @@
...
+            prompt=item.test_item.prompts[0],
+            response=SUTResponseAnnotations(completions=[annotations]),
+        )
+        measurement = self.actual_test.measure_quality(TestItemAnnotations(test_item=item.test_item, interactions=[a]))
+        item.add_measurement(measurement)
+
+    def aggregate_measurements(self, items: List["TestRunItem"]):
+        mtis = []
+        for i in items:
+            measurements = i.measurements
```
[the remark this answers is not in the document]
wpietri: Well spotted. Fixed.

src/modelbench/benchmark_runner.py
```diff
@@ -0,0 +1,397 @@
...
+    def add_finished_item(self, item: "TestRunItem"):
+        if item.completion() and item.annotations and not item.exception:
+            self.finished_items[item.sut.key][item.test.uid].append(item)
+        else:
+            self.failed_items[item.sut.key][item.test.uid].append(item)
+
+    def add_test_record(self, test_record: TestRecord):
+        self.test_records[test_record.test_uid][test_record.sut_uid] = test_record
+
+    def finished_items_for(self, sut, test):
```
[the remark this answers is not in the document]
wpietri: Good idea. Done.

src/modelbench/benchmark_runner.py
```diff
@@ -0,0 +1,397 @@
...
+        super().__init__(runner)
+        self.benchmarks = runner.benchmarks
+        self.benchmark_scores = defaultdict(dict)
+
+        for b in self.benchmarks:
+            for h in b.hazards():
+                for t in h.tests(self.secrets):
+                    self.add_test(t)
+
+    def benchmark_scores(self) -> Mapping[BenchmarkDefinition, Mapping[ModelGaugeSut, BenchmarkScore]]:
```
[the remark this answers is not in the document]
wpietri: Ah, just a leftover. I may end up with some access methods as I wire this into the CLI, but it's sure not helping now, so I've removed it.

src/modelbench/benchmark_runner.py
wpietri: Fair enough. I'll rename it to test_run.

tests/test_benchmark_runner.py
wpietri: It's just there to comply with the superclass contract. I've renamed the class to AFakeTest to make it clearer that this is just a test dummy.

tests/test_benchmark_runner.py
wpietri: Just a debugging leftover. Removed!

tests/test_benchmark_runner.py
wpietri: Leftover debugging code. Removed!

rogthefrog (approved):
👍🏻

dhosterman (approved):
This looks great, William! I'm excited about how nice the interface is and also by the speeds you were reporting!

bkorycki (commented):
Thanks for putting so much work into this! 
I do find the object dependency graph a bit complicated/confusing. For example, `BenchmarkRunner` inherits attribute `tests` which is never used,  but `BenchmarkRun` on the other hand *does* use its’ `tests` attribute. `BenchmarkRunner` also inherits method `run_tests` as a part of it's interface which shouldn't be called by a user.
I’m also not sure how I feel about `ModelgaugeTestWrapper`. The object space already feels a bit cluttered, so I’m not personally a fan of the idea to create an entire wrapper object just to handle TestRunItem  transformation in two of it’s methods. But it’s not that big of a deal for me.

src/modelbench/benchmark_runner.py
```diff
@@ -0,0 +1,392 @@
...
+
+class BenchmarkRunner(TestRunner):
+    def __init__(self, data_dir: pathlib.Path):
+        super().__init__(data_dir)
+        self.benchmarks = []
+
+    def add_benchmark(self, benchmark: BenchmarkDefinition):
+        self.benchmarks.append(benchmark)
+
+    def _check_ready_to_run(self):
```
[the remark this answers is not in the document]
wpietri: Fixed!

wpietri:
>  For example, `BenchmarkRunner` inherits attribute `tests` which is never used, but `BenchmarkRun` on the other hand _does_ use its’ `tests` attribute. `BenchmarkRunner` also inherits method `run_tests` as a part of it's interface which shouldn't be called by a user.

This is now fixed. I confess I think the TestRunner/TestRun extraction is a bit artificial, in that I'm not sure it will ever get used. If there are further potential improvements in relation to that, I would like them to wait for the PR that actually uses them.

> I’m also not sure how I feel about `ModelgaugeTestWrapper`. The object space already feels a bit cluttered, so I’m not personally a fan of the idea to create an entire wrapper object just to handle TestRunItem transformation in two of it’s methods. But it’s not that big of a deal for me.

I think the ModelGauge object graph is hideously cluttered, as I've mentioned [here](https://discord.com/channels/1137054779013615616/1276557295723089953/1276557433648713939) among many other places. And I think that clutter is both confusing and contagious, overcomplicating related code like this. ModelgaugeTestWrapper is an attempt to isolate this code from a lot of the clutter. What I'd eventually like to do is make ModelgaugeTestWrapper go away by either pushing the complexity into the tests or by further cleanup there. I just am trying to stop the clutter from metastasizing further.

bkorycki:
That makes sense. I agree modelgauge needs a lot of simplification as well. Thanks!

bkorycki (approved):

PR #444, opened 2024-09-12
Merge modelgauge

bkorycki:
A first stab at merging modelgauge and modelbench.

I omitted the huggingface plugin from this merge because it requires torch which is not compatible with python 3.12. I’ll bring it over in another PR.

I also copied publish_all.py over as-is. I haven’t tested that nor the docker file yet. Same with the docs. Not sure if any of that works.

I would also appreciate some scrutiny over the way I merged the github workflows.
Thanks!


wpietri (approved):
I confess I was expecting something different, which was just copying in the modelbench source and test folders. Merging the whole history seems fine if you can get it working. But if you're doing it this way, please separate out the tests into modelbench and modelgauge folders, and generally make sure things aren't getting jumbled together. Please also look carefully at places where files collide, like the workflows, project files, and docs.

rogthefrog:
I would also recommend "just" copying the source code files and directories from one repo into the other, without any git history. If we want to refer back to the pre-merge history, we can check out the repo again. Once the merge is done, we can leave the old repo up in archived status, and include a note pointing people to this. I think it'll make the merge a bit easier to reason about.

bkorycki:
Closed with #455

PR #446, opened 2024-09-13
New v1 hazards with real prompts

bkorycki:
- Updated modelgauge version
- Added new hazards  to v1.0 benchmark which now run actual prompts
- Added **fake v1.0 reference scores** to get the tests to run. We will need to do an actual calibration run soon.


wpietri (approved):
Ooh, exciting! Thanks for doing this.

tests/test_benchmark.py
```diff
@@ -52,8 +54,9 @@ def test_benchmark_v1_definition_basics():
     assert mbb.name() == "General Purpose Ai Chat Benchmark V 1"
     assert mbb.path_name() == "general_purpose_ai_chat_benchmark_v_1"
     h = mbb.hazards()
-    assert len(h) == 1
-    assert h[0].__class__ == SafeDfmHazardV1
+    assert len(h) == 12
+    assert h[0].__class__ == SafeVcrHazardV1
```
rogthefrog, line 58: Any reason we're not using `assert isinstance(h[0], SafeVcrHazardV1)`?  

(this is just a personal preference)

rogthefrog (approved):
Yay!

PR #448, opened 2024-09-14
Parallel runner in use

wpietri:
Applying the parallel runner to the CLI. Removing some unused CLI commands. Assorted minor tidying.


src/modelbench/run.py
```diff
@@ -81,7 +70,7 @@ def cli() -> None:
     help="Path to directory containing custom branding.",
 )
 @click.option("--anonymize", type=int, help="Random number seed for consistent anonymization of SUTs")
-@click.option("--parallel", default=False, help="Experimentally run SUTs in parallel")
+@click.option("--parallel", default=False, help="Obsolete flag, soon to be removed")
```
rogthefrog, line 73: I was hoping Click provided a `deprecated` argument for options, but unfortunately it doesn't (natively). It does for commands. An enterprising soul added that for options here:

https://github.com/sqlfluff/sqlfluff/pull/3904/commits/a38512f217f911ffb575a2c3d54f8bbf3593dd95

Do we care?

src/modelbench/run.py
```diff
@@ -134,68 +125,27 @@ def find_suts_for_sut_argument(sut_args: List[str]):
...
-        benchmark_end_time = datetime.now(timezone.utc)
-        sut_scores.append(benchmark_definition.score(sut, hazard_scores, benchmark_end_time))
-    return sut_scores
+
+def run_benchmarks_for_suts(benchmarks, suts, max_instances, debug=False, json_logs=False, thread_count=32):
+    runner = BenchmarkRunner(pathlib.Path("./run"))
+    runner.secrets = load_secrets_from_config()
+    runner.benchmarks = benchmarks
+    runner.suts = suts
+    runner.max_items = max_instances
```
rogthefrog, line 143: Can `max_instances` default to something sensible based on the machine's capabilities?

tests/test_benchmark_runner.py
```diff
@@ -317,3 +317,42 @@ def test_sut_caching(self, item_from_test, a_wrapped_test, tmp_path):
     # TODO: add stats
     # TODO: track errors
     # TODO: handle logs
+
+
+class TestRunTrackers:
+    def test_null(self, capsys):
```
rogthefrog, line 323: This is cool!

rogthefrog (approved):
👍🏻 with just a couple of questions.

src/modelbench/run.py
wpietri: I looked for that argument as well. That particular code requires you to have an option you'd rather they use, which isn't the case here. There is an open issue for this: https://github.com/pallets/click/issues/2263

I've explained our use case, so perhaps one day they'll support it.

src/modelbench/run.py
wpietri: Speed doesn't vary much based on the user's hardware; it's about the capabilities of service providers like TogetherAI and what the user already has cached. A default here doesn't make much sense, except perhaps `None`, which would mean running everything. 

The higher-up default of 100 for running benchmarks should be about 3 minutes per SUT run, which seems like a good compromise between seeing a useful benchmark and not starting something that takes a week to run.

But the default for calibration should cover all the prompts, because we don't want anybody accidentally recalibrating the standards based on only a fraction of the data.

tests/test_benchmark_runner.py
wpietri: Thanks! The capsys thing was new to me too. Nice to just have it handled.

dhosterman (approved):
Looks great!

src/modelbench/run.py
wpietri: I tried it out. If I do it without a specific warning, it prints this:

```
Version 0.5 of this benchmark is a proof of concept only. Results are not intended to indicate actual levels of AI system safety.

/home/william/projects/mlcommons/modelbench/src/modelbench/run.py:98: UserWarning: --parallel option is unnecessary; benchmarks are now always run in parallel
  warnings.warn("--parallel option is unnecessary; benchmarks are now always run in parallel")
```

And if I add DeprecationWarning to it it doesn't print anything at all. (That appears to be on purpose; I think you have to run `python` with -X dev to get it to show deprecation warnings.

Overall, I prefer the bare message.

PR #455, opened 2024-09-17
Merge modelgauge with no git history

bkorycki:
Attempt 2 based on feedback from PR #444.
- Copied over most modelgauge files.
  - NOTE: huggingface_client.py was omitted following discussion on discord. TLDR: The ability to download/run models locally is not necessary + the torch dependency causes issues in certain runtime environments.
- Separate test directories for modelgauge and modelbench.
- Merged github workflows

Points requiring extra scrutiny:
- The github workflows.
- Publishing: I just did a straight-copy of modelgauge's publish_all.py and did not touch the existing Dockerfile. I would appreciate some extra guidance here if more steps are necessary!

To-Do:
- Merge docs. I will do this in a separate PR.


.github/workflows/python-app.yml
```diff
@@ -31,11 +32,14 @@ jobs:
...
-      run: poetry install
+      run: poetry install --no-interaction --with dev --extras all_plugins
 
     - name: Lint formatting
       run: poetry run black --check .
 
     - name: Test with pytest
       run: poetry run pytest
 
+    - name: Run mypy
```
wpietri, line 43: As long as this doesn't cause problems with modelbench, I'm fine with it. Otherwise, let's restrict it to modelgauge.

wpietri (approved):
This looks fine to me to merge. Do we want the demo_plugin and plugins directory at the top level? Fine to merge either way.

.github/workflows/python-app.yml
bkorycki: What kind of problems?

bkorycki: It was indeed causing problems. I excluded modelbench from this check.

rogthefrog (approved):
Thanks for doing this! The workflows look OK to me.

dhosterman (approved):
I think this looks solid! Thanks!

PR #456, opened 2024-09-18
add number of exceptions to hazard score output

dhosterman:
This adds the number of exceptions to hazards in the output json file of a run. It should be sufficient to direct an interested party to the forthcoming journal for additional data, as well as to populate the web interface with a numeric indication of issues that may have occurred during the run.


wpietri (approved):
Looks good as is, but I'm not sure exceptions is the precise thing to count at this level, in that we may end up with more ways for something to not be usable. For example, if our annotators have low confidence, we might just not count the item in the score. So maybe want to count two of a) total items started, b) items successfully process, and c) items unsuccessfully processed.

rogthefrog (approved):

rogthefrog:
Unrelated to these changes: are the `Hazard.score()` calls expensive, and would they benefit from memoization?

bkorycki (approved):
I think this will be very useful! I also like William's suggestion of broadening the scope to capture items skipped for any reason, not just exceptions. Maybe the fields can be `num_items_scored` and `num_items_skipped`?

wpietri:
> Unrelated to these changes: are the `Hazard.score()` calls expensive, and would they benefit from memoization?

They are not expensive. They're pretty simple calculations on top of the test scores.

Issue #453, opened 2024-09-17
AttributeError: 'SutDescription' object has no attribute '_instance'


danielwsap:
I'm running into an issue when trying to go through the tutorial for adding a SUT. After installing with poetry, I'm trying to run the demo yes no SUT but continue running into this error when trying to run it. Here is what my terminal output looks like

```
Version 0.5 of this benchmark is a proof of concept only. Results are not intended to indicate actual levels of AI system safety.

  0%|                                                                                                                           | 0/70 [00:00<?, ?items/s]Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/Library/Caches/pypoetry/virtualenvs/modelbench-zjKb4YZW-py3.12/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Library/Caches/pypoetry/virtualenvs/modelbench-zjKb4YZW-py3.12/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/Library/Caches/pypoetry/virtualenvs/modelbench-zjKb4YZW-py3.12/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Library/Caches/pypoetry/virtualenvs/modelbench-zjKb4YZW-py3.12/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Library/Caches/pypoetry/virtualenvs/modelbench-zjKb4YZW-py3.12/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Documents/modelbench/src/modelbench/run.py", line 100, in benchmark
    benchmark_scores = score_benchmarks([benchmark], suts, max_instances, json_logs, debug)
[...]

PR #457, opened 2024-09-18
Quick fix for small issues with custom SUTs

wpietri:
Hopefully fixes https://github.com/mlcommons/modelbench/issues/453

Just a few small things that only make a difference when running a custom SUT. We should probably add a smoke test for this, but let's talk about what form that should take.

### DIFF
--- a/src/modelbench/static_site_generator.py
+++ b/src/modelbench/static_site_generator.py
@@ -132,12 +132,14 @@ class StaticSiteGenerator:
         return self._content[item][key]
 
     @content.register
-    def content_sut(self, item: SutDescription, key: str):
-        if item.key in self._content:
-            return self._content[item.key][key]
+    def content_sut(self, sut_description: SutDescription, key: str):
+        if sut_description.key in self._content:
+            return self._content[sut_description.key][key]
+        elif key=="name":
+            return casefy.titlecase(sut_description.key)
         else:
             return f"{key} ({item})"
-            raise ValueError(f"Unknown SUT key for {item} and {key}")
+            warnings.warn(f"Can't find SUT content string for {item} and {key}")
 
     @content.register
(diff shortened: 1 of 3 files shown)


src/modelbench/static_site_generator.py
```diff
@@ -136,6 +136,7 @@ def content_sut(self, item: SutDescription, key: str):
         if item.key in self._content:
             return self._content[item.key][key]
         else:
+            return f"{key} ({item})"
```
rogthefrog, line 139: The `raise` below is unreachable now. Should we get rid of it?

bkorycki (approved):
Looks like a solid fix to me!
This may be a little out of scope, but this bug brought up a question for me: What's the intention in making `ModelGaugeSut` separate from `SutDescription`?  I could understand this design choice if we had other types of SUTs besides `ModelGaugeSut`. But I think all SUTs will necessarily have to be a `ModelGaugeSut`.

src/modelbench/static_site_generator.py
wpietri: Ah, right. I specifically wanted to ask @dhosterman about this. For SUTs that we control, we'll put in friendly names. For others, I'm not sure how to handle the in-report name. This change keeps things from catching on fire, but it's not great otherwise. Any ideas?

wpietri:
What's the intention in making `ModelGaugeSut` separate from `SutDescription`?

Good question. The other subclass of SutDescription is FakeSut, which is used when anonymizing reports.

src/modelbench/static_site_generator.py
wpietri: Since Daniel's out, I've change this to warn about the issue, as well as to do something reasonable when it's a name being looked for.

PR #466, opened 2024-09-20
Call the HuggingFace token a token consistently

rogthefrog:
The HuggingFace token was referred to as api_key in some places. This PR makes the naming consistent.

Together with that change, the token/key secret classes are moved out of the client code that uses them and into their own modules, so they can be imported into client code and tests and other modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ run/
 __pycache__/
 web/
 secrets/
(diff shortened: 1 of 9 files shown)


wpietri (approved):
Ooh, much tidier. Thanks!

PR #470, opened 2024-09-20
Update prompt file version

rogthefrog:
This addresses an error in the locale string (HI_HI instead of HI_IN).


bkorycki (approved):
thanks!

PR #471, opened 2024-09-20
Register mistral nemo

bkorycki:
Registers a mistral-nemo huggingface SUT that points to workstream 3's dedicated huggingface endpoint. 

Note: WS3 keeps this endpoint paused when not in use. `HuggingFaceInferenceSUT` automatically resumes an endpoint if it's scaled to 0. But paused endpoints need to be manually started. If this becomes an issue, we can change the SUT to automatically resume paused endpoints as well.


wpietri (approved):
makes sense to me!

PR #476, opened 2024-09-22
Fix measurement bug

wpietri:
Aha! I found it. The cause: I shot myself in the foot with a standard python footgun. Not sure how to test for this in the general case. Is there a linter that would catch this?

### DIFF
--- a/tests/modelbench_tests/test_benchmark_runner.py
+++ b/tests/modelbench_tests/test_benchmark_runner.py
@@ -146,7 +146,7 @@ class TestRunners:
     def hazard(self):
         pass
 
-    def test_test_run_items_properly_isolated(self,a_wrapped_test):
+    def test_test_run_items_properly_isolated(self, a_wrapped_test):
 
         a = TestRunItem(a_wrapped_test, self.make_test_item("one", "1"))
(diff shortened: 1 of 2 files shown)


dhosterman (approved):
Good find! I don't know of any linter that catches stuff like that, but it makes sense there should be one, right?

bkorycki (approved):

tests/modelbench_tests/test_benchmark_runner.py
```diff
@@ -73,7 +74,10 @@ def load_plugins(self):
 
     @pytest.fixture()
     def item_from_test(self):
-        return TestItem(prompts=[PromptWithContext(prompt=TextPrompt(text="Hello!"), source_id="hello")])
+        return self.make_test_item()
+
+    def make_test_item(self, text="Hello!", source_Id="hello"):
```
rogthefrog, line 79: Is the capital I in `source_Id` intentional?

rogthefrog (approved):

tests/modelbench_tests/test_benchmark_runner.py
wpietri: It is not. Thanks for catching that!

PR #480, opened 2024-09-23
Throw out bad data file items but warn about it

wpietri:
If a raw data file has a bad hazard code, this will warn about it and discard the line.


bkorycki (approved):
Looks good! Thanks:)

dhosterman (approved):
👍

rogthefrog (approved):

PR #487, opened 2024-09-24
Tests specify annotator ids

bkorycki:
This addresses the separation of concerns detailed [here](https://github.com/mlcommons/modelgauge/issues/570).

PR #510, opened 2024-09-27
Benchmark runner instantiates annotators

bkorycki:
PR #487 did not apply its test annotator instantiation change to modelbench's runner. This fixes that.

I also snuck in a bug fix for the "sxc" prompt issue that was causing some tests to fail.

### DIFF
--- a/src/modelbench/benchmark_runner.py
+++ b/src/modelbench/benchmark_runner.py
@@ -192,7 +192,6 @@ class TestRunBase:
         self.test_records = defaultdict(dict)
 
     def add_test(self, test: PromptResponseTest):
-        assert self._test_lookup.get(test, None) is None, f"Test {test.uid} already added"
         wrapped = ModelgaugeTestWrapper(test, self.test_data_path)
         self.tests.append(wrapped)
         self._test_lookup[test] = wrapped
--- a/src/modelbench/benchmark_runner.py
+++ b/src/modelbench/benchmark_runner.py
@@ -353,7 +353,6 @@ class TestRunAnnotationWorker(IntermediateCachingPipe):
         return item
 
     def collect_annotations(self, item):
-        # TODO: Parallelize annotators
         test_annotators = self.test_run.test_annotators[item.test.uid]
         for annotator in test_annotators:
             annotator_request = annotator.translate_request(item.prompt_with_context(), item.completion())
--- a/src/modelbench/benchmark_runner.py
+++ b/src/modelbench/benchmark_runner.py
@@ -14,7 +14,9 @@ from typing import Mapping, Iterable, Sequence, List, Optional, Any
 import diskcache
 from modelgauge.annotation import Annotation
 from modelgauge.annotator import CompletionAnnotator
+from modelgauge.annotator_registry import ANNOTATORS
 from modelgauge.base_test import PromptResponseTest, TestResult
+from modelgauge.config import raise_if_missing_from_config
 from modelgauge.dependency_helper import FromSourceDependencyHelper
 from modelgauge.pipeline import Source, Pipe, Sink, Pipeline, NullCache
 from modelgauge.records import TestRecord
@@ -179,6 +181,7 @@ class TestRunBase:
         self.suts = runner.suts
         self.max_items = runner.max_items
         self.tests = []
+        self.test_annotators = {}
         self._test_lookup = {}
         self.run_tracker = runner.run_tracker
         self.completed_item_count = 0
@@ -189,9 +192,23 @@ class TestRunBase:
         self.test_records = defaultdict(dict)
 
     def add_test(self, test: PromptResponseTest):
+        assert self._test_lookup.get(test, None) is None, f"Test {test.uid} already added"
         wrapped = ModelgaugeTestWrapper(test, self.test_data_path)
         self.tests.append(wrapped)
         self._test_lookup[test] = wrapped
+        self._add_test_annotators(test)
+
+    def _add_test_annotators(self, test: PromptResponseTest):
+        # Check for missing secrets without instantiating any objects
+        missing_secrets = []
+        for annotator_uid in test.get_annotators():
+            missing_secrets.extend(ANNOTATORS.get_missing_dependencies(annotator_uid, secrets=self.secrets))
+        raise_if_missing_from_config(missing_secrets)
+
+        annotators = []
+        for annotator_uid in test.get_annotators():
+            annotators.append(ANNOTATORS.make_instance(annotator_uid, secrets=self.secrets))
+        self.test_annotators[test.uid] = annotators
 
     def add_finished_item(self, item: "TestRunItem"):
         if item.completion() and item.annotations and not item.exception:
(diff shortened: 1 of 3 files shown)


src/modelbench/benchmark_runner.py
```diff
@@ -189,9 +192,23 @@ def __init__(self, runner: "TestRunnerBase"):
         self.test_records = defaultdict(dict)
 
     def add_test(self, test: PromptResponseTest):
+        assert self._test_lookup.get(test, None) is None, f"Test {test.uid} already added"
```
wpietri, line 195: Is it a bad thing if there are duplicates in tests? If so, that should be solved at the TestRunner or TestRunnerBase level.

src/modelbench/benchmark_runner.py
```diff
@@ -336,7 +353,9 @@ def handle_item(self, item: TestRunItem) -> TestRunItem:
         return item
 
     def collect_annotations(self, item):
-        for annotator_key, annotator in item.test.get_annotators().items():
+        # TODO: Parallelize annotators
```
wpietri, line 356: This is already parallel if the `thread_count` parameter is greater than 1. If a particular run needs faster annotation, easiest just to increase that number. It would be possible to also split this into 3 stages (an assigner stage, a single-annotation worker stage, and a collection/measurement stage), but if you'd like to do that please make an empirical case that it'll improve the experience for some user, as my intuition is that it won't make much of a practical difference.

src/modelbench/benchmark_runner.py
```diff
@@ -336,7 +353,9 @@ def handle_item(self, item: TestRunItem) -> TestRunItem:
...
+        test_annotators = self.test_run.test_annotators[item.test.uid]
```
wpietri, line 357: This might be better as a convenience method in test_run, like `annotators_for(test)` or `annotators_for(test_item)`.

tests/modelbench_tests/test_benchmark_runner.py
```diff
@@ -36,26 +36,35 @@ def fake_all_secrets(value="some-value") -> RawSecrets:
     return raw_secrets
 
 
+class FakeExplodingAnnotator(FakeAnnotator):
```
wpietri, line 39: This pollutes a global variable with test-specific data, and as a general rule tests should not have side effects.

tests/modelbench_tests/test_benchmark_runner.py
```diff
@@ -146,6 +159,30 @@ def sut_response(self):
     def hazard(self):
         pass
 
+    def test_test_run_loads_annotators(self, tmp_path, item_from_test, benchmark):
```
wpietri, line 162: Ah, good stuff. Since the runner and the run are so tightly bound, I like the choice of testing it like this.

tests/modelbench_tests/test_benchmark_runner.py
```diff
@@ -204,12 +241,15 @@ def test_benchmark_sut_worker_throws_exception(self, item_from_test, a_wrapped_t
         assert result.sut_response is None
         assert isinstance(result.exception, ValueError)
 
-    def test_benchmark_annotation_worker(self, a_wrapped_test, tmp_path, item_from_test, sut_response, a_sut):
-        baw = TestRunAnnotationWorker(self.a_run(tmp_path, suts=[a_sut]))
+    def test_benchmark_annotation_worker(
+        self, a_wrapped_test, tmp_path, item_from_test, sut_response, a_sut, benchmark
+    ):
+        baw = TestRunAnnotationWorker(self.a_run(tmp_path, suts=[a_sut], benchmarks=[benchmark]))
```
wpietri, line 247: I'm not following why this needs now need benchmarks.

wpietri (changes requested):

tests/modelbench_tests/test_benchmark_runner.py
bkorycki: Because the worker gets the annotators from the tests in the benchmark. Without setting benchmarks here, it won't run any annotators and this test fails.

src/modelbench/benchmark_runner.py
bkorycki: For example, the ensemble annotator from ws3 uses annotators hosted on different platforms (e.g. together, huggingface, etc..). The sequential annotation process introduces an avoidable bottleneck when running a diverse annotator set. We could get a significant speed boost from parallelizing the annotation stage. I don't think it's super high priority, just something to think about in the future.

wpietri: I understand that multiple annotators will sometimes be involved, and that splitting the annotation would have a theoretical benefit,  but I'm not convinced there is a significant speed boost to be had for real world runs. So please make an empirical case, one with actual measured numbers, to demonstrate that the additional complexity is justified by some user's improved experience.

src/modelbench/benchmark_runner.py
bkorycki: I think we should catch if duplicate tests are added, I don't see any scenarios where that would be intentional.
I did want to put the check in the runner originally, but the BenchmarkRunner and TestRunner don't add tests the same way so there would be duplicate code.

src/modelbench/benchmark_runner.py
bkorycki: I don't have those numbers so I will delete the todo.

src/modelbench/benchmark_runner.py
bkorycki: Okay! Updated.

tests/modelbench_tests/test_benchmark_runner.py
bkorycki: That's a good point, I added another test for the TestRun/Runner. I kept the benchmark test because the BenchmarkRun and TestRun add tests differently and I think it's worth making sure it works for both,

wpietri: Sounds great!

tests/modelbench_tests/test_benchmark_runner.py
wpietri: In general, I prefer to avoid global variables, which neatly solves this problem. But since fixing that is probably out of scope for this, I think the next-best solution is either setup/teardown (which add the fakes and then restore state after the test has run) or mocking such that the test diverts the production code so that it uses a fake registry which is discarded after the test has run.

src/modelbench/benchmark_runner.py
wpietri: I'm still not seeing the bad thing. If somebody accidentally adds it twice, can't we just be idempotent and ignore the second one? Generally the closer we get to a human user, the more I think code should be forgiving and honor DWIM.

If we do have to throw an error because what they're asking for doesn't make sense, since somebody using the BenchmarkRunner won't be adding tests at all, I think we could just put a human-focused error message in the TestRunner, so that they are alerted to the problem immediately. If there's a problem, that is.

bkorycki: I see your point, I'll remove this assertion!

tests/modelbench_tests/test_benchmark_runner.py
bkorycki: That makes sense! Updated.

wpietri (approved):
This all looks great now.

dhosterman (approved):
👍  Looks great!

PR #538, opened 2024-10-01
Update to more recent API versions.

wpietri:
Shouldn't change anything material, just keeping up with the world.


dhosterman (approved):
👍

Issue #509, opened 2024-09-26
Update benchmark runner with new test interface


bkorycki:
PR #487 did not update annotator instantiation in the benchmark runner.

PR #547, opened 2024-10-01
Fix the smoke test (except the benchmark runs)

wpietri:
This fixes a number of smoke test issues, but disables the benchmark runs in the hopes that they'll be fixed by #509, after which they should be reenabled.


dhosterman (approved):
👍 Get outta here, email alerts!

bkorycki (approved):
Looks good, thanks!

PR #552, opened 2024-10-02
Annotator runner test doesn't consider order of items

bkorycki:
This fixes the failing python 3.12 test.


wpietri (approved):
I like where you're going, and if you've tested this I guess it works, but I'm skeptical, as I'm not sure what asserting that something is in an iterator means. I think it's probably clearer to read in all the items and then do the `assert in` on a list of things.

bkorycki:
@wpietri Updated!

PR #553, opened 2024-10-02
Use prompts with new locale codes

bkorycki:
- Updated locale codes in both the prompt files and the code (e.g. EN_US -> en_US).
- Always pull prompts from "all_practice_skilled_unskilled.csv" in modellab


wpietri (approved):
:rocket:

PR #561, opened 2024-10-07
Add documentation about poetry environment and installation

bkorycki:


README.md
```diff
@@ -43,13 +43,14 @@ install ModelBench as a CLI tool or library to use in your own projects.
 ```shell
 pipx install poetry
 ```
+Please be sure to install Poetry in a dedicated python virtual environment separate from the environment used for this project.
```
wpietri, line 46: This is a surprise to me. Doesn't pipx create its own environment?

If it is something you want people to do, you'll have to give more detail.

README.md
```diff
@@ -43,13 +43,14 @@ install ModelBench as a CLI tool or library to use in your own projects.
...
 
 2. Clone this repository.
 ```shell
 git clone https://github.com/mlcommons/modelbench.git
 ```
 
-3. Install ModelBench and dependencies.
+3. Install ModelBench and dependencies. Make sure the dedicated python environment you used to install Poetry is not activated when running poetry commands.
```
wpietri, line 53: What's written here seems to me not to be the case. I haven't had problems running poetry from the project's environment, or from the default one. If there is something you want people to do, you'll at least have to be more explicit about it, and possibly mention the error message they'd be seeing if they don't do what you want.

wpietri (changes requested):

README.md
dhosterman: Pipx does create its own virtual environment(s) for everything you install in it.

README.md
bkorycki: It's not the project's environment that's the problem, it's the environment that was created to install poetry. This is based on the installation section of the poetry docs https://python-poetry.org/docs/

bollacker (approved):
LGTM

PR #567, opened 2024-10-08
exclude python 3.12.5, which contains a regression that breaks black

rogthefrog:
There's a regression in cpython 3.12.5

https://github.com/python/cpython/issues/123821

This excludes that version of Python so our black tests don't fail or refuse to run just in case our runtime is 3.12.5.


wpietri (approved):
Neat! I didn't even know you could do that.

PR #592, opened 2024-10-14
Fix multiple suts bug

bkorycki:


wpietri (approved):
Looks great! Thanks for fixing this.

PR #598, opened 2024-10-15
Secret runs

wpietri:
Redo secret annotation loading to make it controllable.
Make `key` methods consistent for accessing TOML.
Improve HasUid machinery to aid in key generation.
Update uids to be consistent with new scheme.
Update TOML files to match.


src/modelbench/run.py
```diff
@@ -144,15 +152,27 @@ def find_suts_for_sut_argument(sut_args: List[str]):
...
-def get_benchmark(version: str, locale: Optional[Locale]) -> BenchmarkDefinition:
+def ensure_ensemble_annotators_loaded():
+    try:
+        from modelgauge.private_ensemble_annotator_set import EnsembleAnnotatorSet, ensemble_secrets
+
+        private_annotators = EnsembleAnnotatorSet(secrets=ensemble_secrets(load_secrets_from_config()))
+        modelgauge.tests.safe_v1.register_private_annotator_tests(private_annotators, "ensemble")
+        return True
+    except Exception as e:
+        warnings.warn(f"Can't load private ensemble annotators: {e}")
```
rogthefrog, line 163: I take it that this failure isn't a blocker for the run?

src/modelbench/run.py
```diff
@@ -144,15 +152,27 @@ def find_suts_for_sut_argument(sut_args: List[str]):
...
-            warnings.warn("Locale is not used in v0.5 benchmarks.")
         return GeneralPurposeAiChatBenchmark()
     elif version == "1.0":
-        if locale is None:
-            locale = Locale.EN_US
-        return GeneralPurposeAiChatBenchmarkV1(locale)
+        if evaluator == "ensemble":
+            if not ensure_ensemble_annotators_loaded():
+                print("Can't build benchmark for {str} {locale} {evaluator}")
+                exit(1)
```
rogthefrog, line 174: This answered my question above. :)

src/modelbench/run.py
```diff
@@ -243,21 +266,23 @@ def update_standards_to(standards_file):
...
+        print("Can't load private annotators needed for calibration")
+        exit(1)
+
+    benchmarks = [GeneralPurposeAiChatBenchmark()]
+    for l in Locale:
+        benchmarks.append(GeneralPurposeAiChatBenchmarkV1(l, "ensemble"))
+    run_result = run_benchmarks_for_suts(benchmarks, reference_suts, 100)
+    all_hazard_numeric_scores = defaultdict(list)
     for benchmark, scores_by_sut in run_result.benchmark_scores.items():
+        print(benchmark)
```
rogthefrog, line 279: Needed?

src/modelbench/run.py
```diff
@@ -243,21 +266,23 @@ def update_standards_to(standards_file):
...
         for sut, benchmark_score in scores_by_sut.items():
             for hazard_score in benchmark_score.hazard_scores:
                 all_hazard_numeric_scores[hazard_score.hazard_definition.uid].append(hazard_score.score.estimate)
 
+    print(all_hazard_numeric_scores)
```
rogthefrog, line 284: Needed? Or is `print` the desired way to show this output? (as opposed to logging, etc)

src/modelbench/templates/content/hazards.toml
```diff
@@ -26,194 +26,50 @@ description = "Responses that enable, encourage, or endorse acts of intentional
...
+[safe_hazard-1_0-prv]
 name = "Privacy Violations"
-description = "Mind your business!"
+description = "Responses that violate individual privacy."
 
-[safe_hazard_ipv_en_us_1_0]
+[safe_hazard-1_0-ipv]
 name = "Intellectual Property Violations"
-description = "Hey that was my idea!"
+description = "Responses that provide more than 1_0 words from a copyrighted work."
```
rogthefrog, line 59: What is "more than 1_0" ? And doesn't this include responses that help you break IP rules?

src/modelgauge/tests/safe_v1.py
```diff
@@ -172,34 +182,28 @@ class SafeTestVersion1(BaseSafeTestVersion1):
...
-
-
-def register_tests(cls, postfix=None):
+def register_tests(cls, evaluator=None):
     for locale in Locale:
         for hazard in cls.hazards:
-            test_uid = create_safe_test_v1_uid(hazard, locale, postfix=postfix)
-            TESTS.register(cls, test_uid, hazard, locale, ALL_PERSONAS)
+            test_uid = BaseSafeTestVersion1.create_uid(hazard, locale, evaluator)
+            if not test_uid in TESTS.keys():
```
bkorycki, line 189: I'm curious about the reasoning behind adding this check. The test registry raises an error when you attempt to register multiple tests to the same UID. This is the preferred behavior IMHO, because otherwise you don't know which test is actually being used if execution continues normally.

src/modelbench/templates/content/hazards.toml
```diff
@@ -26,194 +26,50 @@ description = "Responses that enable, encourage, or endorse acts of intentional
...
-description = "Responses that demean or dehumanize people on the basis of their sensitive, personal characteristics."
-
-[safe_hazard_ssh_zh_cn_1_0]
-name = "Suicide & Self-Harm"
-description = "Responses that enable, encourage, or endorse acts of intentional self-harm."
-
-[safe_hazard_sxc_zh_cn_1_0]
-name = "Sexual content"
-description = "Definitely nsfw!"
+description = "Responses that are sexually expcicit."
```
rogthefrog, line 75: Typo: "explicit"

src/modelbench/benchmarks.py
```diff
@@ -57,7 +56,11 @@ def name(cls):
         return casefy.titlecase(cls.__name__.replace(BenchmarkDefinition.__name__, ""))
 
     def path_name(self):
```
bkorycki, line 58: Is path_name still used anymore? If not, could we get rid of it:)

src/modelbench/benchmarks.py
```diff
@@ -100,14 +103,22 @@ class GeneralPurposeAiChatBenchmarkV1(BenchmarkDefinition):
...
 
-    def __init__(self, locale: Locale):
+    def __init__(self, locale: Locale, evaluator="default"):
         self.locale = locale
+        self.evaluator = evaluator
         super().__init__()
 
-    def _make_hazards(self) -> Sequence[HazardDefinition]:
-        return [SafeHazardV1(hazard_key, self.locale) for hazard_key in SafeHazardV1.all_hazard_keys]
+    def key(self):
```
bkorycki, line 111: Are we certain that the benchmark should have identical TOML content for different evaluators? I think there should probably be differences in the UI that indicate which evaluator was used.

tests/modelgauge_tests/test_safe.py
```diff
@@ -256,10 +251,10 @@ def safe_test(self):
     def test_make_test_items_skips_out_of_scope_prompts(self, tmpdir, safe_test):
         csv_header = ["prompt_uid", "prompt_text", "hazard", "locale", "persona"]
         rows = [
-            ["1", "a", self.hazard, Locale.EN_US, SafePersonasVersion1.NORMAL],
-            ["2", "a", self.hazard, Locale.EN_US, SafePersonasVersion1.SKILLED],
-            ["3", "a", self.hazard, Locale.FR_FR, SafePersonasVersion1.NORMAL],
-            ["4", "a", self.hazard, Locale.FR_FR, SafePersonasVersion1.SKILLED],
+            ["1", "a", self.hazard, "en_US", "normal"],
```
rogthefrog, line 254: Mixed case in the locale keys makes me nervous. We know those strings are easy to get wrong (cf. "HI_HI" from a few weeks ago) and I wonder if mixed case will make that worse. 

At the same time, xx_YY is the standard case for locale language strings, so that's as good a reason as any to use xx_YY.

(I don't have an alternative proposal, I'm just bringing it up as a concern)

tests/modelgauge_tests/test_safe.py
```diff
@@ -321,3 +316,12 @@ def test_registered_tests(self):
             for locale in locales:
                 uid = f"safe-{hazard}-{locale}-1.0"
                 assert TESTS._lookup.get(uid)
+
+    def test_uid_construction(self):
+        # locale variants
+        assert BaseSafeTestVersion1.create_uid("xyz", "en_US") == "safe-xyz-en_us-1.0"
+        assert BaseSafeTestVersion1.create_uid("xyz", Locale.EN_US) == "safe-xyz-en_us-1.0"
```
rogthefrog, line 323: Should we consider updating the `Locale` enum to use mixed-case language strings?

rogthefrog (approved):
This is neat! A few questions in comments.

tests/modelgauge_tests/test_safe.py
wpietri: Yeah, personally I'd rather use all lower case, as I have in the CLI options and the UIDs. I considered making Locale accept any case, but decided this PR was huge anyhow.

src/modelbench/run.py
```diff
@@ -144,15 +152,27 @@ def find_suts_for_sut_argument(sut_args: List[str]):
...
-        if locale is not None:
-            warnings.warn("Locale is not used in v0.5 benchmarks.")
         return GeneralPurposeAiChatBenchmark()
     elif version == "1.0":
-        if locale is None:
-            locale = Locale.EN_US
-        return GeneralPurposeAiChatBenchmarkV1(locale)
+        if evaluator == "ensemble":
+            if not ensure_ensemble_annotators_loaded():
+                print("Can't build benchmark for {str} {locale} {evaluator}")
```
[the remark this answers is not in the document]
wpietri: Good catch. It totally should.

src/modelbench/uid.py
```diff
@@ -34,16 +36,21 @@ class ClassyObject(ABC, HasUid):
...
+        _uid_definition = {"class": "self", "flavor": "self.flavor"}
 
-        uid_def = self.__class__._uid_definition
+    This object's UID would be "ice_cream-chocolate"
+    """
 
+    @staticmethod
+    def _render_uid(self, uid_def):
         def clean_string(s):
+            if isinstance(s, Locale):
```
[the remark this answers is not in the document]
wpietri: Good idea. I was just solving for a problem, but that makes good sense.

src/modelgauge/tests/safe_v1.py
wpietri: Ah, got it. Something was blowing up, so it seemed to me that making this idempotent was reasonable. Could you say more about the case you're concerned about? I

src/modelbench/benchmarks.py
wpietri: I'll check.

src/modelbench/benchmarks.py
wpietri: There perhaps should be, but currently there isn't. There's already a bunch of duplication in the TOML and I didn't want to make that worse for no gain. Personally I'm hoping we throw out the TOML shortly, as it was based on a theory — non-engineers keeping the TOML updated — that didn't pan out.

Maybe add a card for that? A simple solution might be putting the UIDs into the info box at the bottom of the page.

src/modelbench/run.py
wpietri: Totally not. Debugging code I left in. Thanks for catching it.

src/modelbench/run.py
wpietri: Nope, leftover debugging code again. Thanks.

src/modelbench/templates/content/hazards.toml
wpietri: Thanks, that's a typo,which I'll fix. I got the content from the taxonomy: https://docs.google.com/spreadsheets/d/1ZCxh99kHciZwAgpwT-UTf0UkGXJsde3CMKrrf7agd9M/edit?gid=1325194442#gid=1325194442

src/modelbench/templates/content/hazards.toml
wpietri: Thanks!

tests/modelgauge_tests/test_safe.py
wpietri: Ah, agreed. Since we both want it, I'll do that now.

dhosterman (approved):
Looks good, nothing to add to above comments!

src/modelbench/run.py
wpietri: Sometimes yes, sometimes no. I started out always throwing an exception, but sometimes you want to just know if they are available.

src/modelgauge/tests/safe_v1.py
wpietri: I found what blows up. In `get_benchmark` it calls `ensure_ensemble_annotators_loaded` if it needs an ensemble annotator. And If we're doing multiple languages, that gets called multiple times. So I'd like something to be idempotent, but it doesn't have to be this bit.

src/modelbench/benchmarks.py
wpietri: Yes, still being used by the static site generator.

bkorycki: Ah ok

src/modelbench/run.py
```diff
@@ -144,15 +152,27 @@ def find_suts_for_sut_argument(sut_args: List[str]):
     return suts
 
 
-def get_benchmark(version: str, locale: Optional[Locale]) -> BenchmarkDefinition:
+def ensure_ensemble_annotators_loaded():
+    try:
+        from modelgauge.private_ensemble_annotator_set import EnsembleAnnotatorSet, ensemble_secrets
+
+        private_annotators = EnsembleAnnotatorSet(secrets=ensemble_secrets(load_secrets_from_config()))
+        modelgauge.tests.safe_v1.register_private_annotator_tests(private_annotators, "ensemble")
```
bkorycki, line 159: Would it be possible to keep the registration statements in the ensemble file?

PR #601, opened 2024-10-16
Add olmo and qwen suts

bkorycki:
Olmo isn't accessible via the chat_completion API so I created a new type of SUT.


wpietri (approved):
I'm a little concerned about the stability of that Olmo URL, but that's what they're advertising in the dashboard at the moment, and I don't see an API to fetch it, so let's run with this and see how it goes.

plugins/huggingface/modelgauge/suts/huggingface_api.py
```diff
@@ -0,0 +1,75 @@
...
+    def translate_response(self, request: HuggingFaceChatRequest, response: HuggingFaceResponse) -> SUTResponse:
+        return SUTResponse(completions=[SUTCompletion(text=response.generated_text)])
+
+
+HF_SECRET = InjectSecret(HuggingFaceInferenceToken)
+
+SUTS.register(
+    HuggingFaceSUT,
+    "olmo-7b-0724-instruct-hf",
+    "https://flakwttqzmq493dw.us-east-1.aws.endpoints.huggingface.cloud",
```
rogthefrog, line 73: @bkorycki to add to @wpietri 's comment, this worries me a little. Maybe we could make a mlcommons.org cname to put in the code, and update the cname if the HF endpoint changes? Or in secrets.toml?

wpietri: Long term, I'd like to create a layer that lets a more user-y person say they want `olmo-7b-0724-instruct` for something, and then lets an ops person say, "we prefer to use huggingface for that and here's how". And presumably that latter bit will be via a configuration file. That would help us, and it would also help other users of our code, because they may want to use a different llamaguard provider, including an internal one, rather than the one we've hardcoded.

But for now updating the source isn't much harder than updating a cname. And a cname only solves the problem for us, not for other people using this code. So I'm fine with just leaving this like this and then fixing it when we have more information about where we need things to flex.

PR #617, opened 2024-10-20
Smoke test improvements

wpietri:
* stop creating issues, as we don't look at them
* reenable the benchmark runs as part of the test


dhosterman (approved):
👍

rogthefrog (approved):

PR #626, opened 2024-10-22
updated for poetry usage

bollacker:


wpietri (commented):
I'm not getting the purpose of this. I think it's ok if they don't use poetry and instead use some other method, like PyPI.

Issue #607, opened 2024-10-18
Mistral model integrated


rogthefrog:

rogthefrog:
<html>
<body>

Mistral | Mistral Large | Mistral Nemo (12b)
-- | -- | --



</body>
</html>

PR #629, opened 2024-10-22
Add mistral large sut hosted on vertex ai

rogthefrog:
[Story](https://github.com/mlcommons/modelbench/issues/607)


plugins/mistralai_sut/modelgauge/suts/vertexai_client.py
```diff
@@ -0,0 +1,83 @@
...
+        return headers
+
+    def request(self, req: dict) -> dict:
+        try:
+            client = httpx.Client()
+            response = client.post(self._build_endpoint_url(), json=req, headers=self._headers(), timeout=None)
+            if response.status_code == 200:
+                return response.json()
+            else:
+                raise RuntimeError(f"VertexAI response code {response.status_code}")
```
wpietri, line 81: Any production SUT will need to have some retry logic sufficient to overcome transient errors and to keep retrying on rate limits.

wpietri (changes requested):

plugins/mistralai_sut/modelgauge/suts/vertexai_client.py
rogthefrog: OK

plugins/mistralai_sut/modelgauge/suts/vertexai_client.py
```diff
@@ -0,0 +1,83 @@
...
+class VertexAIClient:
+    def __init__(
+        self,
+        publisher: str,
+        model_name: str,
+        model_version: str,
+        streaming: bool,
+        api_key: VertexAIAPIKey,
+        project_id: VertexAIProjectId,
+        region: str = os.environ.get("GOOGLE_REGION"),
```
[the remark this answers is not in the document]
rogthefrog: OK

plugins/mistralai_sut/modelgauge/suts/vertexai_client.py
rogthefrog: I've added retry logic to the MistralAI sut, and left a comment to revisit this client if/when we use VertexAI again in the future.

rogthefrog:
@wpietri @bkorycki could you please take another look, and approve if it's OK (to remove "merging is blocked")?

wpietri:
> could you please take another look, and approve if it's OK (to remove "merging is blocked")?

I have approved, so I think it's down to @bkorycki  now.

plugins/vertexai/modelgauge/suts/vertexai_client.py
```diff
@@ -0,0 +1,95 @@
...
+        self.streaming = streaming
+        if isinstance(region, str):
+            self.region = region
+        elif isinstance(region, VertexAIRegion):
+            self.region = region.value
+        else:
+            raise ValueError("Incorrect GCP region.")
+
+    def _get_access_token(self) -> str:
+        credentials, _ = google.auth.default(scopes=["https://www.googleapis.com/auth/cloud-platform"])
```
bkorycki, line 66: How does this work? Does it require you to have credentials setup in gcloud CLI?

plugins/mistral/modelgauge/suts/mistral_sut.py
```diff
@@ -0,0 +1,87 @@
...
+
+model_name = "ministral-8b"
+model_version = "2410"
+model_uid = f"{model_name}-latest"
+SUTS.register(MistralAISut, model_uid, model_uid, model_version, MISTRAL_API_KEY)
+
+
+model_name = "mistral-large"
+model_version = "2411"
+model_uid = f"{model_name}-latest"
```
bkorycki, line 86: The version should be explicit in the uid e.g. `mistral-large-2411`

plugins/vertexai/modelgauge/suts/vertexai_mistral_sut.py
```diff
@@ -0,0 +1,111 @@
...
+        return SUTResponse(completions=completions)
+
+
+VERTEX_KEY = InjectSecret(VertexAIAPIKey)
+VERTEX_PROJECT_ID = InjectSecret(VertexAIProjectId)
+VERTEX_REGION = InjectSecret(VertexAIRegion)
+
+model_name = "mistral-large"
+model_version = "2407"
+model_uid = f"{model_name}-{model_version}"
```
bkorycki, line 108: I think @bollacker has mentioned in the past that the provider should also be a part of the UID. Especially if we supply multiple providers for one SUT.

bkorycki (changes requested):

bkorycki:
I get an error when I run `poetry run modelgauge run-sut --sut mistral-large-latest`:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for TextPrompt
text
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

plugins/vertexai/modelgauge/suts/vertexai_client.py
rogthefrog: You don't need the gcloud CLI itself, but you do need the application default credentials. I'll add that to the documentation.

bkorycki (approved):
I'm not sure that the authentication in the vertex AI plugin works the way we would want. But I don't want to hold up merging this PR since we won't actually be using that plugin for any upcoming runs, so we should just revisit that later on.

PR #630, opened 2024-10-23
Huggingface Chat Completion SUT produces logprobs

bkorycki:


plugins/huggingface/tests/test_huggingface_chat_completion.py
```diff
@@ -40,10 +40,30 @@ def prompt():
...
+        max_tokens=5,
+        temperature=1.0,
+    )
+
+
+@pytest.fixture
+def sut_request_with_logprobs():
+    return HuggingFaceChatCompletionRequest(
+        messages=[ChatMessage(role="user", content="some text prompt")],
+        logprobs=True,
```
wpietri, line 65: Given the duplication between these two fixtures, I think it'd be better to skip using fixtures here and just make it a function or method you call from the tests.

wpietri (approved):

plugins/huggingface/tests/test_huggingface_chat_completion.py
bkorycki: Updated!

wpietri (approved):
:rocket:

PR #632, opened 2024-10-23
Basic test suite for huggingface SUTs

bkorycki:


plugins/huggingface/tests/test_huggingface_api.py
```diff
@@ -0,0 +1,77 @@
...
+    return HuggingFaceChatRequest(
+        inputs="some text prompt", parameters=HuggingFaceChatParams(max_new_tokens=5, temperature=1.0)
+    )
+
+
+def test_huggingface_api_translate_text_prompt_request(fake_sut, prompt, sut_request):
+    request = fake_sut.translate_text_prompt(prompt)
+
+    assert isinstance(request, HuggingFaceChatRequest)
+    assert request == sut_request
```
wpietri, line 39: I think this test depends on the two fixtures happening to use the same "some text prompt", which isn't obvious from the test. I think the long-term solution here is simplifying the domain model so it's easy to construct real examples. But in the short term I'd suggest making `prompt()` into a method that takes the text as a parameter, and asserting specific things on the `request` object that demonstrate your intent as to what bits are important to be there.

plugins/huggingface/tests/test_huggingface_api.py
```diff
@@ -0,0 +1,77 @@
...
+        headers=ANY,
+        json={"inputs": "some text prompt", "parameters": {"max_new_tokens": 5, "temperature": 1.0}},
+    )
+
+
+@patch("requests.post", side_effect=mocked_requests_post)
+def test_huggingface_api_evaluate_dumps_result(mock_post, fake_sut, sut_request):
+    output = fake_sut.evaluate(sut_request)
+
+    assert output == HuggingFaceResponse(generated_text="response")
```
wpietri, line 69: Same concern here; to know what the test is testing you have to look elsewhere, and then it's not clear which are the important bits. It's best to think of tests as documenting your intent for programmers who come along later, but in a way that computers can verify still hold.

wpietri (changes requested):

plugins/huggingface/tests/test_huggingface_api.py
bkorycki: That makes sense. I made some changes, let me know what you think

wpietri (approved):
Looks great! Thanks.

PR #633, opened 2024-10-24
Add run journal

wpietri:
Adds a run journal.


bollacker (changes requested):
A few fieldname changes:

For prompts "queuing item":
source_id --> prompt_uid
prompt --> prompt_text
sut --> sut_uid

For responses "translated sut response"
text --> response_text
sut --> sut_uid

src/modelbench/benchmark_runner.py
```diff
@@ -185,6 +109,10 @@ def __init__(self, runner: "TestRunnerBase"):
         self.run_tracker = runner.run_tracker
         self.completed_item_count = 0
 
+        self.data_dir.mkdir(exist_ok=True, parents=True)
+        self.run_id = datetime.now().strftime("run-%Y%m%d-%H%M%S-%f")
+        self.journal = RunJournal(self.data_dir / f"journal-{self.run_id}.jsonl")
```
bkorycki, line 114: The data directory is used to stored data needed for a run e.g. dependencies, caches. Would it be possible to store the journal in a separate directory? Modelgauge writes its records to "output" by default-- the journals could go there as well.

src/modelbench/benchmark_runner.py
```diff
@@ -322,15 +252,23 @@ def __init__(self, test_run: TestRunBase):
 
     def handle_item(self, item: TestRunItem):
         for sut in self.test_run.suts:
-            self.downstream_put(TestRunItem(item.test, item.test_item, sut))
+            run_item = TestRunItem(item.test, item.test_item, sut)
+            self.test_run.journal.item_entry(
```
bkorycki, line 256: I think its also important to record the SUT options

src/modelbench/benchmark_runner.py
```diff
@@ -322,15 +252,23 @@ def __init__(self, test_run: TestRunBase):
...
 class TestRunSutWorker(IntermediateCachingPipe):
     def __init__(self, test_run: TestRunBase, cache: MBCache, thread_count=1):
+
         super().__init__(cache, thread_count)
         self.test_run = test_run
 
-    def handle_item(self, item):
+    def handle_item(self, item: TestRunItem):
         mg_sut = item.sut.instance(self.test_run.secrets)
         raw_request = mg_sut.translate_text_prompt(item.prompt_with_context().prompt)
```
bkorycki, line 273: Can we record the sut-specific request? To 1) ensure that the SUT options were translated as expected and 2) keep a record of the actual request objects.

src/modelbench/benchmark_runner.py
```diff
@@ -361,28 +307,48 @@ def handle_item(self, item: TestRunItem) -> TestRunItem:
...
+                        "fetched annotator response",
+                        item,
+                        annotator=annotator.uid,
+                        run_time=timer,
+                        response=annotator_response,
+                    )
+
+                annotation = annotator.translate_response(annotator_request, annotator_response)
+                self.test_run.journal.item_entry(
+                    "annotation translated", item, annotator=annotator.uid, annotation=annotation
```
bkorycki, line 343: I think in the future we should consider recording the raw annotation as well

src/modelbench/benchmark_runner.py
```diff
@@ -361,28 +307,48 @@ def handle_item(self, item: TestRunItem) -> TestRunItem:
...
+                self._debug(f"looking for {cache_key} in cache")
+                if cache_key in self.cache:
+                    self._debug(f"cache entry found")
+                    annotator_response = self.cache[cache_key]
+                else:
+                    self._debug(f"cache entry not found; processing and saving")
+                    with Timer() as timer:
+                        annotator_response = annotator.annotate(annotator_request)
+                    self.cache[cache_key] = annotator_response
+                    self.test_run.journal.item_entry(
```
bkorycki, line 333: I think we should also record if the response if fetched from the cache.

src/modelbench/benchmark_runner.py
```diff
@@ -503,13 +473,37 @@ def _check_ready_to_run(self):
     def run(self) -> BenchmarkRun:
         self._check_ready_to_run()
         benchmark_run = BenchmarkRun(self)
+        benchmark_run.journal.raw_entry(
```
bkorycki, line 476: Maybe we should make it clear which tests are associated with which benchmarks? Also, I think we may want to include more information about the SUTs/tests at some point. e.g. class type, plugin, initialization records, test-dependencies/file names

src/modelbench/benchmark_runner.py
```diff
@@ -525,7 +519,26 @@ def _calculate_benchmark_scores(self, benchmark_run):
 
                     assert test_records, f"No records found for {benchmark_definition} {sut} {hazard}"
 
-                    hazard_scores.append(hazard.score(test_records))  # TODO: score needs way less
-                benchmark_run.benchmark_scores[benchmark_definition][sut] = BenchmarkScore(
-                    benchmark_definition, sut, hazard_scores, end_time=datetime.now()
+                    hazard_score = hazard.score(test_records)
+                    hazard_scores.append(hazard_score)  # TODO: score needs way less
+                    benchmark_run.journal.raw_entry(
```
bkorycki, line 524: Maybe we should include record which hazard this is as well

src/modelbench/run_journal.py
```diff
@@ -0,0 +1,128 @@
...
+from modelbench.benchmark_runner_items import TestRunItem, Timer
+from modelgauge.sut import SUTResponse
+
+
+def for_journal(o):
+    """Turns anything into a collection of primitives suitable for JSON rendering."""
+    if isinstance(o, TestRunItem):
+        return {"test": o.test.uid, "item": o.source_id(), "sut": o.sut.uid}
+    if isinstance(o, SUTResponse):
+        completion = o.completions[0]
```
bkorycki, line 21: Can we add an assertion here that there is only 1 completion in the response? I'm worried about a potential future scenario in which we ask SUTs for multiple completions and forget to update this code because it's not obvious that the runs are only equipped to handle single-completion responses right now.

bkorycki (commented):
This looks really good! I think it captures a lot of the necessary information. I think the most important thing that is missing is SUT options.

src/modelbench/benchmark_runner.py
bkorycki: This stuff can probably wait for another PR though.

src/modelbench/run_journal.py
wpietri: Is that a plausible scenario? I was thinking that we'd eliminate this along with some of the other unused pluralities.

src/modelbench/benchmark_runner.py
wpietri: Well right now this is storing to a directory called `run`, and this seems like run stuff to me. I see the merit in splitting up input and and output, but caches are kinda in between. How about when folks are back we talk it over and figure out something that seems sensible for what we have and has room for what we expect in the future?

src/modelbench/benchmark_runner.py
wpietri: The SUT options are the same the whole way through, right? If so, I'd like to put them in once per SUT. What's the best way to get them?

src/modelbench/benchmark_runner.py
wpietri: Sure, no problem.

src/modelbench/benchmark_runner.py
wpietri: I think we're including the raw annotation response, the `fetched` entry above.

src/modelbench/benchmark_runner.py
bkorycki: >  The SUT options are the same the whole way through, right?

In practice, yes. Each prompt has it's own SUT options, so technically sut options could vary by prompt. But all of safe tests for 0.5 and 1.0 use the same options for all of their respective test items, so it's probably fine to record them once. 

> I'd like to put them in once per SUT.

The option values are provided by the test, so I think it should be recorded once per test instead. 

> What's the best way to get them?

Maybe from a test's first test item via `test_item.prompts[0].prompt.options`

src/modelbench/benchmark_runner.py
wpietri: Agreed! I meant to put that in; thanks for catching it.

src/modelbench/benchmark_runner.py
bkorycki: Could we at least store them in a subdirectory like `run/journals` in the meantime? To keep the `run` directory organized.

src/modelbench/benchmark_runner.py
wpietri: I think that's a fine follow-on PR. You should feel free.

src/modelbench/run_journal.py
bkorycki: I don't know how likely it is.

src/modelbench/benchmark_runner.py
wpietri: Yes! It was there but mislabeled. I'll fix.

src/modelbench/benchmark_runner.py
bkorycki: Oh you're right, I missed that!

src/modelbench/run_journal.py
wpietri: My general approach to possible futures is to deal with them as they come. Because the number of possible needs is much greater than the number of needs we'll actually have to fill, I think it's much more effective to keep the code simple and be ready to adapt than to build speculatively. Happy to make exceptions for things that we are pretty sure will happen, doing a little extra work to keep our options open, and to deal with things that are low probability but very high risk.

It sounds like in this case we should apply YAGNI, and if it turns out we're wrong we can pay the cost when we know exactly what we do need.

src/modelbench/benchmark_runner.py
wpietri: Great idea.

src/modelbench/benchmark_runner.py
wpietri: And is there an easy way to say what files were used?

bkorycki: You could maybe do `dependency_helper.versions_used()` which maps the dependency key to a has of the file contents.
```
"dependency_versions": {
        "all_practice_skilled_unskilled": "fd103633bbfe747a2a90e9daee9d49ace3064733d43ace70fbac88a71b3827d0"
    }
```

It's also *possible* to get the actual source URL, but it's not simple to get. If we want the source URL recorded as well, I think we should defer that to another PR.

bkorycki: Also yeah that looks good! Thanks.

src/modelbench/run_journal.py
bkorycki: Fair enough! Thanks.

src/modelbench/benchmark_runner.py
wpietri: I rummaged in the debugger and this turned out not to be too bad. In `ModelgaugeTestWrapper` I've been accumulating bits of code that serve to bridge what ModelBench wants a test to be and what ModelGauge thinks of it. It was pretty easy to add there. So now this record looks like this:

```
{
  "timestamp": "2024-10-25T16:13:57.513810+00:00",
  "message": "test info",
  "class": "BenchmarkRunner",
  "method": "run",
  "test": "safe-sxc-en_us-1.0",
  "initialization": {
    "module": "modelgauge.tests.safe_v1",
    "class_name": "SafeTestVersion1",
    "args": [
      "safe-sxc-en_us-1.0",
      "sxc",
      "en_US",
      [
        "normal",
        "skilled",
        "unskilled"
      ]
    ],
    "kwargs": {}
  },
  "sut_options": {
    "max_tokens": 500,
    "temperature": 0.01
  },
  "dependencies": {
    "all_practice_skilled_unskilled": {
      "source": "https://ai-safety-dev.uc.r.appspot.com/files/download/all_practice_skilled_unskilled.csv",
      "local_path": "run/tests/all_practice_skilled_unskilled/fd103633bbfe747a2a90e9daee9d49ace3064733d43ace70fbac88a71b3827d0",
      "timestamp": "2024-10-25T15:26:18.073360+00:00"
    }
  }
}
```

I included the timestamp of the local file so that we can catch issues of updating something on the server but not locally.

bollacker (approved):
LGTM

PR #654, opened 2024-10-30
Run ops improvements

wpietri:
Added a number of small improvements and fixes to make things work better when running things.

### DIFF
--- a/src/modelbench/benchmark_runner.py
+++ b/src/modelbench/benchmark_runner.py
@@ -1,9 +1,9 @@
 import json
+import logging
 import pathlib
 import random
 import sys
 import time
-import traceback
 from abc import abstractmethod
 from collections import defaultdict
 from datetime import datetime
@@ -29,6 +29,8 @@ from modelgauge.records import TestRecord
 from modelgauge.single_turn_prompt_response import (
     TestItem,
 )
+
+logger = logging.getLogger(__name__)
 
 
 class RunTracker:
@@ -115,7 +117,7 @@ class TestRunBase:
         self.run_id = datetime.now().strftime("run-%Y%m%d-%H%M%S-%f")
         journal_dir = self.data_dir / "journals"
         journal_dir.mkdir(exist_ok=True, parents=True)
-        self.journal = RunJournal(journal_dir / f"journal-{self.run_id}.jsonl")
+        self.journal = RunJournal(journal_dir / f"journal-{self.run_id}.jsonl.zst")
 
         self.caches = {}
         self.cache_starting_size = {}
@@ -225,8 +227,8 @@ class IntermediateCachingPipe(Pipe):
 
 
 class TestRunItemSource(Source):
-    def __init__(self, run: TestRunBase):
-        super().__init__()
+    def __init__(self, run: TestRunBase, queue_maxsize=0):
+        super().__init__(queue_maxsize)
         self.test_run = run
 
     def new_item_iterable(self, quiet=False) -> Iterable[TestRunItem]:
@@ -299,8 +301,7 @@ class TestRunSutWorker(IntermediateCachingPipe):
         except Exception as e:
             item.exceptions.append(e)
             self.test_run.journal.item_exception_entry("sut exception", item, e)
-            print(f"failure handling item {item}:", file=sys.stderr)
-            print(traceback.format_exc(), file=sys.stderr)
+            logger.error(f"failure handling sut item {item}:", exc_info=e)
         return item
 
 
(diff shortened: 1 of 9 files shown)


src/modelgauge/pipeline.py
```diff
@@ -70,11 +70,11 @@ class PipelineSegment(ABC):
 
     default_timeout = 0.1
 
-    def __init__(self):
+    def __init__(self, queue_maxsize=0):
         super().__init__()
         self._work_done = Event()
         self._upstream: Optional[PipelineSegment] = None
-        self._queue: Queue = Queue()
+        self._queue: Queue = Queue(queue_maxsize)
```
bkorycki, line 77: I'm curious what this is meant for

src/modelgauge/suts/together_client.py
```diff
@@ -29,7 +32,7 @@ def _retrying_post(url, headers, json_payload):
     """HTTP Post with retry behavior."""
     session = requests.Session()
     retries = Retry(
-        total=6,
+        total=10,
```
bkorycki, line 35: Dang, did you notice a difference after increasing this number?

bkorycki (commented):
These are solid improvements, thanks!

src/modelbench/benchmark_runner.py
```diff
@@ -312,19 +313,21 @@ def __init__(self, test_run: TestRunBase, cache: MBCache, thread_count=1, cache_
     def handle_item(self, item: TestRunItem) -> TestRunItem:
         try:
             if item.completion():
-                self.collect_annotations(item)
-                item.test.measure_quality(item)
-                self.test_run.journal.item_entry("item quality measured", item, measurements=item.measurements)
+                with Timer() as timer:
```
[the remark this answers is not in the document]
wpietri: I think that's happening at line 288, in the TestRunSutWorker. Were you looking for something else?

src/modelbench/benchmark_runner.py
```diff
@@ -115,7 +117,7 @@ def __init__(self, runner: "TestRunnerBase"):
         self.run_id = datetime.now().strftime("run-%Y%m%d-%H%M%S-%f")
         journal_dir = self.data_dir / "journals"
         journal_dir.mkdir(exist_ok=True, parents=True)
-        self.journal = RunJournal(journal_dir / f"journal-{self.run_id}.jsonl")
+        self.journal = RunJournal(journal_dir / f"journal-{self.run_id}.jsonl.zst")
```
rogthefrog, line 120: Modellab currently looks for a file named `run_{uid}_journal.jsonl`. It's trivial to change it there and I don't have a preference. Do you? @dhosterman WDYT?

rogthefrog (approved):
LGTM pending decision on file naming convention.

src/modelgauge/pipeline.py
wpietri: Before we had unbounded queues. This meant that it would, e.g., read in all prompts and queue items for every prompt * sut * annotator. And then those would sit there until one of the workers was ready. That's fine for small numbers of things, but when it's a lot of things, you get a lot of unnecessary memory usage, and can also get other behaviors.

For example, I think some of our overloading Together enough to get errors was making SUT requests long before we were ready to annotate them. But now, since the queue has a limited size, it'll only do so many ahead of need.

src/modelgauge/suts/together_client.py
wpietri: I did. This was based on me doing runs, getting failures, and tuning until we stopped getting failures.

src/modelbench/benchmark_runner.py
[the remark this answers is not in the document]
bkorycki: Ohh I didn't realize we already had that. thanks!

src/modelgauge/pipeline.py
bkorycki: Oh got it, that makes sense. Thanks!

bkorycki (approved):

src/modelbench/benchmark_runner.py
wpietri: Unfortunately, the run_ids for the benchmarks and the journal are not consistent. Each benchmark result file gets its own run id, but the journal is for all the benchmarks (or no benchmarks; it works for test runs too).

I'm happy to resolve it any way that makes sense, but I'd rather merge this and do the change in a separate PR unless we can get a quick answer here.

rogthefrog: OK let's merge this and I will discuss with Daniel.

rogthefrog (approved):

PR #657, opened 2024-10-31
Use final practice prompts

wpietri:
Use final practice prompts. Also disable non-English locales, as we no longer have prompts for them.


bkorycki (approved):

dhosterman (approved):
👍

rogthefrog (approved):

PR #659, opened 2024-10-31
Fix huggingface response caching bug

bkorycki:
The huggingface chat_completions API returns a dataclass/dict object `ChatCompletionOutput`. It inherits from `dict`, which is a "typeable" data structure and thus allows it to be cached. However, there was a bug when running this SUT on a prompt that has been previously cached. Because it's a dataclass and not a plain dict, the deserialization produced an error.

This PR addresses this bug by putting the response object into pydantic form before returning it in evaluate.

I also added some testing to make sure all SUTs are cacheable.

### DIFF
--- a/plugins/huggingface/modelgauge/suts/huggingface_chat_completion.py
+++ b/plugins/huggingface/modelgauge/suts/huggingface_chat_completion.py
@@ -32,7 +32,7 @@ class HuggingFaceChatCompletionRequest(BaseModel):
 class HuggingFaceChatCompletionOutput(BaseModel):
     choices: List[Dict]
     created: Optional[int] = None
-    id: Optional[int] = None
+    id: Optional[str] = None
     model: Optional[str] = None
     system_fingerprint: Optional[str] = None
     usage: Optional[Dict] = None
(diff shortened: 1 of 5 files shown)


plugins/huggingface/modelgauge/suts/huggingface_chat_completion.py
```diff
@@ -79,29 +85,38 @@ def translate_text_prompt(self, prompt: TextPrompt) -> HuggingFaceChatCompletion
...
+        )
 
     def translate_response(
-        self, request: HuggingFaceChatCompletionRequest, response: ChatCompletionOutput
+        self, request: HuggingFaceChatCompletionRequest, response: HuggingFaceChatCompletionOutput
     ) -> SUTResponse:
         completions = []
         for choice in response.choices:
-            text = choice.message.content
+            text = choice["message"]["content"]
```
rogthefrog, line 109: You may want to try the [benedict library](https://pypi.org/project/python-benedict/). It wraps `dict` and gives you convenient functionality like accessing elements via dot notation or dict notation interchangeably (`choice["message"]` or `choice.message`).

rogthefrog (approved):
Nice! I bet this one was a bear to debug.

wpietri (approved):
Thanks for sorting this out.

PR #664, opened 2024-11-01
More ops fixes

wpietri:
Add a little more journaling.
Make sure our external resources are live before starting a massive run.

### DIFF
--- a/src/modelbench/benchmark_runner.py
+++ b/src/modelbench/benchmark_runner.py
@@ -150,7 +150,7 @@ class TestRunBase:
         self.test_annotators[test.uid] = annotators
 
     def add_finished_item(self, item: "TestRunItem"):
-        if item.completion() and item.annotations and not item.fatal_exceptions:
+        if item.completion() and item.annotations and not item.exceptions:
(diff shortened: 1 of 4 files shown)


src/modelbench/benchmark_runner.py
```diff
@@ -485,7 +552,9 @@ def _check_ready_to_run(self):
 
     def run(self) -> BenchmarkRun:
         self._check_ready_to_run()
```
dhosterman, line 554: Do you want to collect the RuntimeErrors from both annotators and suts here so that all issues can be addressed simultaneously? If I'm understanding correctly, right now, it won't check the annotators if there's an issue with the suts.

dhosterman (approved):
👍

src/modelbench/benchmark_runner.py
wpietri: I was thinking to keep them separate, as right now our annotators can take up to 15 minutes to decide whether or not they're ready. This will get people quick notification of a SUT problem. But if we can get off HuggingFace to something faster, then that would make sense to me.

PR #670, opened 2024-11-05
Practice prompt calibration

wpietri:


src/modelbench/run.py
```diff
@@ -269,23 +269,25 @@ def calibrate(update: bool, file) -> None:
...
         print("Can't load private annotators needed for calibration")
         exit(1)
 
-    benchmarks = [GeneralPurposeAiChatBenchmark()]
-    for l in Locale:
+    benchmarks = []
+    for l in [Locale.EN_US]:
         benchmarks.append(GeneralPurposeAiChatBenchmarkV1(l, "ensemble"))
-    run_result = run_benchmarks_for_suts(benchmarks, reference_suts, 100)
+    run_result = run_benchmarks_for_suts(benchmarks, reference_suts, 2000)
```
bkorycki, line 282: I'm curious about the reasoning behind 2000 max items

src/modelbench/run.py
```diff
@@ -269,23 +269,25 @@ def calibrate(update: bool, file) -> None:
...
+        ModelGaugeSut.for_key("llama-3.1-8b-instruct-turbo-together"),
     ]
     if not ensure_ensemble_annotators_loaded():
         print("Can't load private annotators needed for calibration")
         exit(1)
 
-    benchmarks = [GeneralPurposeAiChatBenchmark()]
-    for l in Locale:
+    benchmarks = []
+    for l in [Locale.EN_US]:
```
bkorycki, line 280: I might have missed this, but is 1.0 only going to include english prompts?

bkorycki (commented):
Thanks for doing this!

src/modelbench/run.py
wpietri: Ah, sorry, I shouldn't have left that in. The calibration set should be all the items, but I was doing practice runs with smaller numbers. 2000 is just larger than the 1000 max. I'll change it so it's always everything.

src/modelbench/run.py
wpietri: Yup. @bollacker knows more, but there were a number of challenges, so new languages will come shortly after 1.0.

bkorycki (approved):

bollacker (approved):
I see nothing obviously wrong.

src/modelbench/run.py
bkorycki: Ah ok, sounds good

dhosterman (approved):
👍

PR #672, opened 2024-11-08
Add some basic journal documentation

wpietri:
No feedback on the journal docs, so I guess we'll merge this and then add to it later.


bkorycki (approved):
Yeah this looks good to me! It's been helpful in writing the consistency checker, so I think it serves its purpose well.

PR #680, opened 2024-11-13
Journal consistency checker

bkorycki:
I don't particularly like this. I think the class hierarchy is too complicated. My general goal was to make it easy to 1) add new checks and 2) understand what each check does. 

I didn't put any effort towards efficiently querying the journal, so there's room for optimization there.
I only added a handful of checks to demonstrate what this looks like. Once we decide on the approach we want to take here, I will add some more easy checks + some basic unit testing with sample run journals.

Here is what the output of a sample run looks like:
```
Results for Test x SUT level checks
Entity                           EachPromptRespondedToOnce    EachResponseTranslatedOnce    EachItemMeasuredOnce
-------------------------------  ---------------------------  ----------------------------  ----------------------
demo_yes_no, safe-vcr-en_us-1.0  ✅                            ✅                             ✅
demo_yes_no, safe-ncr-en_us-1.0  ✅                            ✅                             ✅
demo_yes_no, safe-src-en_us-1.0  ✅                            ✅                             ✅
demo_yes_no, safe-cse-en_us-1.0  ❌                            ✅                             ✅
demo_yes_no, safe-dfm-en_us-1.0  ✅                            ✅                             ✅
demo_yes_no, safe-spc-en_us-1.0  ✅                            ✅                             ✅
demo_yes_no, safe-prv-en_us-1.0  ✅                            ✅                             ✅
demo_yes_no, safe-ipv-en_us-1.0  ✅                            ✅                             ✅
demo_yes_no, safe-iwp-en_us-1.0  ✅                            ✅                             ✅
demo_yes_no, safe-hte-en_us-1.0  ✅                            ✅                             ✅
demo_yes_no, safe-ssh-en_us-1.0  ✅                            ✅                             ✅
demo_yes_no, safe-sxc-en_us-1.0  ✅                            ✅                             ✅

Results for Test x SUT x Annotator checks
Entity                                          EachResponseAnnotatedOnce
----------------------------------------------  ---------------------------
demo_yes_no, llama_guard_2, safe-vcr-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-ncr-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-src-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-cse-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-dfm-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-spc-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-prv-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-ipv-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-iwp-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-hte-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-ssh-en_us-1.0  ✅
demo_yes_no, llama_guard_2, safe-sxc-en_us-1.0  ✅

------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Failed checks for Test x SUT level checks:
EachPromptRespondedToOnce: Expected exactly 1 SUT response for each prompt in the test.
        The following duplicate prompts were found: ['airr_practice_1_0_22916']
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
All Test x SUT x Annotator checks checks passed!
```

### DIFF
--- a/src/modelbench/consistency_checker.py
+++ b/src/modelbench/consistency_checker.py
@@ -100,8 +100,6 @@ class OneToOneCheck(JournalCheck):
 
 class EachPromptRespondedToOnce(OneToOneCheck):
     def __init__(self, search_engine: JournalSearch, sut, test):
-        print("Test prompts:", search_engine.test_prompt_uids(test))
-        print("SUT responses:", search_engine.sut_response_prompt_uids_for_test(sut, test))
         super().__init__(
             search_engine.test_prompt_uids(test), search_engine.sut_response_prompt_uids_for_test(sut, test)
         )
@@ -138,6 +136,25 @@ class EachItemMeasuredOnce(OneToOneCheck):
 
     def failure_message(self) -> str:
         message = "Expected every prompt-response to be measured exactly once.\n"
+        # Call super() to get specific details about duplicates/missing/extra prompts.
+        return message + super().failure_message()
+
+
+class EachResponseAnnotatedOnce(OneToOneCheck):
+    def __init__(self, search_engine: JournalSearch, sut, test, annotator):
+        self.annotator = annotator
+        self.sut = sut
+        self.test = test
+        translated_responses = search_engine.query("translated sut response", sut=sut, test=test)
+        cached_annotations = search_engine.query(
+            "using cached annotator response", sut=sut, test=test, annotator=annotator
+        )
+        fetched_annotations = search_engine.query("fetched annotator response", sut=sut, test=test, annotator=annotator)
+        all_prompts = [response["prompt_id"] for response in cached_annotations + fetched_annotations]
+        super().__init__([response["prompt_id"] for response in translated_responses], all_prompts)
+
+    def failure_message(self) -> str:
+        message = f"Expected exactly 1 {self.annotator} annotation for each response from SUT {self.sut} in test {self.test}\n\t"
         # Call super() to get specific details about duplicates/missing/extra prompts.
         return message + super().failure_message()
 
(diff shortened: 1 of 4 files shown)


wpietri:
I think you're making progress for sure. I love the use of emojis in the output. I think the columns are quite wide, so maybe at some point it's worth breaking the LongCamelCaseWords in short words that wrap on the spaces, but that can come later.

In your shoes I'd probably try to find some object structure that I was more happy with before adding more examples, but if nothing occurs to you than maybe more examples will help show you where you can simplify.

And I think it's fine if this is not very fast. If it takes 10-30s to process a full benchmark run, that's ok. Especially once we automate this as part of a benchmark run, adding another 30s to a half-hour process won't be noticeable. Past that point we can look at speedups, but I suspect a little manual indexing with dicts will be plenty. 

Given the time pressure, I also think it's fine to merge something incomplete now, use it where we can, and add to if after we get all the SUTs in. We also have my first-pancake version, which I wouldn't want to extend, but it does check a bunch of things.

wpietri (approved):
Seems like a step forward to me.

src/modelbench/consistency_checker.py
```diff
@@ -0,0 +1,304 @@
...
+from itertools import product
+from tabulate import tabulate
+from typing import Dict, List
+
+from modelbench.run_journal import journal_reader
+
+LINE_WIDTH = os.get_terminal_size().columns
+
+
+class JournalSearch:
```
rogthefrog, line 14: I really like this approach.

rogthefrog (approved):
Thanks for tackling this!

src/modelbench/consistency_checker.py
bkorycki: Thanks! I think this will make it easy to independently optimize the journal querying later on.

PR #693, opened 2024-11-18
More consistent retrying

wpietri:
Some SUTs didn't retry at all; others have built-in retry that's set pretty low. Let's see how this does in production.


bkorycki (approved):
I didn't try it but it looks good to me.

PR #696, opened 2024-11-18
Use actual heldback prompts in official tests

bkorycki:


wpietri (approved):
Hooray! I think that's the final piece!

PR #702, opened 2024-11-21
Fix annotator cache bug in benchmark runner

bkorycki:
Note: we should come back to this and create separate caches for each annotator (and SUT?).


wpietri (approved):
Looks great!

PR #707, opened 2024-11-21
Register Phi 3.5 moe SUT + Add "instruct" to Phi UIDs

bkorycki:


plugins/azure/modelgauge/suts/azure_client.py
```diff
@@ -140,9 +140,21 @@ def scope(cls) -> str:
...
     AzureChatSUT,
-    "phi-3.5-mini",
+    "phi-3.5-mini-instruct",
     "https://Phi-3-5-mini-instruct-hfkpb.eastus2.models.ai.azure.com",
     InjectSecret(PhiMiniKey),
 )
+SUTS.register(
+    AzureChatSUT,
+    "phi-3.5-moe-instruct",
+    "https://Phi-3-5-MoE-instruct-abneb.eastus2.models.ai.azure.com",
```
rogthefrog, line 158: Do we want to make this a secret? That way it's not hardcoded.

bkorycki: I think it's fine to hardcode unless its actually a "secret"

wpietri: We really need to run these tests, so let's hardcode it for now. In the future we should talk about config that isn't secrets.

wpietri (approved):

wpietri:
And sorry to be rude, but I'm merging this so I can build the hopefully-final docker image.

PR #712, opened 2024-11-23
Tweak the grading function

rogthefrog:
This primarily replaces the hardcoded "30% unsafe is the worst possible score" logic with a cutoff based on the reference score.


bollacker (approved):
Have not run it.  Do not see anything obviously wrong.

wpietri (approved):

PR #729, opened 2024-12-05
Benchmark scoring fix.

wpietri:
Fix the problem where benchmark scoring combines official and practice prompts for the standard. Now official and practice each have their own standards.

This shifts responsibility for  knowing a benchmark's standard to the BenchmarkDefinition, in parallel with how it works for hazards. Between this and the 0.5 legacy stuff this leaves a lot of dead or deprecated code, but I didn't want to mess with too much for an emergency fix.

### DIFF
--- a/src/modelbench/benchmarks.py
+++ b/src/modelbench/benchmarks.py
@@ -1,14 +1,15 @@
+import statistics
 from abc import ABC, abstractmethod
 from datetime import datetime
 from typing import List, Sequence
 
 import casefy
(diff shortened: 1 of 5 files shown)


dhosterman (approved):
👍 

I think this makes sense to me, but I'm not in a position to thoroughly dig into it at the moment.

rogthefrog (approved):
Surprisingly straightforward. Thanks!

Issue #733, opened 2024-12-06
Update disclaimer text referencing version 0.5


rogthefrog:
The old copy is still present in some places. 

```Version 0.5 of this benchmark is a proof of concept only. Results are not intended to indicate actual levels of AI system safety.```

- docs/add-a-sut.md 
- src/modelbench/templates/content/general.toml 
- src/modelbench/templates/content_mlc/general.toml    

This language shows up in job logs, for example.

rogthefrog:
* Remove all disclaimer copy

rogthefrog:
Started but paused until Officially Vetted :tm: requirements are in. Need Kurt/Peter approval.

PR #735, opened 2024-12-06
update version string from 0.5 to 1.0 after release

rogthefrog:
# WIP

## Needs some clarification about a few things

* assets used by the static site generator: ignore until the site generator is removed?
* modelbench/src/modelbench/templates/content_mlc/general_purpose_ai_chat_benchmark.toml and a couple other files have scopes named `general_purpose_ai_chat_benchmark-0_5`. Should those be replaced altogether?

https://github.com/mlcommons/modelbench/issues/733


rogthefrog:
Closing until we have a group discussion about how this should be handled.

PR #740, opened 2024-12-10
Mistral SUTs with moderation

bkorycki:
Note: This changes the cache keys for the already existing mistral SUTs. So the old caches won't work for them.

### DIFF
--- a/plugins/mistral/modelgauge/suts/mistral_sut.py
+++ b/plugins/mistral/modelgauge/suts/mistral_sut.py
@@ -164,32 +164,28 @@ class MistralAIModeratedSut(PromptResponseSUT):
         return SUTResponse(completions=[SUTCompletion(text=str(safest_completion))])
 
 
-MISTRAL_API_KEY = InjectSecret(MistralAIAPIKey)
+def register_suts_for_model(model_name):
+    MISTRAL_API_KEY = InjectSecret(MistralAIAPIKey)
+    # Register standard SUT.
+    model_uid = f"mistralai-{model_name}"
+    SUTS.register(MistralAISut, model_uid, model_name, MISTRAL_API_KEY)
+    # Register moderated SUT.
+    moderation_model_name = "mistral-moderation-latest"
+    num_generations = 5
+    temperature = 0.3
+    threshold = 0.2
+    moderated_model_uid = f"mistralai-{model_name}-moderated"
+    SUTS.register(
+        MistralAIModeratedSut,
+        moderated_model_uid,
+        model_name,
+        moderation_model_name,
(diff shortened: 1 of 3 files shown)


plugins/mistral/modelgauge/suts/mistral_sut.py
```diff
@@ -73,17 +73,119 @@ def translate_response(self, request: MistralAIRequest, response: MistralAIRespo
...
+
+    @property
+    def client(self):
+        if not self._client:
+            self._client = MistralAIClient(self.model_name, self._api_key)
+        return self._client
+
+    def translate_text_prompt(self, prompt: TextPrompt) -> MistralAIRequest:
+        # Warn if prompt options are overridden.
+        if prompt.options.temperature is not None and prompt.options.temperature != self.temperature:
```
dhosterman, line 119: Would this be clearer and have the same functionality if you did something like:

`if prompt.options.temperature and ...`?

Same for ones below?

plugins/mistral/modelgauge/suts/mistral_sut.py
```diff
@@ -73,17 +73,119 @@ def translate_response(self, request: MistralAIRequest, response: MistralAIRespo
...
+            scores = response.moderations[choice.index]
+            assert len(scores.results) == 1, f"Expected 1 moderation object, got {len(scores.results)}."
+            category_scores = scores.results[0].category_scores
+            max_score = max(category_scores.values())
+            scores_list.append((text, max_score))
+
+        sorted_responses = sorted(scores_list, key=lambda x: x[1])
+        lowest_score = sorted_responses[0][1]
+        if lowest_score >= self.threshold:
+            safest_completion = "I'm sorry I cannot assist with this request."
```
dhosterman, line 161: I'm probably missing something obvious here, but why are we hard coding a response here?

plugins/mistral/modelgauge/suts/mistral_sut.py
bkorycki: No because `0.0` evaluates to `False`!

plugins/mistral/modelgauge/suts/mistral_sut.py
bkorycki: This was in the script that Mistral gave us.

dhosterman: I think it's ridiculous, but it's nothing to do with your code!

rogthefrog: At least they had the self-awareness not to use the classic HAL response.

PR #743, opened 2024-12-11
Remove 0.5 code

bkorycki:
TODOs:
- Remove 0.5 code from scoring logic (TBD in a branching PR by @rogthefrog )
- Remove static site generator and replace with a simpler report. I left it for now just so that there is some local output when you do a run.


tests/modelbench_tests/test_benchmark_grading.py
```diff
@@ -5,9 +5,9 @@
 import pytest
 
 from modelbench.benchmarks import BenchmarkScore, GeneralPurposeAiChatBenchmarkV1, Locale
-from modelbench.hazards import HazardScore, SafeCaeHazard, SafeCbrHazard, SafeHatHazard, Standards
+from modelbench.hazards import HazardScore, SafeHazardV1, Standards
```
rogthefrog, line 8: Do you know the original motivation for having one individual subclass for each hazard?

rogthefrog (approved):
👍🏻

tests/modelbench_tests/test_benchmark_grading.py
bkorycki: No, I don't

PR #747, opened 2024-12-13
remove old grading function implementations

rogthefrog:


src/modelgauge/main.py
```diff
@@ -207,7 +200,7 @@ def run_test(
     secrets = load_secrets_from_config()
     # Check for missing secrets without instantiating any objects
     test_cls = TESTS._get_entry(test).cls
-    annotators = test_cls.get_annotators()
+    annotators = test_cls.get_annotators(test_cls)
```
bkorycki, line 203: why add this arg? I think this is the cause of some of the failing tests

plugins/mistral/tests/test_mistral_sut.py
```diff
@@ -41,7 +41,7 @@ def response():
 
 @pytest.fixture
 def sut():
-    return MistralAISut("ministral-8b-latest", "ministral-8b-latest", "latest", MistralAIAPIKey("fake"))
+    return MistralAISut("ministral-8b-latest", "ministral-8b-latest", MistralAIAPIKey("fake"))
```
bkorycki, line 44: Same here.

src/modelgauge/main.py
```diff
@@ -1,40 +1,33 @@
...
 import pathlib
 import warnings
 from typing import List, Optional
 
+import click
+
 from modelgauge.annotator import CompletionAnnotator
 from modelgauge.annotator_registry import ANNOTATORS
 from modelgauge.base_test import PromptResponseTest
 from modelgauge.command_line import (
```
bkorycki, line 12: I kind of like it when constants are grouped together at the beginning of an import. This is just a preference though,

bkorycki (approved):
Looks good! thanks

src/modelgauge/main.py
rogthefrog: The changes here are done automatically by the standard rules in the Black linter. I can tweak them if you'd like.

bkorycki: Hm, I also use black but it doesn't format stuff this way for me. Maybe we have different configurations?

PR #749, opened 2024-12-13
fix how AI is capitalized

rogthefrog:
It's the little things.


dhosterman (approved):
As much as I like "Ai", I guess this is fine.

Issue #750, opened 2024-12-13
Run calibration for French prompts


rogthefrog:

rogthefrog:
https://github.com/mlcommons/modelbench/pull/751

PR #751, opened 2024-12-13
reference scores for an initial set of fr_fr practice prompts

rogthefrog:
https://github.com/mlcommons/modelbench/issues/750


dhosterman (approved):
👍

PR #754, opened 2024-12-16
Modelbench SUT cleanup + testing infra improvements

bkorycki:
These are some initial improvements along the road to removing the SUT wrapper class in modelbench. I'm splitting the changes up into smaller PRs to make the more manageable to review.

Changes in this PR:
- **Get rid of DEFAULT_SUTS list,** which was leftover from v0.5. The SUTs in that list aren't really that relevant, and I think it's reasonable to assume + expect that user will specify SUT uids at the command line.
- Relatedly, you now **must set at least 1 `--sut <uid>` arg when running the CLI.**
- **Modelbench does not register SUTs**.  I think we should keep SUT registration in modelgauge in order to avoid registering an identical SUT to multiple UIDs. Also, none of those SUTs that modelbench registered were used in v1.0 anyway. 
- **Testing infra improvements.** I created a conftest file to register a fake SUT during setup. This removes a lot of code duplication in the tests (i.e. creating + registering dummy SUTs). And the centralization will make it easier to fully murder the ModelGaugeSut wrapper.


src/modelbench/run.py
```diff
@@ -189,25 +189,17 @@ def consistency_check(journal_path, verbose):
...
-        suts = DEFAULT_SUTS
+def find_suts_for_sut_argument(sut_uids: List[str]):
+    suts = []
+    registered_sut_keys = set(i[0] for i in SUTS.items())
+    for sut_uid in sut_uids:
+        if sut_uid in registered_sut_keys:
+            suts.append(ModelGaugeSut.for_key(sut_uid))
+        else:
+            raise click.BadParameter(
+                f"Unknown uid '{sut_uid}'. Valid options are {sorted(registered_sut_keys, key=lambda x: x.lower())}",
```
rogthefrog, line 200: Would it be a pain to have the exception print out the SUTs one per line, instead of a long list? That would make it easier to fix your CLI invocation.

rogthefrog (changes requested):
Having an easy-to-read sut ID list would be super nice.

src/modelbench/run.py
bkorycki: Sure, is that better?

rogthefrog: Yay! Thank you.

rogthefrog (approved):

PR #758, opened 2024-12-17
Remove SUT wrapper

bkorycki:
Review this PR after PR #754. 
When everything is approved, this PR should be merged first, followed by #754, #747, and then #743.

Summary: This removes the modelbench's `ModelGaugeSut` wrapper class, which adds an unnecessary layer of abstraction.


src/modelbench/run.py
```diff
@@ -190,18 +191,27 @@ def consistency_check(journal_path, verbose):
 
 
 def find_suts_for_sut_argument(sut_uids: List[str]):
```
rogthefrog, line 193: Maybe this function should be named `find_suts_for` to match other functions?

src/modelbench/run.py
```diff
@@ -190,18 +191,27 @@ def consistency_check(journal_path, verbose):
...
-        else:
-            valid_suts = sorted(registered_sut_keys, key=lambda x: x.lower())
-            valid_suts_str = "\n\t".join(valid_suts)
-            raise click.BadParameter(
-                f"Unknown uid '{sut_uid}'. Valid options are: {valid_suts_str}",
-                param_hint="sut",
-            )
+        try:
+            missing_secrets.extend(SUTS.get_missing_dependencies(sut_uid, secrets=secrets))
+            suts.append(SUTS.make_instance(sut_uid, secrets=secrets))
```
rogthefrog, line 203: A thought, not for this PR: while I'd normally be in favor of the client code injecting dependencies into the objects it creates (`secrets` in this case) because it makes testing easier, in this case it feels like the SUT should know how to configure itself, and the client code shouldn't need to. WDYT?

src/modelbench/run.py
```diff
@@ -190,18 +191,27 @@ def consistency_check(journal_path, verbose):
...
+        except KeyError:
+            unknown_uids.append(sut_uid)
+    if len(unknown_uids) > 0:
+        valid_suts = sorted(SUTS.keys(), key=lambda x: x.lower())
+        valid_suts_str = "\n\t".join(valid_suts)
```
rogthefrog, line 208: 🥇

src/modelbench/run.py
```diff
@@ -272,10 +281,8 @@ def print_summary(benchmark, benchmark_scores, anonymize):
         counter = 0
         for bs in benchmark_scores:
             counter += 1
-            key = f"sut{counter:02d}"
-            name = f"System Under Test {counter}"
-
-            bs.sut = FakeSut(key, name)
+            uid = f"sut{counter:02d}"
+            bs.sut = AnonSUT(uid)
```
rogthefrog, line 285: What is this for?

tests/modelbench_tests/test_benchmark_runner.py
```diff
@@ -121,6 +122,7 @@ def teardown_class(cls):
                 del ANNOTATORS._lookup[uid]
         cls._original_registered_annotators = None
 
+    # TODO: Remove? I think this any setup should be handled in conftest.
```
rogthefrog, line 125: Agreed.

tests/modelbench_tests/test_run.py
```diff
@@ -162,11 +165,13 @@ def test_benchmark_multiple_suts_produces_json(self, runner, version, locale, pr
             version, locale if locale else Locale.EN_US, prompt_set if prompt_set else "practice", "default"
         )
 
+        FakeSUT
```
rogthefrog, line 168: What's this?

rogthefrog (changes requested):
Other than the question around line 168 in tests/modelbench_tests/test_run.py 👍🏻

src/modelbench/run.py
bkorycki: Which other functions?

src/modelbench/run.py
bkorycki: It's to anonymize results if someone passes the `--anonymize` option. I don't know if it's actually used, but I don't want to get rid of it because it might be important.

tests/modelbench_tests/test_run.py
bkorycki: Oops, that shouldn't be there. Thanks for catching it.

src/modelbench/run.py
bkorycki: Dependency injection allows us to know all of the required secrets at the beginning of a run + cleanly exit and report missing secrets without having to initialize any objects, which might be expensive. I believe those were the guiding principles that lead us to this paradigm.

tests/modelbench_tests/test_benchmark_runner.py
bkorycki: Ah thanks for reminding me. I just tested it and it's actually not necessary, so I removed it.

dhosterman (approved):
I think this looks like a solid stab at getting rid of some of that 0.5 code!

PR #763, opened 2024-12-18
Fix secrets bug in SafeTest

bkorycki:
This fixes a bug that requires you to have the secret prompts token even when running the public practice prompts. I believe this will also address some issues we're encountering in the smoke tests.


rogthefrog (approved):
👍🏻

PR #768, opened 2024-12-20
simplify locale handling by using lowercase strings everywhere

rogthefrog:
The current locale implementation uses an enum, which is handy for validation, but it isn't used consistently throughout the code. Some functions declare a locale argument of the Locale type, but in actuality sometimes they get plain strings, so the function code has to 1) check the type 2) validate whether the string is a valid locale. The Locale object itself is just a bag of strings. 

@bkorycki, @bollacker and I discussed simplifying and using strings instead.


src/modelgauge/tests/safe_v1.py
```diff
@@ -155,9 +140,8 @@ def _check_annotators(cls):
 
     def __localize_filename(self) -> str:
         # TODO implement a better standard for prompt file names by locale
-        locale = self.locale.value.lower()
-        if locale and locale != "en_us":
-            tail = f"_{locale}"
+        if self.locale and self.locale != EN_US:
```
bkorycki, line 143: `self.locale` should never be None, right?

bkorycki (changes requested):
I think this is a solid improvement! Thanks.

src/modelgauge/tests/safe_v1.py
rogthefrog: That is true as long as someone doesn't set `self.locale` outside the constructor. It's a remote possibility, though, and should probably error out if someone does that, so I'll remove the guard clause.

bkorycki (approved):

PR #776, opened 2025-01-02
fix spurious sample_cache.sqlite changes

wpietri:
The sample_cache.sqlite file was often changed for mysterious reasons. This fixes the problem and removes the footgun that was causing it.


bkorycki (approved):
Thanks for fixing this!

rogthefrog (approved):

PR #783, opened 2025-01-09
Add list-annotators command

bkorycki:
This adds `list-annotators` to the CLI, which is complementary to the already existing `list-suts` and `list-tests`. This command is useful for debugging.


rogthefrog (approved):
Handy!

PR #787, opened 2025-01-13
Add pypi test back

rogthefrog:
This properly belongs with modelbench. We should be alerted if a Pypi build is broken.

https://github.com/mlcommons/modelbench/issues/790


.github/workflows/python-app.yml
```diff
@@ -21,64 +21,52 @@ jobs:
...
+          api_key = "fake"
 
           [nvidia-nim-api]
-          api_key = "${{ secrets.NVIDIA_API_KEY }}"
+          api_key = "fake"
 
           [demo]
           api_key="12345"
 
+          [modellab_files]
```
wpietri, line 61: What's this token have access to?

wpietri (approved):
Good idea. As long as it really needs the secret, I'm fine with it.

.github/workflows/python-app.yml
wpietri: In that case, i'd rather remove this unless there's a strong need. The official prompts shouldn't be used by anybody downloading this from PyPi, and any leak of that token would be very expensive, as the best response would be to create all new leaked prompts.

PR #788, opened 2025-01-13
Modelgauge registers the ensemble-based tests.

bkorycki:
- Ensemble-based tests are now registered in modelgauge, not modelbench.
- Can run these tests directly in modelgauge CLI now.
- A test is only ever registered once.
- Remove secrets from annotator sets. They weren't doing anything.

### DIFF
--- a/tests/modelgauge_tests/test_default_annotator_set.py
+++ b/tests/modelgauge_tests/test_default_annotator_set.py
@@ -1,16 +1,15 @@
 import pytest
 from unittest.mock import MagicMock
 from modelgauge.default_annotator_set import DefaultAnnotatorSet
-from modelgauge.suts.together_client import TogetherApiKey
 
 
 def test_constructor():
-    annotators = DefaultAnnotatorSet(secrets={"together_api_key": TogetherApiKey("fake")})
+    annotators = DefaultAnnotatorSet()
     assert len(annotators.annotators) == 1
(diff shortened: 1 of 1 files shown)


wpietri (approved):
Another nice cleanup. Thanks!

src/modelgauge/tests/safe_v1.py
```diff
@@ -249,29 +249,27 @@ def register_tests(cls, evaluator=None):
...
+# Register tests with "default" annotator.
+register_tests(SafeTestVersion1)
+# Try to register test with private "ensemble" annotator.
+try:
+    from modelgauge.private_ensemble_annotator_set import PRIVATE_ANNOTATOR_SET
 
-        print(f"unexpected failure registering annotators for {uid_key} and {private_annotators}")
-        raise
+    register_private_annotator_tests(PRIVATE_ANNOTATOR_SET, "ensemble")
+except Exception as e:
```
rogthefrog, line 274: Do we want to provide feedback about the failure to load the ensemble?

rogthefrog (approved):
👍🏻

src/modelgauge/tests/safe_v1.py
bkorycki: Users don't have access to the ensemble in the standard use case. I think the only place that feedback is necessary is when you try to request the benchmark with the ensemble evaluator in modelbench.

Issue #781, opened 2025-01-07
Improve how locals and prompt file name interact with each other


rogthefrog:
The safe test downloads prompt files from modellab. It supports multiple locales in a given file, but doesn't do well with different files for different locales. 

Workarounds exist, but they're hacky.

Options:

* Improve file naming conventions to support locales. Proposal: if a file includes one locale, include that locale in the filename. English prompt files don't include en_us and that makes the code messy.
* Ensure that we don't try to make test items in locale X using a file that does not include locale X (e.g. in the smoke test).
* Handle file not found errors better, especially in unit or smoke tests.
* If any unit tests rely on downloading an asset from modellab, replace that with a local file.

References:

[Workaround for test items](https://github.com/mlcommons/modelbench/pull/780)

[Workaround for filenames with and without locales](https://github.com/mlcommons/modelbench/pull/744)

rogthefrog:
https://github.com/mlcommons/modelbench/pull/801

PR #801, opened 2025-01-16
Decouple the prompt set type (practice, official) from its file name

rogthefrog:
# Story 
https://github.com/mlcommons/modelbench/issues/781

# Rationale

## Filenames
https://github.com/mlcommons/modelbench/pull/744

## Test Items
https://github.com/mlcommons/modelbench/pull/780

# Main Change

New uploaded files need to be added to the `PROMPT_SETS` dict in `src/modelgauge/prompt_sets.py` and they need to contain the whole key. E.g. if you want to use the key "demo_hi_in" then your prompt file name should include "demo_hi_in" in it.

To use new prompt sets in modellab, you need to add them to the modellab drop down, and to the modelrunner list of prompt sets. We may want to change that so it's less manual.

### DIFF
--- a/src/modelgauge/locales.py
+++ b/src/modelgauge/locales.py
@@ -9,6 +9,8 @@ DEFAULT_LOCALE = "en_us"
 
 # add the other languages after we have official and practice prompt sets
 LOCALES = (EN_US, FR_FR)
+# all the languages we have official and practice prompt sets for
+PUBLISHED_LOCALES = (EN_US,)
 
 
 def is_valid(locale: str) -> bool:
--- a/src/modelgauge/prompt_sets.py
+++ b/src/modelgauge/prompt_sets.py
@@ -25,13 +25,13 @@ TEST_PROMPT_SETS = {
 PROMPT_SET_DOWNLOAD_HOST = "ailuminate.mlcommons.org"
 
 
-def prompt_set_file_base_name(prompt_set: str) -> str:
-    filename = PROMPT_SETS.get(prompt_set, TEST_PROMPT_SETS.get(prompt_set, ""))
+def prompt_set_file_base_name(prompt_set: str, prompt_sets: dict = PROMPT_SETS) -> str:
+    filename = prompt_sets.get(prompt_set, None)
     return filename
 
 
-def validate_prompt_set(prompt_set: str) -> bool:
-    filename = prompt_set_file_base_name(prompt_set)
+def validate_prompt_set(prompt_set: str, prompt_sets: dict = PROMPT_SETS) -> bool:
+    filename = prompt_set_file_base_name(prompt_set, prompt_sets)
     if not filename:
-        raise ValueError(f"Invalid prompt set {prompt_set}. Must be one of {PROMPT_SETS.keys()}.")
+        raise ValueError(f"Invalid prompt set {prompt_set}. Must be one of {prompt_sets.keys()}.")
     return True
--- a/src/modelgauge/prompt_sets.py
+++ b/src/modelgauge/prompt_sets.py
@@ -1,0 +1,37 @@
+from modelgauge.secret_values import RequiredSecret, SecretDescription
+
+
+class ModellabFileDownloadToken(RequiredSecret):
+    @classmethod
+    def description(cls) -> SecretDescription:
+        return SecretDescription(
+            scope="modellab_files",
+            key="token",
+            instructions="Please ask MLCommons admin for permission.",
+        )
+
+
+# file name format:
+# {prefix}_{version}_{type}(_{locale})_prompt_set_release
+PROMPT_SETS = {
+    "practice": "airr_official_1.0_practice_prompt_set_release",
+    "official": "airr_official_1.0_heldback_prompt_set_release",
+    "practice_fr": "airr_official_1.0_practice_fr_fr_prompt_set_release",
+}
+TEST_PROMPT_SETS = {
+    "fake-prompts": "fake-prompts",
+}
+
+PROMPT_SET_DOWNLOAD_HOST = "ailuminate.mlcommons.org"
+
+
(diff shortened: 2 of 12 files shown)


src/modelgauge/prompt_sets.py
```diff
@@ -0,0 +1,37 @@
...
+
+# file name format:
+# {prefix}_{version}_{type}(_{locale})_prompt_set_release
+PROMPT_SETS = {
+    "practice": "airr_official_1.0_practice_prompt_set_release",
+    "official": "airr_official_1.0_heldback_prompt_set_release",
+    "practice_fr_fr": "airr_official_1.0_practice_fr_fr_prompt_set_release",
+}
+TEST_PROMPT_SETS = {
+    "fake-prompts": "fake-prompts",
```
bkorycki, line 22: Maybe this should be in the unit test code if its only used by unit tests? What do you think?

src/modelgauge/prompt_sets.py
```diff
@@ -0,0 +1,37 @@
...
+}
+
+PROMPT_SET_DOWNLOAD_HOST = "ailuminate.mlcommons.org"
+
+
+def prompt_set_file_base_name(prompt_set: str) -> str:
+    filename = PROMPT_SETS.get(prompt_set, TEST_PROMPT_SETS.get(prompt_set, ""))
```
bkorycki, line 29: Hm, maybe this should loudly fail if it gets an unknown prompt set instead of returning an empty string. I think it might be hard to debug otherwise. Or maybe it could return None instead? So that its more obvious that there's not corresponding file name for the prompt set?

bkorycki (changes requested):
Looks much better organized now. Thanks!

rogthefrog:
> Would this code get cleaner if there were a PromptSet object and possibly an enum for the known good ones?

That was my initial approach, and it got more complicated than I wanted.

src/modelgauge/prompt_sets.py
rogthefrog: I agree this isn't pretty. The main obstacle is that the objects being tested (e.g. the safe test v1) have direct access to the shared global `PROMPT_SETS` structure (for validation and file names). The string values in that structure filter down to the dependency checker. Adding an arbitrary prompt set name and filename in tests no longer works (it gets rejected in the dependency checker), which is the desired effect (we only want to accept blessed prompt sets), but it may go too far in that direction. 

A cleaner way would be to pass the blessed prompt sets around instead of sharing a global structure, DI-style. Maybe I should do that!

rogthefrog:
@bkorycki @wpietri @dhosterman ready for re-review when you have a chance!

wpietri:
> > Would this code get cleaner if there were a PromptSet object and possibly an enum for the known good ones?
> By contrast, the prompt sets (their short names and filenames) are user-defined, unbounded, and arbitrary. It feels more right (to me) to use a bag of strings to represent a bag of strings (short names and filenames).

To me an object is a name plus data plus behavior. I think we have those conditions here, if marginally. The name is already in the filename, and it's a domain term. You have the data in the dict. And there's a bit of behavior in the file itself, with possibly more elsewhere. But it is pretty basic, so I can live with it staying as procedural programming until more behavior turns up.

But this is a problem for me:

```
class GeneralPurposeAiChatBenchmarkV1(BenchmarkDefinition):
    def __init__(self, locale: str, prompt_set: str, prompt_sets: dict = PROMPT_SETS, evaluator: str = "default"):
        validate_locale(locale)
        validate_prompt_set(prompt_set, prompt_sets)
[...]
    def _make_hazards(self) -> Sequence[HazardDefinition]:
        return [
            SafeHazardV1(hazard_key, self.locale, self.prompt_set, PROMPT_SETS, self.evaluator)

[...]
```

Previously, what defined a GeneralPurposeAiChatBenchmarkV1 was a locale, a prompt set, and an evaluator. The code more or less matched our domain language. But now PROMPT_SETS is part of the definition, which is definitely outside our domain language, and it's weird from an object modelling perspective.

What typing would get us here is a clearer interface, and using the type system to do the validation that ends up here. Even going with strings, I'd rather validation of UI/CLI inputs end up much closer to the UI/CLI. (If we wanted to be extra careful here and have a double-check, I'd rather have it be something like `assert prompt_set.valid_set_id(prompt_set)`. But as long as things will blow up harmlessly with the wrong prompt set key, I wouldn't go that far.) And PROMPT_SETS definitely shouldn't be part of the constructor signature for a benchmark or a hazard; that's an implementation detail distorting the domain model.

rogthefrog:
@wpietri @bkorycki third pass :)

@wpietri I still think using a bag of strings for prompt sets is preferable at this stage. There's a world in which we may need to add functionality or data to the concept of a prompt set, e.g. a locale if it only contains one locale, and then we can use a class for that. But I don't think we're there yet, and because types aren't enforced and we need to support plain strings to identify prompt sets (e.g. as arguments in the CLI) then I think it's best to stick with strings at this stage.

wpietri (approved):
Thanks for persisting!

src/modelgauge/locales.py
```diff
@@ -9,6 +9,8 @@
 
 # add the other languages after we have official and practice prompt sets
 LOCALES = (EN_US, FR_FR)
+# all the languages we have official and practice prompt sets for
+PUBLISHED_LOCALES = (EN_US,)
```
bkorycki, line 13: Is this the set of locales that we can run in modelbench?

bkorycki (approved):
Looks good!

src/modelgauge/locales.py
rogthefrog: LOCALES should be runnable in modelbench. PUBLISHED_LOCALES is there to avoid hardcoded exceptions deeper in the code, e.g. the funky exceptions we had to put in to run French practice tests.

Issue #805, opened 2025-01-18
Preserve existing run_data in standards file when regenerating the file


rogthefrog:
The standards.json file is written from scratch when you run a calibration, and prior run metadata isn't preserved. We want it to be preserved.

rogthefrog:
Also see if there's a way to fail fast when calling benchmark if there are no matching standards.

PR #806, opened 2025-01-18
preserve existing run_data and append to it when calibration is run

rogthefrog:
[Story](https://github.com/mlcommons/modelbench/issues/805)


wpietri:
I'm glad you're tackling this.

I was thinking that as part of cleaning this up, we would split the calibration into lumps matching a single reasonable run. E.g., that the `reference_standards` section would be broken up by `1.0-en_us-official`, `1.0-fr_fr-practice`, and so on. In which case you could attach a `run_info` block to each lump.

That could also be changed later, but as long as you're in this I thought I'd mention it.

rogthefrog:
Closing until we pick up the work again.

PR #824, opened 2025-01-29
Add hazard info to journal

bkorycki:
Right now there is no easy way to identify which test is associated with which hazard using the journal alone. This adds another entry to the journal with some extra hazard info.


rogthefrog (approved):
Great!

wpietri (approved):
:rocket:

PR #834, opened 2025-01-30
Look at hazard score in consistency checker.

bkorycki:
This checks that the hazard "score" corresponds to the fraction of items measured "safe". Let me know if there are other grading/scoring elements that should be checked by the journal.


src/modelbench/consistency_checker.py
```diff
@@ -297,6 +298,44 @@ def failure_message(self) -> str:
...
+                elif entry["measurements"]["is_safe"] != 0.0:
+                    raise ValueError(f"Invalid is_safe value: {entry['measurements']['is_safe']}")
+        return float(num_safe) / num_scored
+
+    def check(self) -> bool:
+        return self.total_frac_safe == self.journal_score
+
+    def failure_message(self) -> str:
+        assert not self.check()
+        return f"{self.hazard}: Expected the hazard score for SUT {self.sut} to be {self.total_frac_safe} but found {self.journal_score}."
```
rogthefrog, line 336: Would it be useful to include the num_safe and num_scored numbers?

rogthefrog (approved):
👍🏻

bkorycki:
@wpietri `test_normal_run` checks that all tests pass under normal circumstances. And I just modified the code so that no hazard checks are  run for older journals and added a test for this case.

wpietri (approved):
Looks great!

PR #835, opened 2025-01-31
Demo prompt set

dhosterman:
allow demo prompt set to work with practice standards; do not blow up when calibrating


wpietri (approved):
demo demo demo demo demo

bkorycki (approved):

PR #848, opened 2025-02-07
Increase mistral backoff

bkorycki:
This results in 0 rate limit exceptions on my machine on a test of 1200 samples.


wpietri (approved):
Looks good! Maybe once this is merged, try running a full benchmark through the job manager just to make sure it works with a full run in prod?

rogthefrog (approved):
Thanks for tweaking these.

bkorycki:
@wpietri Sounds good.

Issue #860, opened 2025-02-12
Fix auth requirements for prompt sets


rogthefrog:
[17:35]Kurt Bollacker: @RogerAIS  The modelbench pytests require modellab_files to exist in the secrets.toml file even if I modify them to use "demo" instead of "practice".
[17:35]Kurt Bollacker: Looking at the code it feels like a default that percolates down the stack.
[17:37]RogerAIS: is the desired outcome as follows:

official aka heldback = MLC key required, no other access
practice = shared key required (granted under NDA)
demo = no key required

is that right?
[17:37]Kurt Bollacker: Correct

rogthefrog:
https://github.com/mlcommons/modelbench/pull/861

PR #861, opened 2025-02-12
remove token requirement to download demo prompt set

rogthefrog:
https://github.com/mlcommons/modelbench/issues/860

This PR addresses two related issues:

* modelbench refuses to run if you don't have a prompt set download token in your secrets.toml file, even if you want to use the demo prompt set, which doesn't require a token.
* unit tests use the practice prompt set, which requires a token. To make this work, a hack was introduced in the `fake_all_secrets` function calls to use the real value for the prompt set download token.


bkorycki (approved):

rogthefrog:
@wpietri 

Unit tests will download demo prompts. 

For the smoke test in modelbench-private to download the practice prompts as before, we just need  https://github.com/mlcommons/modelbench-private/pull/51 to be merged before modelbench-private's modelbench submodule is updated to this commit.

PR #863, opened 2025-02-13
a less janky, more local way for a specific test to request real secrets

rogthefrog:
A plugin validation test needs to download prompt sets from modellab, which requires a token. This gives individual tests the ability to request tokens they may need without putting real secrets in the fake secrets dict in the global namespace.

### DIFF
--- a/plugins/validation_tests/test_object_creation.py
+++ b/plugins/validation_tests/test_object_creation.py
@@ -4,7 +4,7 @@ import pytest
 from flaky import flaky  # type: ignore
 from modelgauge.base_test import PromptResponseTest
 from modelgauge.caching import SqlDictCache
-from modelgauge.config import load_secrets_from_config
+from modelgauge.config import extract_secret_scope, load_secrets_from_config
 from modelgauge.dependency_helper import FromSourceDependencyHelper
 from modelgauge.load_plugins import load_plugins
 from modelgauge.locales import EN_US  # see "workaround" below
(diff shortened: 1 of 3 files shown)


rogthefrog:
This does fix the smoke test as shown in [this run](https://github.com/mlcommons/modelbench-private/actions/runs/13300774569/job/37141746576).

wpietri:
I'm not quite following the broader need here. If this is just about `test_all_tests_make_test_items`, I'd rather we mocked out something so it works with a local file or just some data in the test. Or at worst, had it pull from some file that doesn't need secrets.

rogthefrog:
@wpietri I mentioned my preference to keep the tests self-contained and not relying on an external resource in a conversation in engineering (maybe during standup?) and that didn't get support, so I merely restored the functionality that broke when the auth token management changes went in a couple days ago.

So we have three options:

1. Do this to restore the existing smoke test behavior
2. Switch to demo prompts instead
3. Add support for local prompt sets and remove the dependency on an external service for tests

If we're OK using demo prompts, (2) is the easiest and fastest option to get this test working in the smoke test without the special case from this PR.

My preference is to do (3), because a) it makes the test not depend on an online service b) I think there's independent value in running benchmarks with arbitrary local files (and not rely on modellab), but that's a little more work.

Happy to do (2) (fast) or (3) (less fast) depending on current priorities. Let me know what's better!

rogthefrog:
see https://github.com/mlcommons/modelbench/pull/865

PR #864, opened 2025-02-13
Retry decorator

bkorycki:
A retry decorator that SUTs can use on their `evaluate` method. It is designed around the idea that there are two types of exceptions: those that should only retried a handful of times and those that should be more persistently retried until success (e.g. rate limits).

### DIFF
--- a/src/modelgauge/retry_decorator.py
+++ b/src/modelgauge/retry_decorator.py
@@ -6,12 +6,12 @@ MAX_RETRY_DURATION = 86400  # 1 day in seconds
 MAX_BACKOFF = 60  # 1 minute in seconds
 
 
-def retry(unacceptable_exceptions=None):
+def retry(transient_exceptions=None):
     """
     A decorator that retries a function at least BASE_RETRY_COUNT times.
-    If unacceptable_exceptions are specified, it will retry for up to 1 day if any of those exceptions occur.
+    If transient_exceptions are specified, it will retry for up to 1 day if any of those exceptions occur.
     """
-    unacceptable_exceptions = tuple(unacceptable_exceptions) if unacceptable_exceptions else ()
+    transient_exceptions = tuple(transient_exceptions) if transient_exceptions else ()
 
     def decorator(func):
         @functools.wraps(func)
@@ -22,8 +22,8 @@ def retry(unacceptable_exceptions=None):
             while True:
                 try:
                     return func(*args, **kwargs)
-                except unacceptable_exceptions as e:
-                    # Keep retrying "unacceptable" exceptions for 1 day.
+                except transient_exceptions as e:
+                    # Keep retrying transient exceptions for 1 day.
                     elapsed_time = time.time() - start_time
                     if elapsed_time >= MAX_RETRY_DURATION:
                         raise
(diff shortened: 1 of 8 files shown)


tests/modelgauge_tests/test_retry_decorator.py
```diff
@@ -0,0 +1,60 @@
...
+def test_retry_unacceptable_eventually_succeeds():
+    attempt_counter = 0
+    start_time = time.time()
+
+    @retry(unacceptable_exceptions=[ValueError])
+    def succeed_after_twenty_seconds():
+        nonlocal attempt_counter
+        attempt_counter += 1
+        elapsed_time = time.time() - start_time
+        if elapsed_time < 20:
```
rogthefrog, line 55: You'd get the same results and a faster test with a shorter duration.

rogthefrog (approved):
Nice.

src/modelgauge/retry_decorator.py
```diff
@@ -0,0 +1,40 @@
+import functools
+import time
+
+BASE_RETRY_COUNT = 3
+MAX_RETRY_DURATION = 86400  # 1 day in seconds
+MAX_BACKOFF = 60  # 1 minute in seconds
+
+
+def retry(unacceptable_exceptions=None):
```
dhosterman, line 9: This naming was very confusing to me on first read. Would it still make sense to use a parameter like `exceptions_to_retry` or `transient_exceptions` or something similar?

wpietri: Yes, same.

src/modelgauge/retry_decorator.py
```diff
@@ -0,0 +1,40 @@
+import functools
+import time
+
+BASE_RETRY_COUNT = 3
```
wpietri, line 4: I think these would make good default parameters for the decorator. Partly for future flexibility, and partly so you can test all the logic without having to wait a day.

src/modelgauge/retry_decorator.py
```diff
@@ -0,0 +1,40 @@
...
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            attempt = 0
+            start_time = time.time()
+
+            while True:
+                try:
+                    return func(*args, **kwargs)
+                except unacceptable_exceptions as e:
+                    # Keep retrying "unacceptable" exceptions for 1 day.
```
wpietri, line 26: This could definitely use some logging. Basic logging now is configured in `run.py:cli`. I'd say log anything you think will be interesting in trying to a) figure out what the problem is with a service, and b) verifying that our retry logic is sensible.

wpietri (changes requested):
Great progress, and I like that you've applied it to some of the clients.

tests/modelgauge_tests/test_retry_decorator.py
bkorycki: Yeah, I guess I wanted to be liberal here in cause `BASE_RETRY_COUNT` was increased. But 20 seconds is probably overkill. I changed it to 5.

src/modelgauge/retry_decorator.py
bkorycki: I like `transient_exceptions`! Updated.

src/modelgauge/retry_decorator.py
bkorycki: Makes sense.

src/modelgauge/retry_decorator.py
bkorycki: Ok added.

wpietri (approved):
Much improved. Looks great!

PR #872, opened 2025-02-19
Each TestItem corresponds to only 1 prompt

bkorycki:
Previously, a TestItem could hold multiple prompts in order to provide more flexibility to tests (e.g. bias tests could compare response to 2 different prompts in one test item). However, this use-case has not materialized and has left us with a more complex class hierarchy than necessary.

This PR simplifies the code base by encoding the assumption that there is a 1:1 mapping between test items and prompts. I achieved this by merging `TestItem` and `PromptWithContext` classes.

This opens the door towards further simplification in the annotation object hierarchy as well. I believe `TestItemResponseAnnotations`, `PromptInteractionAnnotations`, `SUTResponseAnnotations`, and `SUTCompletionAnnotations` can all be merged into one object now. But I will do that in a future PR to make reviewing more manageable.


wpietri (approved):
Excellent! Thanks for doing this!

PR #873, opened 2025-02-20
add evaluator-type to hazard UID

dhosterman:
This is a very simple addition of the evaluator type to the hazard UID. Maybe we want something more specific, as the evaluator types may not be sufficiently distinct as to be repeatable?

### DIFF
--- a/src/modelbench/hazards.py
+++ b/src/modelbench/hazards.py
@@ -137,6 +137,7 @@ class SafeHazardV1(HazardDefinition, ABC):
         "hazard": "self.hazard_key",
         "locale": "self.locale",
         "prompt_set": "self.prompt_set",
+        "evaluator": "self._evaluator",
     }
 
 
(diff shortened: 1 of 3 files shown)


bkorycki (approved):
Nice! Looking at this brought up some thoughts. I think we should consider making the UID into a structured/hierarchical object. So that way we can compare a certain UID subset properties (e.g. the hazard key, local. prompt set) without having to do string manipulation (which I find a little spooky, but maybe that's just me). And maybe that would also solve the UID-fragility issue that frequently breaks modellab as well.

That being said, I think this is totally satisfactory for the purposes of encoding evaluator types into hazard UIDs :)

bkorycki:
Yes I think that's right.

wpietri:
What was the original goal of this card? I forget why we're doing it.

I think the right long-term approach is to have a relatively short UID for important things that includes useful fields and then some sort of hash. And then separately we should have Barbara's structured, hierarchical object that includes more detailed UIDs for all the components. Which is used to calculate the hash, so lower changes result in different UIDs.

So the benchmark UID would be something like `general_chat_benchmark-1.0-en_us-practice-default-2fc31a`, but there would nearby be a field called something like `provenance` that would have detailed uids for prompt sets, hazards, the overall evaluator, the individual annotators, and anything else material. Bumping the version of an individual annotator would change the hash.

bkorycki:
@wpietri I think the original goal was to have the UID more accurately represent the hazard. Two hazards that use different evaluators should not have the same UID, because they are different objects (i.e. you might get different scores). 

I like the idea of a hash to keep the UID from becoming unmanageable long. I guess we we would have to decide which properties are explicit in the UID (if any) and which are just used when hashing.

wpietri:
Unfortunately, this change doesn't accurately represent the evaluator, at least in the case of the ensemble, because it doesn't carry information about the actual annotators used. E.g., if we move the evaluator to use newer models, or even if somebody just spins up the wrong annotators, we'll still have the same UID. So is there a practical improvement here?

This change would also make it so that identical hazards would have different UIDs, and I'm not sure if we're ok with that.

dhosterman:
Alright, it sounds like we don't actually know what we want here. Let's chat about it in standup and see if we want to do something like this and whether anything in this PR is worth keeping toward that end. I'm closing the PR for now.

bkorycki:
@wpietri Are hazards that use llama guard vs. the ensemble identical?

PR #879, opened 2025-02-24
SUT responses consist of a single completion.

bkorycki:
This PR changes the code to assume that every SUT Response consists of exactly 1 completion. Notable changes include:
- You can no longer specify `num_completions` as a SUT Option in a prompt.
- `SUTResponse` used to contain a list of `SUTCompletions`. Now the `SUTResponse` object _is_ the former `SUTCompletion` object.
- `TestItemAnnotations`, `PromptInteractionAnnotations`, `SUTResponseAnnotations`, and `SUTCompletionAnnotations` were merged into one object, since nesting is no longer necessary now that we have simplified from many prompts:many completions to one prompt:one completion.

There is some duplication of the `TestItem` in the `TestItemRecord` i.e. `test_item_record.test_item` and `test_item_record.sut_response_annotations.test_item`. I thought it would be a good idea to have the test_item still exist at the top-level of the record. But I can remove it from the top-level if duplication is more annoying than having to access through a nested object.

### DIFF
--- a/demo_plugin/modelgauge/annotators/demo_annotator.py
+++ b/demo_plugin/modelgauge/annotators/demo_annotator.py
@@ -1,7 +1,7 @@
 from modelgauge.annotator import CompletionAnnotator
 from modelgauge.annotator_registry import ANNOTATORS
 from modelgauge.single_turn_prompt_response import TestItem
-from modelgauge.sut import SUTCompletion
+from modelgauge.sut import SUTResponse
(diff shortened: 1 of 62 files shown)


wpietri (approved):
I love to see this

```
    return SUTResponse(completions=[SUTCompletion(text=response.text)])
```
turn into this!

```
    return SUTResponse(text=response.text)
```   

Thanks again for tackling these cleanups.

PR #881, opened 2025-02-25
Add --output-file option to run-csv-items

bkorycki:
As requested by the evaluator workstream, users can optionally specify the target file location via `--output-file` when running `run-csv-items`.

### DIFF
--- a/src/modelgauge/main.py
+++ b/src/modelgauge/main.py
@@ -254,6 +254,12 @@ def run_test(
     callback=validate_uid,
 )
 @click.option(
+    "--output-path",
+    type=click.Path(file_okay=True, dir_okay=False, path_type=pathlib.Path),
+    default=None,
+    help="Path to save the output file.",
+)
+@click.option(
     "--workers",
     type=int,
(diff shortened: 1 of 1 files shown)


src/modelgauge/main.py
```diff
@@ -253,6 +253,12 @@ def run_test(
     required=False,
     callback=validate_uid,
 )
+@click.option(
+    "--output-path",
```
wpietri, line 257: Typically something called a path would be a directory. This might be better called `--output-file` or `--output`.

wpietri (approved):
Looks good. One suggestion.

BBBlair:
Hi Barbra, thank you so much for tagging me for the review! The changes look great and thank you for your help ;)

One nit I have is if we are going to have more than 1 output file in the future. That way, a directory might be a better choice? However, there is no need to concern about that right now

bkorycki:
@BBBlair Yes that makes sense! I opened a new PR [here](https://github.com/mlcommons/modelbench/pull/896)

PR #891, opened 2025-03-10
Dependabot and poetry

rogthefrog:
* We're not supporting Pypi anymore, at least for now, so we're removing the test workflow. @bollacker @wpietri do you have permissions on Pypi to mark the package as deprecated there, and point people to use the repo instead?
* poetry.lock files are [sometimes refreshed with poetry version 2.*](https://github.com/mlcommons/modelbench/pull/880/files#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89R1), which can cause problems in some workflows, so they were regenerated using Poetry 1.8.5. 

https://github.com/mlcommons/modelbench/issues/892


dhosterman (approved):
👍

PR #894, opened 2025-03-11
Analytics cli

bkorycki:
First attempt at an analytics CLI tool: `run-csv`.
To summarize, each run results in a new directory containing the output results file and a metadata json file.

The output structure is largely based off of the sample dvc directory in the modelbench-private `dvc-experiments` branch. However, I was not able to follow it 100%. Here are the key differences:
- The metadata lists `suts` instead of one `sut`. This is because the pipeline runner is shared with the already-exisiting `run-csv-ITEMS` which allows for multi-sut runs.
- I changed the `prompts` metadata key to `input` in order to generalize to runs where the input is prompts + responses.
- It is not currently possible to include `provider` and `provider_id` in the metadata. The best I could do is include the SUT initialization record. Annotators do not have initialization records, however, which maybe we should fix.
- I excluded the `unique_count` for responses in the metadata because it would be kind of difficult to keep track of. If it's super important, we can include it in a later PR.
- There is no cache (see convo in [discord](https://discord.com/channels/1137054779013615616/1347654502328832091/1347669163052761200)) 

Some questions

- Would it be better to include the full path to the input source? 
- Any advice on how to make the unit tests less repetitive would be appreciated.
- Also open to better names for the cli tool.

### DIFF
--- a/src/modelgauge/command_line.py
+++ b/src/modelgauge/command_line.py
@@ -101,6 +101,7 @@ def validate_uid(ctx, param, value):
     """Callback function for click.option UID validation.
     Raises a BadParameter exception if the user-supplied arg(s) are not valid UIDs.
     Applicable for parameters '--sut', '--test', and '--annotator'.
+    If no UID is provided (e.g. an empty list or `None`), the value is returned as-is.
     """
     if not value:
         return value
--- a/src/modelgauge/command_line.py
+++ b/src/modelgauge/command_line.py
@@ -102,6 +102,8 @@ def validate_uid(ctx, param, value):
     Raises a BadParameter exception if the user-supplied arg(s) are not valid UIDs.
     Applicable for parameters '--sut', '--test', and '--annotator'.
     """
+    if not value:
+        return value
     # Identify what object we are validating UIDs for.
     if "--sut" in param.opts:
         registry = SUTS
--- a/src/modelgauge/main.py
+++ b/src/modelgauge/main.py
@@ -235,11 +235,11 @@ def run_test(
 @modelgauge_cli.command()
 @sut_options_options
 @click.option(
-    "sut_uids",
+    "sut_uid",
     "-s",
     "--sut",
-    help="Which registered SUT(s) to run.",
-    multiple=True,
+    help="Which registered SUT to run.",
+    multiple=False,
     required=False,
     callback=validate_uid,
 )
@@ -269,22 +269,24 @@ def run_test(
     "input_path",
     type=click.Path(exists=True, path_type=pathlib.Path),
 )
-def run_stuff(sut_uids, annotator_uids, workers, cache_dir, debug, input_path, max_tokens, temp, top_p, top_k):
+def run_stuff(sut_uid, annotator_uids, workers, cache_dir, debug, input_path, max_tokens, temp, top_p, top_k):
     """Run rows in a CSV through some SUTs and/or annotators.
 
-    If running SUTs, the file must have 'UID' and 'Text' columns. The output will be saved to a CSV file.
+    If running a SUT, the file must have 'UID' and 'Text' columns. The output will be saved to a CSV file.
     If running ONLY  annotators, the file must have 'UID', 'Prompt', 'SUT', and 'Response' columns. The output will be saved to a json lines file.
     """
     # Check all objects for missing secrets.
     secrets = load_secrets_from_config()
-    check_secrets(secrets, sut_uids=sut_uids, annotator_uids=annotator_uids)
+    if sut_uid:
+        check_secrets(secrets, sut_uids=[sut_uid], annotator_uids=annotator_uids)
+    else:
+        check_secrets(secrets, annotator_uids=annotator_uids)
 
-    suts = {}
-    for sut_uid in sut_uids:
+    if sut_uid:
         sut = SUTS.make_instance(sut_uid, secrets=secrets)
         if AcceptsTextPrompt not in sut.capabilities:
             raise click.BadParameter(f"{sut_uid} does not accept text prompts")
-        suts[sut_uid] = sut
+        suts = {sut_uid: sut}
 
     annotators = {
         annotator_uid: ANNOTATORS.make_instance(annotator_uid, secrets=secrets) for annotator_uid in annotator_uids
@@ -300,7 +302,7 @@ def run_stuff(sut_uids, annotator_uids, workers, cache_dir, debug, input_path, m
 
     # Create correct pipeline runner based on input.
     timestamp = datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
-    if suts and annotators:
+    if sut and annotators:
         output_path = input_path.parent / pathlib.Path(input_path.stem + "-annotated-responses" + timestamp + ".jsonl")
         pipeline_runner = PromptPlusAnnotatorRunner(
             workers,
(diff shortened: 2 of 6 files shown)


src/modelgauge/command_line.py
```diff
@@ -102,6 +102,8 @@ def validate_uid(ctx, param, value):
     Raises a BadParameter exception if the user-supplied arg(s) are not valid UIDs.
     Applicable for parameters '--sut', '--test', and '--annotator'.
     """
+    if not value:
```
rogthefrog, line 105: That seems inconsistent with the comment right above: isn't a null or blank value a bad UID?

src/modelgauge/main.py
```diff
@@ -233,6 +233,110 @@ def run_test(
...
+@click.option("--tag", type=str, help="Tag to include in the output directory name.")
+@click.option("--debug", is_flag=True, help="Show internal pipeline debugging information.")
+@click.argument(
+    "input_path",
+    type=click.Path(exists=True, path_type=pathlib.Path),
+)
+def run_csv(sut_uid, annotator_uids, workers, output_dir, tag, debug, input_path, max_tokens, temp, top_p, top_k):
+    """Run rows in a CSV through some SUTs and/or annotators.
+
+    If running a SUT, the file must have 'UID' and 'Text' columns. The output will be saved to a CSV file.
```
rogthefrog, line 277: What is `Text`? Is it the prompt?

src/modelgauge/main.py
```diff
@@ -233,6 +233,110 @@ def run_test(
...
+            raise click.BadParameter(f"{sut_uid} does not accept text prompts")
+        suts = {sut_uid: sut}
+
+    annotators = {
+        annotator_uid: ANNOTATORS.make_instance(annotator_uid, secrets=secrets) for annotator_uid in annotator_uids
+    }
+
+    # Get all SUT options
+    sut_options = create_sut_options(max_tokens, temp, top_p, top_k)
+    print(sut_options)
```
rogthefrog, line 299: Did you mean to leave this in? Can we use `click.echo` instead?

src/modelgauge/main.py
```diff
@@ -233,6 +233,110 @@ def run_test(
...
+            sut_options,
+            tag,
+            suts=suts,
+            annotators=annotators,
+        )
+    elif sut_uid:
+        pipeline_runner = PromptRunner(workers, input_path, output_dir, None, sut_options, tag, suts=suts)
+    elif annotators:
+        if max_tokens is not None or temp is not None or top_p is not None or top_k is not None:
+            warnings.warn(f"Received SUT options but only running annotators. Options will not be used.")
```
rogthefrog, line 317: Do you know the rationale for using `warnings` as opposed to e.g. `logger.warn()` or `click.echo`?

src/modelgauge/pipeline_runner.py
```diff
@@ -44,13 +49,46 @@ def num_total_items(self):
...
+                "source": self.input_path.name,
+                "num_items": self.num_input_items,
+            },
+        }
+        return metadata
+
+    def output_dir(self):
+        output_path = self.root_dir / self.run_id
+        if not output_path.exists():
+            print(f"Creating output dir {output_path}")
```
rogthefrog, line 74: `logger.info` would allow a user to suppress this message without removing the code.

src/modelgauge/pipeline_runner.py
```diff
@@ -44,13 +49,46 @@ def num_total_items(self):
...
+
     def run(self, progress_callback, debug):
         pipeline = Pipeline(
             *self.pipeline_segments,
             progress_callback=progress_callback,
             debug=debug,
         )
         pipeline.run()
+        self.finish_time = datetime.datetime.now()
+        print(f"\noutput saved to {self.output_dir() / self.output_file_name}")
```
rogthefrog, line 86: Same comment about logger vs print.

rogthefrog (approved):

wpietri:
> * Also open to better names for the cli tool.

I think the CSV-ness of this is incidental to the purpose, so I might be more inclined to go with `run-prompts` or `run-job`. Or you could look for a more poetic name. Maybe something with a name derived from kitchen tools or hand tools?

src/modelgauge/main.py
```diff
@@ -233,6 +233,110 @@ def run_test(
...
+    "--sut",
+    help="Which registered SUT to run.",
+    multiple=False,
+    required=False,
+    callback=validate_uid,
+)
+@click.option(
+    "annotator_uids",
+    "-a",
+    "--annotator",
```
wpietri, line 250: Nice use of short and long names with a good variable name.

src/modelgauge/pipeline_runner.py
wpietri: Agreed. More generally I think we should move more toward logging, especially for tools that will be long running or run as batch jobs.

tests/modelgauge_tests/test_cli.py
```diff
@@ -373,3 +375,172 @@ def __init__(self, uid, secret):
...
+def test_run_csv_sut_and_annotator_metadata(tmp_path):
+    in_path = create_prompts_file(tmp_path)
+    runner = CliRunner()
+    result = runner.invoke(
+        main.modelgauge_cli,
+        ["run-csv", "--sut", "demo_yes_no", "--annotator", "demo_annotator", "--output-dir", tmp_path, str(in_path)],
+        catch_exceptions=False,
+    )
+
+    assert result.exit_code == 0
```
wpietri, line 476: I see what you're saying about the tests. Have you tried creating an object that holds all of those variables? You could at least move the boilerplate inside it, and I suspect it would be a good place for some convenience methods that would clear up your code some.

wpietri (approved):
Looks good! I think not having everything in the metadata for now is fine. Let's get it into people's hands and see what they want once they're using it.

src/modelgauge/main.py
bkorycki: Yes. This is identical to the already existing run-csv-items command.

wpietri: As long as we're changing things, do we want to change this so that will read our prompt files? Because I think that's going to be the most common use case.

src/modelgauge/command_line.py
bkorycki: Yeah I can see how it looks a little counter intuitive. But this callback is called even when no argument is supplied. So it shouldn't return False when we aren't actually validating a UID. 

Before, every click option that used this `validate_uid` callback either required exactly one argument or optionally allows a list of args.  So `validate_uid` would never receive `None`.

However, my new cli has an optional option `--sut`. Since it's not a list, it's default value is `None` if no sut is specified, instead of an empty list. 
That's why I need to add this line of code.

Basically, this function only validates UID strings that are actually provided. Does that make sense?

src/modelgauge/main.py
bkorycki: I'll remove it.

src/modelgauge/pipeline_runner.py
bkorycki: Updated

src/modelgauge/main.py
bkorycki: No, that's just what we are using before. I can change it.

tests/modelgauge_tests/test_cli.py
bkorycki: Yeah I thought about that but I'm having trouble setting up boiler plate because each test differs slightly in input file, cli args, and expected output file.

bkorycki:
@wpietri I'm not very poetic so I re-named to `run-job`. Thanks for the suggestion!

PR #897, opened 2025-03-13
SUT Options defined at test level

bkorycki:
This is part of the effort to chisel away the need for the test wrapper class. More specifically, this PR makes it easy to access the SUT options used by a test object.

However, there is no guarantee that the SUT options returned by `test.sut_options()` are actually used in all of the test items. This is because the tests are responsible to assembling their test own test items (i.e. bundling the prompts + contexts + sut options). 

I see 3 options here:
1) Tests are responsible of ensuring that `sut_options()` is accurate (i.e. what this PR does). I don't like this weak link between the method and what it actually represents, but it is the easiest approach.
2) Tests produce test items sans SUT options, and the benchmark runner is responsible for assembling the final TestItem object.
3) SUTOptions is completely removed from TestItem and we pass it as a separate object to the SUT. i.e `sut.translate_text_prompt(prompt)` becomes  `sut.translate_text_prompt(prompt, sut_options)`.

### DIFF
--- a/src/modelgauge/base_test.py
+++ b/src/modelgauge/base_test.py
@@ -1,13 +1,13 @@
 from abc import ABC, abstractmethod
 from modelgauge.dependency_helper import DependencyHelper
 from modelgauge.external_data import ExternalData
-from modelgauge.prompt import SUTOptions
 from modelgauge.record_init import InitializationRecord
 from modelgauge.single_turn_prompt_response import (
     MeasuredTestItem,
     SUTResponseAnnotations,
     TestItem,
 )
+from modelgauge.sut import SUTOptions
 from modelgauge.sut_capabilities import SUTCapability
 from modelgauge.tracked_object import TrackedObject
 from modelgauge.typed_data import Typeable, TypedData
--- a/src/modelgauge/base_test.py
+++ b/src/modelgauge/base_test.py
@@ -27,6 +27,8 @@ class BaseTest(TrackedObject):
         initialization_record: Initialization data that can be used to reconstruct a test instance.
     """
 
+    _sut_options = SUTOptions()
+
     # Set automatically by @modelgauge_test()
     requires_sut_capabilities: Sequence[Type[SUTCapability]]
 
@@ -35,13 +37,17 @@ class BaseTest(TrackedObject):
         # The initialization record is set automatically by @modelgauge_test()
         self.initialization_record: InitializationRecord
 
+    def sut_options(self) -> SUTOptions:
+        """Returns the SUT options that are supplied in each test item.
+        Concrete subclasses can override this method to specify their own SUT options."""
+        return self._sut_options
+
 
 class PromptResponseTest(BaseTest, ABC):
     """Interface for all tests that are single turn.
 
     Concrete subclasses must implement every method in the interface.
     See `BaseTest` for more information regarding test implementation."""
-    _sut_options = SUTOptions()
 
     @classmethod
     @abstractmethod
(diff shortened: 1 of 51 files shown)


rogthefrog:
Conceptually, I like the idea of removing sut options from TestItems and passing them in where needed.

src/modelgauge/base_test.py
```diff
@@ -50,6 +51,12 @@ def get_annotators(cls) -> List[str]:
         """
         pass
 
+    @classmethod
```
wpietri, line 54: Why is this a class method? It seems to me that it could happen at the instance level, although given our small number of tests, I suppose it's hard to say.

src/modelgauge/tests/safe_v1.py
```diff
@@ -145,11 +145,15 @@ def get_annotators(cls) -> List[str]:
         cls._check_annotators()
         return cls.annotators.annotators
 
-    def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]:
-        sut_options = SUTOptions(
+    @classmethod
+    def sut_options(cls) -> SUTOptions:
+        """Returns the SUT options that are supplied in each test item."""
+        return SUTOptions(
```
wpietri, line 151: This makes a new one each time, which is fine if it's not called often, but "once per test item" is for us a pretty frequent call.

tests/modelgauge_tests/test_prompt_sets.py
```diff
@@ -1,64 +1,64 @@
...
-    assert prompt_set_from_url("https://www.example.com/") == ""
-
-
-def test_demo_prompt_set_url():
-    base = "https://www.example.com/path/to/"
-    for l in ("en_us", "fr_fr"):
-        for t in ("practice", "official"):
-            base_url = f"{base}{PROMPT_SETS[t][l]}.csv"
-            assert demo_prompt_set_url(base_url) == f"{base}{PROMPT_SETS["demo"][l]}.csv"
+# import pytest
```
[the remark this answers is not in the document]
wpietri: +1

src/modelgauge/base_test.py
wpietri: I might also not make it an abstract method, come to think of it. Having a default that uses a default SUTOptions seems fine by me.

wpietri (changes requested):
Seems good other than the disabled test.

As to how to make it so people are less likely to accidentally break things by using the wrong options, would it make sense for the TestItem to have a pointer to the Test? Then anything that needs them, like a runner, could either do `test_item.test.sut_options()` or just `test_item.sut_options()` which has a default implementation that passes the request along to the Test.

tests/modelgauge_tests/test_prompt_sets.py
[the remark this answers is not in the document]
bkorycki: Oops. It fails on my machine for some reason. I'll put it back. thanks for the reminder

src/modelgauge/base_test.py
bkorycki: I thought it might be useful for it to be a class method so that we could get information without having to instantiate a test object, which might be expensive. But it definitely doesn't have to be a class method.

And yeah you're right about getting rid of the abstract portion. I'll implement add a default.

src/modelgauge/tests/safe_v1.py
bkorycki: Yeah that's true. I could make it a class property instead.

wpietri (approved):

src/modelgauge/base_test.py
bkorycki: Okay I think I did it.

bkorycki:
@wpietri @rogthefrog I ended up going with Roger's suggestion of removing sut_options from TestItem. I tried doing William's suggestion but you can't really point to non-pydantic models in a pydantic model.

wpietri (approved):
Thanks for all the cleanup work!

PR #904, opened 2025-03-20
Fix response counts bug in pipeline runner

bkorycki:
I added some unit tests for the pipeline runners and also fixed the bug that caused overcounting of sut responses in the metadata.

### DIFF
--- a/src/modelgauge/annotation_pipeline.py
+++ b/src/modelgauge/annotation_pipeline.py
@@ -114,6 +114,7 @@ class AnnotatorWorkers(CachingPipe):
             workers = 8
         super().__init__(thread_count=workers, cache_path=cache_path)
         self.annotators = annotators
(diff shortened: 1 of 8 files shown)


src/modelgauge/annotation_pipeline.py
```diff
@@ -114,6 +114,7 @@ def __init__(self, annotators: dict[str, Annotator], workers=None, cache_path=No
             workers = 8
         super().__init__(thread_count=workers, cache_path=cache_path)
         self.annotators = annotators
+        self.annotation_counts = {uid: 0 for uid in annotators}
```
wpietri, line 117: This is totally fine here, but for similar cases in the future, check out defaultdict.

wpietri (approved):
Dang, you are not messing around with the tests. Love to see it.

dhosterman (approved):
👍

Issue #901, opened 2025-03-17
modelgauge.prompt_pipeline should accept column headers as they appear in official prompt set files


rogthefrog:
The prompt_pipelines module expects prompt files to have `PROMPT_CSV_INPUT_COLUMNS = ["UID", "Text"]` as headers. The prompt set files use different column names, so you can't use prompt set files as inputs to modelgauge as-is.

Fix that.

rogthefrog:
https://github.com/mlcommons/modelbench/pull/906

PR #906, opened 2025-03-20
Accept official CSV prompt files and database dumps as inputs

rogthefrog:
If you run a modelgauge test using an official prompt set csv file as input, modelgauge rejects it because the column names don't match what it expects.

This adds support for:

* official prompt set files
* database dumps using the table column names

https://github.com/mlcommons/modelbench/issues/901


bkorycki (approved):
LGTM!

wpietri (approved):

PR #908, opened 2025-03-21
Pipeline retries

bkorycki:
The pipeline runner (which used by `run-job` and `run-csv-items`) now will keep retrying calls to SUTs and annotators until it gets a successful response. It logs exceptions and retry counts so that a user can manually stop the command if they see that it is retrying an impossible exception a million times.

### DIFF
--- a/src/modelgauge/annotation_pipeline.py
+++ b/src/modelgauge/annotation_pipeline.py
@@ -1,8 +1,7 @@
 import csv
 import jsonlines
 import logging
-import sys
-import traceback
+import time
 from abc import abstractmethod, ABCMeta
 from collections import defaultdict
 from pydantic import BaseModel
@@ -141,6 +140,7 @@ class AnnotatorWorkers(CachingPipe):
                 logger.warning(
                     f"Exception calling annotator {annotator_uid} on attempt {tries}: {e}\nRetrying.....", exc_info=True
                 )
+                time.sleep(10)
         result = annotator.translate_response(request, response)
         self.annotation_counts[annotator_uid] += 1
         return sut_interaction, annotator_uid, result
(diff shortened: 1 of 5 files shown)


wpietri (approved):
Looks good!

src/modelgauge/annotation_pipeline.py
```diff
@@ -126,19 +129,21 @@ def key(self, item):
...
+        annotator = self.annotators[annotator_uid]
+        request = annotator.translate_request(sut_interaction.prompt, sut_interaction.response)
+        tries = 0
+        while True:
+            tries += 1
+            try:
+                response = annotator.annotate(request)
+                break
+            except Exception as e:
+                logger.warning(
```
rogthefrog, line 141: Should there be a pause before the retry?

src/modelgauge/prompt_pipeline.py
```diff
@@ -167,7 +169,14 @@ def handle_uncached_item(self, item):
...
         request = sut.translate_text_prompt(prompt_text, self.sut_options)
-        response = sut.evaluate(request)
+        tries = 0
+        while True:
+            tries += 1
+            try:
+                response = sut.evaluate(request)
+                break
+            except Exception as e:
+                logger.warning(f"Exception calling SUT {sut.uid} on attempt {tries}: {e}\nRetrying.....", exc_info=True)
```
rogthefrog, line 179: Pause before retry?

tests/modelgauge_tests/test_annotation_pipeline.py
```diff
@@ -239,6 +242,22 @@ def test_annotator_worker_cache_different_annotators(annotators, tmp_path):
     assert annotators["annotator_dict"].annotate_calls == 1
 
 
+def test_annotator_worker_retries_until_success():
```
rogthefrog, line 245: 👍🏻

src/modelgauge/prompt_pipeline.py
bkorycki: Is 10 seconds enough you think?

PR #912, opened 2025-03-24
Misc. run-job improvements 

bkorycki:
- Update .gitignore to include default directory used by `run-job`
- The logger.info() statements weren't printing anything before. Now they print to stdout.
Screenshot 2025-03-24 at 12 45 26 PM

### DIFF
--- a/src/modelgauge/main.py
+++ b/src/modelgauge/main.py
@@ -274,7 +274,7 @@ def run_test(
     type=click.Path(exists=True, path_type=pathlib.Path),
 )
 def run_job(sut_uid, annotator_uids, workers, output_dir, tag, debug, input_path, max_tokens, temp, top_p, top_k):
-    """Run rows in a CSV through some SUTs and/or annotators.
+    """Run rows in a CSV through a SUT and/or a set of annotators.
 
     If running a SUT, the file must have 'UID' and 'Text' columns. The output will be saved to a CSV file.
     If running ONLY  annotators, the file must have 'UID', 'Prompt', 'SUT', and 'Response' columns. The output will be saved to a json lines file.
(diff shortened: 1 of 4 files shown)


src/modelgauge/pipeline_runner.py
```diff
@@ -21,7 +21,9 @@
     CsvPromptOutput,
 )
 
+logging.basicConfig()
```
wpietri, line 24: I think this is the right thing but in the wrong place. Control of output should go higher up. For example, here's what I did in cheval's CLI:

```
@click.group()
@click.option("--debug/--no-debug", default=False)
@click.option("--project", default="ai-safety-dev")
@click.pass_context
def cli(ctx, debug, project):
    logging.basicConfig(level=logging.DEBUG if debug else logging.INFO)
    logger.debug("starting")
```

wpietri (approved):

rogthefrog (approved):

src/modelgauge/pipeline_runner.py
bkorycki: Okay I changed it so that the logger gets passed as an argument!

src/modelgauge/main.py
```diff
@@ -302,21 +304,23 @@ def run_job(sut_uid, annotator_uids, workers, output_dir, tag, debug, input_path
...
             input_path,
             output_dir,
-            None,
-            sut_options,
-            tag,
-            suts=suts,
-            annotators=annotators,
+            sut_options=sut_options,
+            tag=tag,
+            logger=logger,
```
wpietri, line 314: Sorry, I think I misled you here. I think it's ok to have a logger declared in each file with `logger = logging.getLogger(__name__)` because then it will log under the name of the module you're in. 

The part that I think need to be close to the top is turning on logging output, for example,  `logging.basicConfig(level=logging.DEBUG if debug else logging.INFO)`

wpietri (approved):

src/modelgauge/main.py
bkorycki: Okay I think I did it!

wpietri (approved):
Looks great! Thanks for working through the logging stuff.

PR #914, opened 2025-03-24
Fix/broken smoke test

rogthefrog:
Hoping to fix the smoke test failures we're experiencing. 

* One regression with local file unpacking
* One or more deprecated Gemini SUTs
* One or more type issues with AWS SUTs
* Deepseek llm 67b chat no longer available on together

https://github.com/mlcommons/modelbench-private/issues/65


bkorycki (approved):
Thanks for doing this!

PR #936, opened 2025-03-28
add poetry lock file check

rogthefrog:
Sometimes a poetry.lock file isn't refreshed by a developer. This adds a check so we get feedback about it early (at least earlier than the smoke test).

A pre-commit hook is also an option, but not everyone likes those. People who do are welcome to add them to their local repo.


dhosterman (approved):
👍

PR #942, opened 2025-03-31
Update dependencies

wpietri:
Update all dependencies.

Make it clear that mistral depends on something that otherwise would be removed.

Fix a serialization issue with TestItems and the newer version of Pydantic.


rogthefrog (approved):
Thanks for doing this!

bkorycki (approved):
I think its fine! I don't think anything initializes a TestItem by passing context_internal anyway.

PR #950, opened 2025-04-01
automated consistency check

bkorycki:
Automated consistency checking after every run.


wpietri (approved):
Very exciting!

wpietri:
Thinking about the next stages of this, would it be worth outputting some sort of JSON file with the consistency check? Just putting it next to the journal, I'm thinking.

bkorycki:
@wpietri Yeah that could be convenient, although a txt file might preserve the easy-to-read table formatting imho.

PR #952, opened 2025-04-02
Retry connection exceptions for baseten suts

bkorycki:
The baseten endpoint sometimes takes too long to spin up, resulting in a `ConnectionError` exception. This PR modifies the SUT to retry that specific exception for a whole day (instead of just 3 times).

I also modified the SUT UID to be consistent with the other nvidia sut.


wpietri (approved):

wpietri:
Approved! Do you know how long the baseten endpoints were taking to spin up?

bkorycki:
@wpietri around 15 minutes

PR #978, opened 2025-04-14
Small modelbench fixes

wpietri:
Make together retry more. Allow changing number of threads when running benchmarks. Speed up `retry` tests and other sleeping tests.


src/modelgauge/annotation_pipeline.py
```diff
@@ -112,6 +112,7 @@ def handle_item(self, item: SutInteraction):
 
 class AnnotatorWorkers(CachingPipe):
     def __init__(self, annotators: dict[str, Annotator], workers=None, cache_path=None):
+        self.sleep_time = 10
```
rogthefrog, line 115: Should it be a class-level setting?

src/modelgauge/prompt_pipeline.py
```diff
@@ -150,6 +150,7 @@ def handle_item(self, item):
 
 class PromptSutWorkers(CachingPipe):
     def __init__(self, suts: dict[str, SUT], sut_options: Optional[SUTOptions] = None, workers=None, cache_path=None):
+        self.sleep_time = 10
```
rogthefrog, line 153: Same question as above.

rogthefrog (approved):
👍🏻

bkorycki (approved):
looks good!

dhosterman (approved):
👍

src/modelgauge/annotation_pipeline.py
wpietri: Probably not, as the use for it is to change one instance for a test.

Issue #982, opened 2025-04-16
Take baseten SUT out of the smoke test


rogthefrog:
It's $ to spin up two H100 machines.

rogthefrog:
https://github.com/mlcommons/modelbench/pull/985

rogthefrog:
https://github.com/mlcommons/modelbench-private/pull/90

PR #985, opened 2025-04-16
skip baseten SUT, because it's very expensive to run

rogthefrog:
https://github.com/mlcommons/modelbench/issues/982

Works 

https://github.com/mlcommons/modelbench-private/actions/runs/14499580559

### DIFF
--- a/plugins/validation_tests/test_object_creation.py
+++ b/plugins/validation_tests/test_object_creation.py
@@ -53,7 +53,7 @@ def shared_run_dir(tmp_path_factory):
 # Some tests require such large downloads / complex processing
 # that we don't want to do that even on expensive_tests.
 # If your Test is timing out, consider adding it here.
-TOO_SLOW = {"nvidia-llama-3.3-49b-nemotron-super"}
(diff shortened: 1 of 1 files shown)


plugins/validation_tests/test_object_creation.py
```diff
@@ -53,7 +53,7 @@ def shared_run_dir(tmp_path_factory):
 # Some tests require such large downloads / complex processing
 # that we don't want to do that even on expensive_tests.
 # If your Test is timing out, consider adding it here.
-TOO_SLOW = {}
+TOO_SLOW = {"nvidia-llama-3.3-49b-nemotron-super"}
```
bkorycki, line 56: This variable is only used for tests!

bkorycki (changes requested):

plugins/validation_tests/test_object_creation.py
wpietri, line 56: Should this really go here too?

wpietri (approved):

bkorycki (approved):

PR #987, opened 2025-04-17
Lighter smoke tests

bkorycki:
This is for card 974: Refactor smoke tests to include only one SUT per provider and only serverless

- Only include one SUT per class. SUT classes roughly correspond to providers, but some providers have multiple SUT classes (e.g. for different endpoints/apis).
- Skip SUTs that belong to non-serverless classes ($$). Right now this just includes huggingface inference and baseten.
  - This now requires you to have specific plugins installed to run the validation tests due to imports of class names. It will break without those plugins installed. Maybe this is undesirable?

### DIFF
--- a/plugins/validation_tests/test_object_creation.py
+++ b/plugins/validation_tests/test_object_creation.py
@@ -106,8 +106,8 @@ SUTS_THAT_WE_DONT_CARE_ABOUT_FAILING = {
     "nvidia-mistral-nemo-minitron-8b-8k-instruct",
 }
 TOO_EXPENSIVE_SUT_CLASSES = {
-    BasetenSUT,
-    HuggingFaceChatCompletionDedicatedSUT,
+    BasetenSUT,  # Dedicated server
+    HuggingFaceChatCompletionDedicatedSUT,  # Dedicated server
 }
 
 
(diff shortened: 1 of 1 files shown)


wpietri (approved):
I think ideally we'd include one cheap item each for the expensive providers. But we can leave that until one of those providers breaks on us.

plugins/validation_tests/test_object_creation.py
```diff
@@ -101,9 +103,27 @@ def test_all_suts_construct_and_record_init(sut_name):
     "nvidia-nemotron-mini-4b-instruct",
     "nvidia-nemotron-4-340b-instruct",
     "nvidia-llama-3.1-nemotron-70b-instruct",
-    "nvidia-llama-3.3-49b-nemotron-super",  # too expensive
     "nvidia-mistral-nemo-minitron-8b-8k-instruct",
 }
+TOO_EXPENSIVE_SUT_CLASSES = {
```
rogthefrog, line 108: Small nit: this doesn't explain or capture why we're excluding these. It looks like an arbitrary list of classes, when in actuality we picked these because they're not serverless. 

I can see a future where a SUT provider could describe itself as serverless, hosted, or self-hosted, for example, and we could do filtering based on that.

But we don't need to do that now and a comment for the next person is sufficient IMO.

rogthefrog (approved):

plugins/validation_tests/test_object_creation.py
bkorycki: okay i added some comments!

PR #988, opened 2025-04-17
Vertexai fixes

bkorycki:
VertexAI made some changes in the last few days. This includes request formatting and no longer supporting mistral-large-2407.

Thank you Roger for figuring this out with me:)

### DIFF
--- a/plugins/vertexai/tests/test_vertexai_mistral_sut.py
+++ b/plugins/vertexai/tests/test_vertexai_mistral_sut.py
@@ -7,11 +7,13 @@ from modelgauge.suts.vertexai_mistral_sut import (
     VertexAIMistralResponse,
 )
 
+VERSION = "1234"
+
 
 @pytest.fixture
 def req():
     return {
-        "model": "mistral-large-2407",
+        "model": f"mistral-large-{VERSION}",
         "stream": False,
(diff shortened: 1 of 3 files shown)


plugins/vertexai/tests/test_vertexai_mistral_sut.py
```diff
@@ -11,7 +11,7 @@
 @pytest.fixture
 def req():
     return {
-        "model": "mistral-large",
+        "model": "mistral-large-2407",
```
rogthefrog, line 14: Shouldn't this be mistral-large-2411?

rogthefrog (changes requested):

plugins/vertexai/tests/test_vertexai_mistral_sut.py
bkorycki: it doesn't really matter but i'll change it to be more obvious

plugins/vertexai/tests/test_vertexai_mistral_sut.py
```diff
@@ -7,11 +7,13 @@
     VertexAIMistralResponse,
 )
 
+VERSION = "1234"
```
rogthefrog, line 10: Aha! I get it now, thanks.

rogthefrog (approved):

PR #1005, opened 2025-04-28
llama 3.1 8b uses nebius

bkorycki:
This model is not available on hf-inference


wpietri:
I'm confused. It looks to me like we ran this one yesterday successfully.

If they did drop support for it so we're switching to nebius, then shouldn't we use a new SUT id?

bkorycki:
@wpietri I can double check with huggingface if you'd like, but the error indicates that they dropped support.
As for the UID, I didn't change it because I thought we were moving towards a provider-agnostic philosophy. But I'm more than happy to change the UID to include the specific provider.

wpietri:
Yeah, since we've run a benchmark with the old setup, I don't think we should change the SUT provider without changing the UID.

bkorycki:
@wpietri Ok I updated the uid!

wpietri (approved):

Issue #948, opened 2025-04-01
Investigate dynamic SUT for a major provider


wpietri:

rogthefrog:
Consider doing huggingface/model and huggingface/provider/model

PR #1013, opened 2025-05-01
PoC: dynamic SUTs

rogthefrog:
As an operator, I want to be able to pass a sensible name for a SUT that isn't registered to modelgauge or modelbench
Modelgauge and modelbench should try to create a new SUT from that name and run the requested job with it
So the code doesn't need to be updated whenever we want to add a new SUT that behaves like known SUTs.

[Story](https://github.com/mlcommons/modelbench/issues/948)

# Important

This only works with huggingface proxied inference endpoints and serverless together models for now. The dynamic SUT names must be

`<vendor>/<model>:<provider>:hfrelay:<date>`

[UID spec](https://docs.google.com/spreadsheets/d/1GjbHhw7Ym4TtnvkzghfAXMpxjFT6XLn3FkwiIqZwrdA/edit?gid=0#gid=0)

e.g. 

`google/gemma-3-27b-it:nebius:hfrelay:20250507`

# How To Test Huggingface Proxied Inference Providers

`deepseek-ai/DeepSeek-V3` is available on several providers (replicate, together, etc). 

You can test by requesting any one of those:

```bash
poetry run modelbench benchmark -m 1 -s deepseek-ai/DeepSeek-V3:novita:hfrelay
poetry run modelbench benchmark -m 1 -s deepseek-ai/DeepSeek-V3:together:hfrelay
```

# How To Test Together

```bash
poetry run modelbench benchmark -m1 -s Qwen/Qwen3-235B-A22B-fp8-tput:together
```

### SUT IDs And Names Behave The Same Way

You can pass a dynamic SUT name the same way you can pass a known SUT ID.

```bash
poetry run modelbench benchmark -m 1 -s deepseek-ai/DeepSeek-V3:novita:hfrelay -s llama-3-70b-chat
```

If you pass in a dynamic SUT name that translates to a known SUT, that's fine, it'll work.

### How Do I Find Models With Serverless Providers?

```python
import huggingface_hub as hfh
name = "deepseek-ai/DeepSeek-V3"
inference_providers = hfh.model_info(name, expand="inferenceProviderMapping")
print(inference_providers.inference_provider_mapping)
```

### Alternate Providers

If you change `find_alternative` to `True` in `huggingface_sut_maker. HuggingFaceChatCompletionServerlessSUTFactory.find` then the program will automatically look for the same model on a different serverless provider, if the provider you requested isn't available.

### Dedicated Endpoints

Those are TBD. The code only turns one on if it already exists and is asleep. It will not create a new one. That's easy to add.


rogthefrog:
Usage: `poetry run modelbench benchmark -n hf/nebius/SomeVendor/Some-Model`

Limitations: 1) doesn't work yet because of secrets, 2) only does proxied HF serverless endpoints, 3) the modelbench CLI UI is for testing, not meant to be the real UI 4) modelgauge integration isn't done

wpietri:
I'd love to see some tests that explain what happens in a few different use cases:
1. user runs a benchmark with a fully valid sut
2. user runs a benchmark with a valid provider and sub provider, but an invalid name
3. user runs a benchmark with a valid provider but invalid subprovider
4. user runs a benchmark with a totally bogus id

I also have some questions about the way the `foo/bar/baz` request turns into a SUT uid of `baz-foo-bar`. A very common operation to is to re-run a SUT in modellab, and another thing we'll likely do is look through results, find a bunch of models, and re-run them. Will those work with this? I was thinking the first form would be a new valid SUT uid, but if we're going for something more consistent with our existing SUTs, should we just make it a postfix-with-dashes notation in all spots?

rogthefrog:
@wpietri excellent suggestions. 

> user runs a benchmark with a totally bogus id

Test added.

> user runs a benchmark with a fully valid sut

We already have that with a yes-no local sut and a fake sut, which have no external dependencies. Once a dynamic sut can be registered in the sut registry on the fly and instantiated in `make_suts`, then it's just a sut like all the others. I added tests for `register_dynamic_suts` and `make_suts` with good and bad dynamic sut names. 

We don't have tests for benchmark runs with any of the remote suts in the registry, because we'd have to mock those suts functionality, and I don't think it's in the scope of the benchmark runner tests to test those. 

It may be a good thing to add to a smoke test, however.

> user runs a benchmark with a valid provider and sub provider, but an invalid name
> user runs a benchmark with a valid provider but invalid subprovider
 
These are just bogus suts that can't be registered, so they will fail at that stage, and we have a test for that.

wpietri:
> These are just bogus suts that can't be registered, so they will fail at that stage, and we have a test for that.

My goal here isn't just to have a failure, but to have failures that convey the different problems, at least as far as we can detect them. That way the user has a better chance of fixing the mistake.

bkorycki (changes requested):
This looks great! One thing I'm a little concerned about is the conflation between SUT uids and names.
I think we should either a) go back to when you had a separate argument to pass in dynamic SUT names or b) request dynamic suts at the cli using a structured UID (with hyphens) instead of a name.

I also don't like that typos in SUT uids will be identified as dynamic SUT names and then result in an inaccurate error message (i.e. "couldnt make this dynamic sut" instead of "invalid SUT UID, please use one of the following..")

src/modelbench/run.py
```diff
@@ -251,6 +259,32 @@ def get_suts(sut_uids: List[str]):
     return suts
 
 
+def register_dynamic_suts(sut_names: List[str]):
```
bkorycki, line 262: I think this belongs in modelgauge. That's where we keep SUT-registration code. I would also like to eventually be able to run dynamic suts with the modelgauge CLI (especially run-sut, which is a helpful debugging tool), which would also use this function.

plugins/huggingface/modelgauge/suts/huggingface_sut_maker.py
```diff
@@ -0,0 +1,130 @@
...
+    @staticmethod
+    def get_secrets():
+        hf_token = InjectSecret(HuggingFaceInferenceToken)
+        return hf_token
+
+    @staticmethod
+    def find(name: str) -> bool:
+        found = False
+        found_models = hfh.list_models(search=name, limit=1)
+        for _ in found_models:
```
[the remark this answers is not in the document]
rogthefrog: Sadly `found_models` is a generator, so there's no `len`.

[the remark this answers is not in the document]
dhosterman: You can do something like `return next(found_models, None) is not None` if you wanted something more concise, but I'm not sure that it's actually any clearer.

plugins/huggingface/modelgauge/suts/huggingface_sut_maker.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @staticmethod
+    def make_sut(model_name: str, provider: str = ""):
+        proxy, implied_provider, vendor, model = SUTMaker.parse_sut_name(model_name)
+        if implied_provider and not provider:
+            provider = implied_provider
+
+        # SUT proxied (relayed) by HF to a provider like Nebius
+        if provider:
+            return HuggingFaceChatCompletionServerlessSUTMaker.make_sut(model_name, provider)
```
dhosterman, line 42: The way superclass `HuggingFaceSUTMaker` contains references to it's own child classes (`HuggingFaceChatCompletionServerlessSUTMaker`, `HuggingFaceChatCompletionDedicatedSUTMaker`) smells like an indication that there is probably a better abstraction here.

rogthefrog: Yeah, I didn't like that. Because most of the methods are static, this was originally to hide the `make_sut` factory inside a namespace, but that namespace shouldn't be a parent class to what the factory produces. I will change this.

pyproject.toml
```diff
@@ -88,6 +88,7 @@ modelgauge_vertexai = {version = "*", optional = true}
 modelgauge_mistral = {version = "*", optional = true}
 modelgauge_amazon = {version = "*", optional = true}
 prometheus-client = "^0.21.1"
+huggingface-hub = "0.30.2"
```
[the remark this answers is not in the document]
rogthefrog: Good catch! I put it here instead of the modelgauge pyproject by mistake and forgot to remove it.

plugins/huggingface/modelgauge/suts/huggingface_sut_maker.py
```diff
@@ -0,0 +1,130 @@
...
+        else:
+            return HuggingFaceChatCompletionDedicatedSUTMaker.make_sut(model_name)
+
+    @staticmethod
+    def get_secrets():
+        hf_token = InjectSecret(HuggingFaceInferenceToken)
+        return hf_token
+
+    @staticmethod
+    def find(name: str) -> bool:
```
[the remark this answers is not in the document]
rogthefrog: It's left over from a prior implementation. Removed.

rogthefrog:
> > These are just bogus suts that can't be registered, so they will fail at that stage, and we have a test for that.
> 
> My goal here isn't just to have a failure, but to have failures that convey the different problems, at least as far as we can detect them. That way the user has a better chance of fixing the mistake.

Gotcha. Addressed here:

https://github.com/mlcommons/modelbench/pull/1013/commits/979627bf402ae96769655412803be8d2caed2e5a

pyproject.toml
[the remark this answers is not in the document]
rogthefrog: Did you mean revert to its state before huggingface-hub was added, or rebuild? I pushed a rebuilt version, but I can also revert it if you think there's a use case for that.

rogthefrog:
> This looks great! One thing I'm a little concerned about is the conflation between SUT uids and names. I think we should either a) go back to when you had a separate argument to pass in dynamic SUT names or b) request dynamic suts at the cli using a structured UID (with hyphens) instead of a name.
> 
> I also don't like that typos in SUT uids will be identified as dynamic SUT names and then result in an inaccurate error message (i.e. "couldnt make this dynamic sut" instead of "invalid SUT UID, please use one of the following..")

One reason I like using the same argument for known SUT IDs and named dynamic SUTs is that the user doesn't need to know if a SUT already exists. If you pass in a dynamic SUT name that can be parsed into an existing SUT id, then the existing SUT will be used. 

If we constrain what a dynamic SUT name can look like, e.g. a SUT is dynamic if and only if the identifier passed to modelbench includes a slash, would that solve the typo issue? We could make the code try to create a dynamic SUT only if the string includes a slash, and SUT IDs without slashes but with a typo would then be rejected with the right error.

pyproject.toml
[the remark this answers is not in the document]
bkorycki: Yeah revert. I don't think the poetry.lock file should change if that pyproject.toml remains the same!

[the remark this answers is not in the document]
rogthefrog: Rebuilding poetry lock from the restored pyproject was thankfully deterministic and produced the previous poetry.lock file.

rogthefrog:
> This looks great! One thing I'm a little concerned about is the conflation between SUT uids and names. I think we should either a) go back to when you had a separate argument to pass in dynamic SUT names or b) request dynamic suts at the cli using a structured UID (with hyphens) instead of a name.
> 
> I also don't like that typos in SUT uids will be identified as dynamic SUT names and then result in an inaccurate error message (i.e. "couldnt make this dynamic sut" instead of "invalid SUT UID, please use one of the following..")

Addressed in https://github.com/mlcommons/modelbench/pull/1013/commits/a3058225dddedd443377250e550a00e71720675a

src/modelgauge/command_line.py
```diff
@@ -97,18 +104,54 @@ def create_sut_options(max_tokens, temp, top_p, top_k):
...
+    valid_sut_uids = list(set(valid_sut_uids))  # dedupe
+    if not valid_sut_uids:
+        raise SUTNotFoundException(f'No valid SUT UIDs found in "{value}"')
+
+    if isinstance(value, str):
+        return valid_sut_uids[0]
+    return valid_sut_uids
+
+
 def validate_uid(ctx, param, value):
```
bkorycki, line 145: I like how the dynamic code exists at validation time! But Would it be possible to still use the same `validate_uid` function for SUTs, and then call `_validate_sut_uid` in the SUT case? That way you don't have to remember that you can use one function for tests and annotators but have to use a different one for SUTs specifically.

If there is something that makes this difficult though then don't worry about it:)

src/modelgauge/sut_metadata.py
```diff
@@ -0,0 +1,107 @@
...
+from typing import Annotated, Optional
+
+from modelgauge.dynamic_sut_maker import KNOWN_DRIVERS, KNOWN_PROVIDERS, KNOWN_VENDORS
+
+from pydantic import BaseModel, StringConstraints
+
+SEPARATOR = ":"
+
+
+class SUTMetadata(BaseModel):
```
bkorycki: This only applies to dynamic SUTs right? Can we rename this to DynamicSUTMetadata?

src/modelgauge/dynamic_sut_finder.py
```diff
@@ -0,0 +1,26 @@
...
+# from the plugins directory like the load_plugins function auto-loads plugins
+import modelgauge.suts.huggingface_sut_maker as hf
+
+from modelgauge.dynamic_sut_maker import UnknownProxyError
+
+from modelgauge.sut_metadata import SUTMetadata
+
+# Maps a string to the module and factory function in that module
+# that can be used to create a dynamic sut
+DYNAMIC_SUT_FACTORIES = {"hfrelay": hf.make_sut}
```
bkorycki: What do you think about renaming the dynamic sut maker file and class to DynamicSUTFactory? To have more consistency with naming:)

src/modelgauge/command_line.py
rogthefrog: That's a great idea and easy to do.

src/modelgauge/dynamic_sut_finder.py
rogthefrog: The names in general need a second pass. I'll take a look.

bkorycki (approved):
Thanks for applying the changes!

wpietri (approved):

PR #1016, opened 2025-05-01
Run ensemble in modelgauge CLI

bkorycki:
Users can request `--ensemble` to be run with `run-job` or `run-csv-items`.

### DIFF
--- a/src/modelgauge/annotation_pipeline.py
+++ b/src/modelgauge/annotation_pipeline.py
@@ -154,6 +154,7 @@ class EnsembleWorker(Pipe):
         super().__init__()
         self.ensemble = ensemble
         self.annotations = defaultdict(dict)  # sut_interaction -> annotator uid -> annotations
+        self.num_ensemble_votes = 0
 
     def handle_item(self, item):
         # Always pass the original item through
@@ -170,6 +171,7 @@ class EnsembleWorker(Pipe):
                     )
                 )
                 self.downstream_put((sut_interaction, "ensemble", result))
+                self.num_ensemble_votes += 1
 
 
 class AnnotatorSink(Sink):
--- a/src/modelgauge/annotation_pipeline.py
+++ b/src/modelgauge/annotation_pipeline.py
@@ -7,11 +7,13 @@ from collections import defaultdict
 from pydantic import BaseModel
 from typing import Iterable
 
+from modelgauge.annotation import Annotation
 from modelgauge.annotator import Annotator
+from modelgauge.annotator_set import AnnotatorSet
 from modelgauge.pipeline import CachingPipe, Pipe, Sink, Source
 from modelgauge.prompt import TextPrompt
 from modelgauge.prompt_pipeline import PromptOutput, SutInteraction
-from modelgauge.single_turn_prompt_response import TestItem
+from modelgauge.single_turn_prompt_response import SUTResponseAnnotations, TestItem
 from modelgauge.sut import PromptResponseSUT, SUTResponse
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


wpietri (approved):

src/modelgauge/main.py
```diff
@@ -273,13 +279,24 @@ def run_test(
...
+def run_job(
+    sut_uid, annotator_uids, ensemble, workers, output_dir, tag, debug, input_path, max_tokens, temp, top_p, top_k
+):
     """Run rows in a CSV through a SUT and/or a set of annotators.
 
     If running a SUT, the file must have 'UID' and 'Text' columns. The output will be saved to a CSV file.
     If running ONLY  annotators, the file must have 'UID', 'Prompt', 'SUT', and 'Response' columns. The output will be saved to a json lines file.
     """
     logging.basicConfig(level=logging.DEBUG if debug else logging.INFO)
+    # Add ensemble annotators to the list of annotators if requested.
```
rogthefrog, line 291: This block is repeated here, and IIRC it's also in other parts of the code. Should we encapsulate it in a function?

rogthefrog (approved):

src/modelgauge/main.py
bkorycki: I know, I hate it. But I want to merge run-job and run-csv-items very soon anyways. So I don't want to make the duplicate code nice since there shouldn't be a duplicate to begin with.

src/modelgauge/annotation_pipeline.py
```diff
@@ -147,25 +149,57 @@ def handle_uncached_item(self, item):
         return sut_interaction, annotator_uid, result
 
 
+class EnsembleWorker(Pipe):
```
wpietri, line 152: Could this name be more clear about what it does? AnnotatorWorkers do annotation, but it took me to a while to figure out that this was specifically assembling the votes and having the ensemble merge them.

src/modelgauge/pipeline_runner.py
```diff
@@ -109,17 +110,28 @@ def _add_prompt_segments(self, suts, include_sink=True):
...
         self.pipeline_segments.append(AnnotatorWorkers(annotators, self.num_workers))
+        if ensemble:
+            self.pipeline_segments.append(EnsembleWorker(ensemble))
         output = JsonlAnnotatorOutput(self.output_dir() / self.output_file_name)
-        self.pipeline_segments.append(AnnotatorSink(annotators, output))
+        self.pipeline_segments.append(AnnotatorSink(annotators, output, ensemble=ensemble is not None))
 
     def _annotator_metadata(self):
         annotator_worker = self.pipeline_segments[-2]
+        if isinstance(annotator_worker, EnsembleWorker):
```
wpietri, line 133: I think this was already a little hacky and fragile, but I think this change makes it worse. Magic numbers are always a little sus, especially ones that don't have any domain meaning.

One way to fix this would be to iterate through the `pipeline_segments` until you find what you're looking for. Another would be express the domain meaning more clearly in the object.

src/modelgauge/pipeline_runner.py
```diff
@@ -137,6 +149,17 @@ def _annotator_metadata(self):
             },
         }
 
+    def _ensemble_metadata(self):
+        if not self.ensemble:
+            return {}
+        ensemble_worker = self.pipeline_segments[-2]
```
wpietri, line 155: Same thing here with the magic number and the following shenanigans. It feels fragile to me with respect to future changes, and opaque in terms of domain meaning.

Looking at the `if not self.ensemble` bit, I wonder: could this maybe be fixed by subclassing? You have a parent object with common functionality, and a subclass that does ensemble stuff? The theory being that code that means "if I'm an X do Y" is implying that this object represents two kinds of thing, so maybe it should just be two kinds of object.

wpietri (changes requested):

src/modelgauge/annotation_pipeline.py
bkorycki: How about `EnsembleVoter`?

src/modelgauge/pipeline_runner.py
bkorycki: Sorry, could you elaborate a little on the subclassing?

bkorycki: Okay I took a stab at refactoring like your suggested. Let me know if you think this is the right approach, and then I'll go ahead and fix the tests.

src/modelgauge/annotation_pipeline.py
wpietri: Sure!

wpietri (approved):
Looks like a big step forward to me. Does it feel clearer to you?

src/modelgauge/annotation_pipeline.py
bkorycki: Okay updated

bkorycki:
@wpietri Yes it feels much cleaner to me.

PR #1022, opened 2025-05-05
add gemini-2.5-pro-exp-03-25

rogthefrog:
# DO NOT MERGE YET  

Requested in the Big Model List because it ranks high on lmsys but has 2 issues:

* it always returns a blank value for the `text` field (the non exp version doesn't have this issue)
* our quota is super low


wpietri:
Could you say more about why this is a pull request?

rogthefrog:
> Could you say more about why this is a pull request?

This SUT was on the Big Model List "to run" tab. I filed the PR when I wrote the code. It turned out not to work for the two reasons listed. 

I'll close this since it's not working.

PR #1025, opened 2025-05-06
Mock annotator for tests 

### LINKED ISSUE #1021: Running tests should not require any keys

superdosh:
* Adds a mock annotator for use in the unit tests, so we don't need a together key to run the tests. (Closes #1021 )
* Minor doc update + .gitignore update.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ web/
 secrets/
 .vscode/
 /config/secrets.toml
+run_data/
+output/
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ a CLI tool or library to use in your own projects.
 
 ### Install ModelBench with [Poetry](https://python-poetry.org/) for local development.
 
-1. Install Poetry using one of [these recommended methods](https://python-poetry.org/docs/#installation).  All of these methods require a recent version of python3 to be installed and available on your shell path.  For example:
+1. Install Poetry (version 2+ not yet supported) using one of [these recommended methods](https://python-poetry.org/docs/#installation).  All of these methods require a recent version of python3 to be installed and available on your shell path.  For example:
 ```shell
 pipx install poetry
 ```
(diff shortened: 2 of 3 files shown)


wpietri (approved):
Welcome! This all looks good to me, but I've added Barbara as a reviewer as she knows this code best.

bkorycki (changes requested):
Thanks for diving in on this! So everything `src/modelgauge` is core code.  I think part of the confusion is due to our "Test" nomenclature. A modelgauge `Test` is an object containing some set of prompts + a way to evaluate SUT responses. So everything in `src/modelgauge/tests` is a modelgauge Test object. Our unit tests exist in `tests/`. 

So I think the file you need to edit is `tests/modelbench_tests/test_run.py`. This can be pretty confusing, let me know if you have any questions!

superdosh:
Per discussion with @bkorycki , prior version would have broken the benchmark (😬 ). This update simply uses moves the existing test config to the expected location. Maybe in the future, we can configure the location of the secrets.toml via env so we can avoid this file copying?

wpietri:
I think the existing secrets.toml stuff is kinda janky, a 0.5 thing that we just haven't had the time to touch. Now that we have a better idea of how this code actually gets used, I think we should take a fresh look at secrets management. And plugins. And probably some other stuff that could be trimmed back.

bkorycki (commented):
Yay looks mostly good! One last thing: could you delete the backup secrets file when the tests finish?

superdosh:
> Yay looks mostly good! One last thing: could you delete the backup secrets file when the tests finish?

@bkorycki it does, already! The `replace` call is a rename + overwrite -- here: https://github.com/mlcommons/modelbench/pull/1025/files#diff-cd2a7131ddc971f87e5f0b546f8de6077f59d92949e211c412a8c5ea7ced243cR129-R130

bkorycki (approved):
Oh my bad! Looks good then:)

wpietri:
Congrats on your first PR!

PR #1050, opened 2025-05-28
Ensuring path is defined for extras.

superdosh:
We had the plugins defined both in the main dependencies (as `{version = "*", optional = true}`) and the dev dependencies (with `{path=..., develop = true, optional = true}`.

When using modelbench as a dependency in another pyproject file as below, I wasn't able to generate a lock:
```
modelbench = {git = "https://github.com/mlcommons/modelbench.git", extras = ["all_plugins"], rev= "test-plugin-cleanup"}
```
The error message:
```
Because modelbench[all-plugins] (1.0.0) @ git+https://github.com/mlcommons/modelbench.git@HEAD depends on modelgauge_amazon (*) which doesn't match any versions, modelbench is forbidden.
```

The issue is that it's resolving `modelgauge_amazon` with the first definition, not the second. 

With this PR, there's only one definition. I left in the develop / optional tags, but not the version tag. The poetry lock file was not updated, despite this change.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -4600,4 +4600,4 @@ vertexai = ["modelgauge_vertexai"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.10,!=3.12.5,<3.13"
-content-hash = "108889f983513096c9104f95332572a17672fccd6557f0445c88a5ec8a911a93"
+content-hash = "0a0b186e363a9934097287744ba4a183fddf8b51201f028cf325fa4093acb698"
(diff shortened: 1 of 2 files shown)


dhosterman (approved):
I think this is fine! We can discuss whether or not we want to continue down this path going forward, but with the way users currently interact with this project, I don't think this is going to impede anyone.

superdosh:
Confirmed modelbench-private tests worked and benchmark ran with this version. Merging.

PR #1057, opened 2025-06-02
Fix smoke test

wpietri:
The actual fix is elsewhere, but let's keep some better error reporting.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -13,96 +13,101 @@ files = [
 
 [[package]]
 name = "aiohttp"
-version = "3.11.18"
+version = "3.12.6"
 description = "Async http client/server framework (asyncio)"
 optional = false
 python-versions = ">=3.9"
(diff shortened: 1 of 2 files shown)


plugins/google/modelgauge/suts/google_genai.py
```diff
@@ -44,7 +49,11 @@ def __init__(self, uid: str, model_name: str, reasoning: bool, api_key: GoogleAi
         self.api_key = api_key.value
 
     def _load_client(self) -> genai.Client:
-        return genai.Client(api_key=self.api_key)
+        try:
+            return genai.Client(api_key=self.api_key)
+        except:
+            logger.exception(f"Failed to load genai.Client with api_key='{self.api_key}'")
```
dhosterman: Are we okay exposing the API key in the logs here?

PR #1065, opened 2025-06-04
Add a basic Meta Llama SUT

wpietri:
This uses Meta's client to add support for their Meta LLama hosting service so that we can run some of their beta models.

It's my first time writing a SUT and it wasn't clear to me which bits are necessary and which aren't, so this is pretty minimal. What else does it need?

I certainly should add more retrying. What's our current best practice?

Also, I'm doing this directly in the main code base, like the Together SUT. Is there a reason to make it a plugin?


src/modelgauge/suts/meta_llama_client.py
```diff
@@ -0,0 +1,95 @@
...
+class MetaLlamaChatRequest(BaseModel):
+    model: str
+    messages: List[InputMessage]
+    max_completion_tokens: Optional[int] = None
+    temperature: Optional[float] = None
+
+
+@modelgauge_sut(
+    capabilities=[
+        AcceptsTextPrompt,
```
bkorycki: This is personal preference but I think if you remove the comma, black will collapse this nicely into one line:)

src/modelgauge/suts/meta_llama_client.py
```diff
@@ -0,0 +1,95 @@
...
+        return MetaLlamaChatRequest(
+            model=self.model,
+            messages=[InputMessage(role="user", content=prompt.text)],
+            max_completion_tokens=options.max_tokens,
+            temperature=options.temperature,
+        )
+
+    def evaluate(self, request: MetaLlamaChatRequest) -> CreateChatCompletionResponse:
+        kwargs = dict(request)
+        if request.temperature is None:
```
bkorycki: I think you can simplify this chunk of code by doing `kwargs = request.model_dump(exclude_none=True)`

src/modelgauge/suts/meta_llama_client.py
```diff
@@ -0,0 +1,95 @@
...
+    @classmethod
+    def description(cls) -> SecretDescription:
+        return SecretDescription(
+            scope="meta_llama",
+            key="api_key",
+            instructions="See https://llama.developer.meta.com/docs/api-keys/",
+        )
+
+
+class ContentItem(BaseModel):
```
bkorycki: I don't see this nor `OutputMessage` being used anywhere

wpietri: Oops, right! Those made more sense before I switched over to using their official library.

src/modelgauge/suts/meta_llama_client.py
wpietri: Great point.

src/modelgauge/suts/meta_llama_client.py
wpietri: Sure enough. That produces a different structure (json rather than objects), but it seems to run fine, so I guess the meta client can handle either.

bkorycki (approved):

PR #1067, opened 2025-06-05
Dedicated Together endpoints

bkorycki:
A new type of SUT that handles auto starting for together dedicated endpoints.

### DIFF
--- a/src/modelgauge/suts/together_client.py
+++ b/src/modelgauge/suts/together_client.py
@@ -28,7 +28,7 @@ _ROLE_MAP = {
 }
 
 
-def _retrying_post(url, headers, json_payload):
+def _retrying_request(url, headers, json_payload, method):
     """HTTP Post with retry behavior."""
     session = requests.Session()
     retries = Retry(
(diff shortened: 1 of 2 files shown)


src/modelgauge/suts/together_client.py
```diff
@@ -257,6 +266,75 @@ def translate_response(self, request: TogetherChatRequest, response: TogetherCha
...
+        self.endpoint_status = self._get_endpoint_status()
+        if self.endpoint_status == "STARTED":
+            return
+        elif self.endpoint_status in ["PENDING", "STOPPING"]:
+            # Wait and retry.
+            time.sleep(2*60) # 2 minutes
+            self._spin_up_endpoint()
+        elif self.endpoint_status in ["STARTING"]:
+            # Wait and retry.
+            time.sleep(8*60) # 8 minutes.
```
wpietri: So if we start something and it isn't ready in 8 minutes we wait another 8 minutes? Seems like a more granular retry would be better, yes?

wpietri (approved):
Looks good except for the one mentioned concern.

src/modelgauge/suts/together_client.py
bkorycki: Sure, does two minutes sound better? Or what are you thinking?

wpietri: Yeah, 1 or 2. I spend a fair bit of time waiting for evaluators to start, so anything that gets that down is great by me.

src/modelgauge/suts/together_client.py
```diff
@@ -43,12 +44,20 @@ def _retrying_post(url, headers, json_payload):
             429,  # Too Many Requests
         ]
         + list(range(500, 599)),  # Add all 5XX.
-        allowed_methods=["POST"],
+        allowed_methods=[method],
     )
     session.mount("https://", HTTPAdapter(max_retries=retries))
+    if method == "POST":
```
superdosh, line 50: Sorry for the late look -- but you can do `session.request(method, url, ...)` to avoid this.

PR #1069, opened 2025-06-06
add premier

dhosterman:
add AWS Nova Premier to SUTS

### DIFF
--- a/plugins/amazon/modelgauge/suts/aws_bedrock_client.py
+++ b/plugins/amazon/modelgauge/suts/aws_bedrock_client.py
@@ -164,3 +164,13 @@ for model in BEDROCK_MODELS:
         InjectSecret(AwsAccessKeyId),
         InjectSecret(AwsSecretAccessKey),
     )
+
+BEDROCK_INFERENCE_PROFILES = ["premier"]
+for model in BEDROCK_INFERENCE_PROFILES:
+    SUTS.register(
+        AmazonNovaSut,
+        f"amazon-nova-1.0-{model}",
+        f"us.amazon.nova-{model}-v1:0",
+        InjectSecret(AwsAccessKeyId),
+        InjectSecret(AwsSecretAccessKey),
+    )


plugins/amazon/modelgauge/suts/aws_bedrock_client.py
```diff
@@ -164,3 +164,13 @@ def translate_response(self, request: BedrockRequest, response: BedrockResponse)
...
     )
+
+BEDROCK_INFERENCE_PROFILES = ["premier"]
+for model in BEDROCK_INFERENCE_PROFILES:
+    SUTS.register(
+        AmazonNovaSut,
+        f"amazon-nova-1.0-{model}",
+        f"us.amazon.nova-{model}-v1:0",
+        InjectSecret(AwsAccessKeyId),
+        InjectSecret(AwsSecretAccessKey),
```
superdosh, line 175: Given the earlier discussion, do we want to allow the user to configure the region explicitly via their config? (I suppose they could configure it via their aws profile regardless.)

dhosterman: I think that might be a good idea in the future, but I also think we're the only real user for the foreseeable future, so I don't expect it matters right now.

superdosh: Am I correct on the resolution? We'd just set AWS_REGION in whatever env we're running this to match what we need?

dhosterman: I just tested that and it doesn't seem to work. The region is hard-coded into the SUT, so either that is taking precedence over the AWS_REGION envvar or the AWS_REGION envvar isn't being read by boto.

If we wanted to make it default to `us-east-1` and allow an optional envvar to override it, we could easily do that, or we could put it into the `config/secrets.toml` file alongside the keys. But it isn't something that is just available for free right now.

I'm open to whatever! I'd like any pattern we adopt for customizing individual SUTs to be consistent and I'm not sure what that looks like. Let's do a quick poll during standup.

superdosh: I didn't realize it was hard-coded. I don't see where? The case I was thinking about was if e.g. I had my env with AWS_REGION set to europe-blahblahblah for whatever reason, would it work (and if not, could I fix it easily). But it sounds like it doesn't use the env anyways, haha, so it might be a not relevant point.

dhosterman: Ah!

It's up on line 117:

```python
def _load_client(self):
        return boto3.client(
            service_name="bedrock-runtime",
            region_name="us-east-1",
            aws_access_key_id=self.access_key_id,
            aws_secret_access_key=self.secret_access_key,
        )
```

So I explicitly tried it with `AWS_REGION=us-west-2 poetry run benchmark ...` to see if it would read the envvar and override the above and it does not.

wpietri (approved):

plugins/amazon/modelgauge/suts/aws_bedrock_client.py
superdosh: Thank you! Makes sense with how we have it set up.

PR #1076, opened 2025-06-10
Fix Pydantic issue with cached values.

wpietri:
The Llama API is returning things that are invalid according to their Pydantic definitions, but this somehow doesn't show up as a problem until cache retrieval. Here I'm just fixing the data to match its typing. Maybe there's a better option?

### DIFF
--- a/src/modelgauge/suts/meta_llama_client.py
+++ b/src/modelgauge/suts/meta_llama_client.py
@@ -105,6 +105,9 @@ class MetaLlamaModeratedSUT(PromptResponseSUT[MetaLlamaChatRequest, MetaLlamaMod
         messages: list = kwargs.get("messages")  # type: ignore
         messages.append(chat_response.completion_message)
         moderation_response = self.client.moderations.create(messages=messages)
+        for r in moderation_response.results:
+            if r.flagged_categories is None:
+                r.flagged_categories = []  # make objects comply with Pydantic definitions
         return MetaLlamaModeratedResponse(sut_response=chat_response, moderation_response=moderation_response)
 
     def translate_response(self, request: MetaLlamaChatRequest, response: MetaLlamaModeratedResponse) -> SUTResponse:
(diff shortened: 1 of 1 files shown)


wpietri:
Per Roger, the error:
````
there's an issue with a meta llama SUT that just went in

E           pydantic_core._pydantic_core.ValidationError: 1 validation error for MetaLlamaModeratedResponse
E           moderation_response.results.0.flagged_categories
E             Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]



FAILED plugins/validation_tests/test_object_creation.py::test_can_cache_all_sut_responses[meta-llama-4-scout-17b-16e-instruct-fp8-moderated-llama] - pydantic_core._pydantic_core.ValidationError: 1 validation error for MetaLlamaModeratedResponse
moderation_response.results.0.flagged_categories
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]

```

src/modelgauge/suts/meta_llama_client.py
```diff
@@ -105,6 +105,9 @@ def evaluate(self, request: MetaLlamaChatRequest) -> MetaLlamaModeratedResponse:
         messages: list = kwargs.get("messages")  # type: ignore
         messages.append(chat_response.completion_message)
         moderation_response = self.client.moderations.create(messages=messages)
+        for r in moderation_response.results:
```
superdosh, line 108: Kind of sucks we have to deal with this. Maybe we also create an issue (https://github.com/meta-llama/llama-api-python) and add a link to the issue here, in case we want to clean this up later?

superdosh (approved):

src/modelgauge/suts/meta_llama_client.py
wpietri: Good idea.

wpietri: And done: https://github.com/meta-llama/llama-api-python/issues/33

PR #1087, opened 2025-06-13
Bug fixes

rogthefrog:
* fix lowercasing of SUT UIDs
* fix bad function call

This will fix the [recent smoke test failures](https://github.com/mlcommons/modelbench-private/actions/runs/15634118184/job/44045009804).


dhosterman (approved):
👍

PR #1111, opened 2025-07-01
try both serverless and dedicated until you find a place to run your model

rogthefrog:
When trying to run dynamic SUTs directly on huggingface, the code only looked for a dedicated endpoint. Now it looks for a serverless option as well.


bkorycki (approved):
👍

PR #1113, opened 2025-07-02
Standardize columns + Data refactor

bkorycki:
The overall goal of this PR is to standardize data formatting so that the outputs of one stage can be used as the input of another. And to avoid head aches.
Two new objects are instrumental in accomplishing this:
1) The Data Schema objects, which define column names in one place.
2) Dataset objects which handle both reading from input and writing to output.
    - The different types of datasets will read CSV rows and output the relevant object for each row e.g. TestItem for prompts, SUTInteraction for prompts-responses, and AnnotatedSUTInteraction for annotations.
    - The Annotation dataset serializes/deserializes annotations as json-strings. It can serialize pydantic objects or dictionaries. Annotations will always be deserialized as dictionaries.

Also in service of this goal, every output file (aside from metadata.json) is a CSV file in which one row corresponds to one prompt and one sut response and/or one annotation. No more `annotations.jsonl.`

### DIFF
--- a/src/modelgauge/dataset.py
+++ b/src/modelgauge/dataset.py
@@ -203,8 +203,6 @@ class AnnotationDataset(BaseDataset):
         )
         response = SUTResponse(text=row[self.schema.sut_response])
         interaction = SUTInteraction(prompt, row[self.schema.sut_uid], response)
-        print(row[self.schema.annotation])
-        print(type(row[self.schema.annotation]))
         annotation = json.loads(row[self.schema.annotation])
         return AnnotatedSUTInteraction(
             sut_interaction=interaction, annotator_uid=row[self.schema.annotator_uid], annotation=annotation
--- a/src/modelgauge/dataset.py
+++ b/src/modelgauge/dataset.py
@@ -1,0 +1,212 @@
+import csv
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import Dict, Iterable, Optional, Union, Any, Sequence
+
+from modelgauge.data_schema import (
+    DEFAULT_ANNOTATION_SCHEMA,
+    DEFAULT_PROMPT_RESPONSE_SCHEMA,
+    DEFAULT_PROMPT_SCHEMA,
+    AnnotationSchema,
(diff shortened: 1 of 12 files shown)


src/modelgauge/dataset.py
```diff
@@ -0,0 +1,229 @@
...
+from modelgauge.single_turn_prompt_response import AnnotatedSUTInteraction, SUTInteraction, TestItem
+from modelgauge.sut import SUTResponse
+
+
+class BaseDataset(ABC):
+    """This class provides common functionality for CSV file handling and context management."""
+
+    quoting = csv.QUOTE_ALL
+
+    def __init__(self, path: Union[str, Path], mode: str):
```
rogthefrog, line 27: Would it be useful to have a BaseDatasetReader and BaseDatasetWriter instead of handling both use cases in one class?

src/modelgauge/dataset.py
```diff
@@ -0,0 +1,229 @@
...
+    def row_to_item(self, row: dict) -> AnnotatedSUTInteraction:
+        prompt = TestItem(
+            prompt=TextPrompt(text=row[self.schema.prompt_text]),
+            source_id=row[self.schema.prompt_uid],
+            context=row,
+        )
+        response = SUTResponse(text=row[self.schema.sut_response])
+        interaction = SUTInteraction(prompt, row[self.schema.sut_uid], response)
+        print(row[self.schema.annotation])
+        print(type(row[self.schema.annotation]))
```
rogthefrog: Are these `print`s intentional?

rogthefrog (approved):
This is great and very clear.

src/modelgauge/dataset.py
bkorycki: Hm that's interesting. But I don't like the idea of having 6 dataset classes instead of just 3. Do you think it would be worth it to split it up?

src/modelgauge/dataset.py
bkorycki: oops they're not! thanks

PR #1114, opened 2025-07-03
clarify provider vs driver nomenclature

rogthefrog:
and in so doing:

* improve exceptions
* improve documentation in comments
* add logic to try to parse an old SUT UID as if it were a dynamic SUT

### DIFF
--- a/plugins/huggingface/modelgauge/suts/huggingface_sut_factory.py
+++ b/plugins/huggingface/modelgauge/suts/huggingface_sut_factory.py
@@ -59,36 +59,15 @@ class HuggingFaceSUTFactory(DynamicSUTFactory):
 class HuggingFaceChatCompletionServerlessSUTFactory(HuggingFaceSUTFactory):
 
     @staticmethod
-    def find(
-        sut_metadata: DynamicSUTMetadata, requested_provider: str = "", find_alternative: bool = False
-    ) -> str | None:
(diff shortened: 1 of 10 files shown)


plugins/huggingface/modelgauge/suts/huggingface_sut_factory.py
```diff
@@ -60,30 +60,34 @@ def get_secrets() -> InjectSecret:
 class HuggingFaceChatCompletionServerlessSUTFactory(HuggingFaceSUTFactory):
 
     @staticmethod
-    def find(model_name, provider: str = "", find_alternative: bool = False) -> str | None:
-        sut_metadata: DynamicSUTMetadata = DynamicSUTMetadata.parse_sut_uid(model_name)
-        if not provider:
-            provider = sut_metadata.provider
+    def find(
+        sut_metadata: DynamicSUTMetadata, requested_provider: str = "", find_alternative: bool = False
```
bkorycki: When does the requested_provider param get used? Wouldn't the provider always be included in sut_metadata?

rogthefrog: That was old code from a previous attempt to add support for looking up alternate providers if the one in the SUT UID isn't available. I took all that out 🔥

superdosh (approved):
LGTM!

PR #1121, opened 2025-07-08
Update consistency checker to approve of our fix for blank responses.

wpietri:
Previously, blank responses would be scored by the evaluator. Now they are always safe. The production code was updated a while ago; this should bring the consistency checker up to date.

### DIFF
--- a/src/modelbench/consistency_checker.py
+++ b/src/modelbench/consistency_checker.py
@@ -240,6 +240,7 @@ class MinValidAnnotatorItems(JournalCheck):
 
 class AnnotationsMergedCorrectly(JournalCheck):
     """Checks that the majority vote (or false if <2 valid annotations) among the annotations is returned by measure_item_quality.
+    Bypasses the check if the prompt is blank, in which case benchmarks ignore the annotators and always mark it safe.
     NOTE: This only will only work on tests who use a majority vote merging strategy and measure 'is_safe'."""
 
     def __init__(self, search_engine: JournalSearch, sut, test):
--- a/src/modelbench/consistency_checker.py
+++ b/src/modelbench/consistency_checker.py
@@ -1,15 +1,15 @@
-import casefy
 import json
 import shutil
 from abc import ABC, abstractmethod
 from collections import Counter, defaultdict
 from itertools import product
-from rich.console import Console
-from rich.table import Table
 from typing import Dict, List
 
+import casefy
+from rich.console import Console
+from rich.table import Table
+
 from modelbench.run_journal import journal_reader
-from modelbench.scoring import score_to_ordinal_grade
 from modelgauge.config import load_secrets_from_config
 from modelgauge.test_registry import TESTS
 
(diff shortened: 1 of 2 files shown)


src/modelbench/consistency_checker.py
```diff
@@ -269,12 +272,21 @@ def __init__(self, search_engine: JournalSearch, sut, test):
...
         vote_counts = Counter(self.prompt_annotations[prompt])
         assert set(vote_counts.keys()).issubset(
             {True, False}
         ), "Incompatible annotation values found. Expected only True/False."
 
         measurement = self.prompt_measurements[prompt]
+
+        if self.response_by_id.get(prompt) == "":
+            if measurement == 1.0:
+                # blank responses are always safe
```
bkorycki, line 286: Can you update the documentation for this class (`AnnotationsMergedCorrectly`) to also describe this check?:)

bkorycki (commented):
Your changes look good to me. The only thing I'm confused about is that I would expect more than just `AnnotationsMergedCorrectly` to be failing with our blank response changes.  We don't send the item for evaluation at all if it's an empty string right? If so, I would think that would also cause failures in `EachResponseAnnotatedOnce` and maybe even `EachItemMeasuredOnce`. Is that not the case?

wpietri:
> Your changes look good to me. The only thing I'm confused about is that I would expect more than just `AnnotationsMergedCorrectly` to be failing with our blank response changes. We don't send the item for evaluation at all if it's an empty string right? If so, I would think that would also cause failures in `EachResponseAnnotatedOnce` and maybe even `EachItemMeasuredOnce`. Is that not the case?

You can take a look at `airr-data-confidential/runs/benchmarks/20250611-192628-benchmark-meta-llama-4-scout-17b-16e-instruct-fp8-moderated-llama-ensemble` if you'd like an example. This was the only check failing there. If you look at [the code](https://github.com/mlcommons/modelbench/blob/692ba112d871826cf7b99feeddedc2bad57324ad/src/modelbench/benchmark_runner.py#L384-L383) you'll see that it overrides the initial measurement after it is complete. I think we could change it to the way you describe, but this way logs what the bypassed votes were, which I think will be helpful in knowing when we've fixed this behavior in the evaluators.

src/modelbench/consistency_checker.py
wpietri: Done, but who's the documentation for?

My general theory is that the code is the last place we should put documentation, as it's duplicative and (as this case demonstrates) can easily get out of sync. So I'd rather put important information in class names, method names, automated test names, automated test content, and other structural elements in both prod and test code.

If this documentation is just for other programmers and you'd like to make things clearer to them, I'd rather we improve the tests to be clearer. There's a fair bit of opportunity here, and I try to think of tests as executable documentation, something where the computer can verify that things are still in sync.

bkorycki: For myself :) I like to be able to reference an explicit class comment over having to dig through tests to figure out whats going on. And I think the class names can only provide so much detail e.g. `AnnotationsMergedCorrectly`. Normally I would agree about concerns of staying in sync, but I think these objects are discrete and isolated enough from the rest of the code base that it should be easy to keep the documentation up to date.

wpietri: This is definitely something we should talk about as a group. I think in this case if the tests and code are too hard to understand, there's a lot we can do to make it clearer. Comments like this are essentially duplicative and unreliable, and so I think they should be the last place we put important information.

PR #1122, opened 2025-07-09
Search up tree for config

bkorycki:
This searches *all the way* to the root.

### DIFF
--- a/src/modelgauge/config.py
+++ b/src/modelgauge/config.py
@@ -12,8 +12,9 @@ SECRETS_PATH = os.path.join(DEFAULT_CONFIG_DIR, DEFAULT_SECRETS)
 CONFIG_TEMPLATES = [DEFAULT_SECRETS]
 
 
-def write_default_config(dir: str = DEFAULT_CONFIG_DIR):
+def write_default_config(parent_dir: str = "."):
     """If the config directory doesn't exist, fill it with defaults."""
+    dir = os.path.join(parent_dir, DEFAULT_CONFIG_DIR)
     if os.path.exists(dir):
         # Assume if it exists we don't need to add templates
         return
--- a/src/modelgauge/config.py
+++ b/src/modelgauge/config.py
@@ -13,7 +13,7 @@ CONFIG_TEMPLATES = [DEFAULT_SECRETS]
 
 
 def find_config_dir(path: str = ".") -> str:
-    """ "Search up the tree for the config directory."""
+    """Search up the tree for the config directory."""
     current_dir = os.path.abspath(path)
     while True:
         config_dir = os.path.join(current_dir, DEFAULT_CONFIG_DIR)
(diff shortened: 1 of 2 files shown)


src/modelgauge/config.py
```diff
@@ -12,21 +12,40 @@
 CONFIG_TEMPLATES = [DEFAULT_SECRETS]
 
 
-def write_default_config(dir: str = DEFAULT_CONFIG_DIR):
+def find_config_dir(path: str = ".") -> str:
+    """Search up the tree for the config directory."""
+    current_dir = os.path.abspath(path)
+    while True:
+        config_dir = os.path.join(current_dir, DEFAULT_CONFIG_DIR)
```
superdosh, line 19: Correct me if I'm wrong, but it looks like the code now assumes the directory will always be called `config` and that is no longer configurable?

src/modelgauge/config.py
```diff
@@ -12,21 +12,40 @@
...
+        dir = os.path.join(parent_dir, DEFAULT_CONFIG_DIR)
+        os.makedirs(dir)
+        for template in CONFIG_TEMPLATES:
+            source_file = str(resources.files(config_templates) / template)
+            output_file = os.path.join(dir, template)
+            shutil.copyfile(source_file, output_file)
 
 
-def load_secrets_from_config(path: str = SECRETS_PATH) -> RawSecrets:
+def load_secrets_from_config(path: str = ".") -> RawSecrets:
```
superdosh, line 45: The way we use this in `modelplane`, we pass the full path (e.g. `flightpaths/config/secrets.toml`) to this function. Will that still work? I'm not sure above whether `find_config_dir` will strip the file name?

src/modelgauge/config.py
bkorycki: Yes thats correct. But I don't believe that configuration was actually used anywhere outside of the unit tests. So I don't think we are losing out on anything.

rogthefrog: Have you considered using `glob("**/secrets.toml", recursive=True)` and walking that list? It may simplify the directory walking logic. You'd have to specify the repo's root directory to start from.

rogthefrog (changes requested):
It may be useful to explore the `glob` method.

src/modelgauge/config.py
bkorycki: Yes that should work! I just added another test for that.

src/modelgauge/config.py
wpietri: For what it's worth, the main use case for me is being in some sub-directory, running a command, and having it generate a sample secrets file that a) I don't want, b) is confusing, and c) I have to clean up.

For me, walking up to the repo root would be fine. Walking down might not hurt, although I'd want to be sure there are no sample files or alternative configuration files that are unhelpful. Right now I definitely have junk auto-created secrets files that I'd have to remove to be sure that would work.

Also, although we only have a secrets file, we have talked about adding config files. For example, we're kinda abusing the secrets file to point at the correct cheval instance. And I think there were other things discussed. I'd like to not lose the concept of a config directory altogether.

bkorycki: I think the issue with walking down is that you have to know what the root directory is. We can't assume the root is `modelbench` because that wouldn't find the config if youre in modelbench-private.

Issue #1102, opened 2025-06-26
One SUT per run in modelbench


bkorycki:
Remove SUT field from all journal records except for one SUT info record at the beginning.

PR #1123, opened 2025-07-10
Remove support for multiple SUTs from modelgauge and modelbench

rogthefrog:
https://github.com/mlcommons/modelbench/issues/1102


src/modelgauge/main.py
```diff
@@ -312,16 +313,18 @@ def run_job(
     else:
         ensemble = None
 
+    sut_uids = listify(sut_uid)
```
bkorycki: why not just `list`?

src/modelgauge/main.py
```diff
@@ -423,7 +426,7 @@ def show_progress(data):
     type=click.Path(exists=True, path_type=pathlib.Path),
 )
 def run_csv_items(
```
bkorycki: Honestly I think we can just delete `run_csv_items` at this point. The main difference between it and `run-job` was multi-sut support.

src/modelgauge/secret_values.py
```diff
@@ -2,10 +2,10 @@
 from dataclasses import dataclass
 from typing import Generic, List, Mapping, Optional, Sequence, Type, TypeVar
 
-from pydantic import BaseModel
-
 from modelgauge.general import get_concrete_subclasses
 
+from pydantic import BaseModel
```
bkorycki: I think your linter is configured differently 🤔

src/modelbench/run.py
```diff
@@ -150,7 +158,11 @@ def benchmark(
         ]
 
     # SUT UIDs are validated in the callback function, so we don't need to validate here
-    suts = make_suts(sut_uids)
+    suts = make_suts(
+        [
+            sut_uid,
```
bkorycki: This formatting is kind of cray cray no?

bkorycki (commented):
I think you hit everything you needed to for this! You might also want to update the journal docs to remove the "sut" field as being something that's listed in every entry.

src/modelbench/run.py
```diff
@@ -94,8 +94,16 @@ def list_suts():
...
 @click.option("--debug", default=False, is_flag=True)
 @click.option("--json-logs", default=False, is_flag=True, help="Print only machine-readable progress reports")
-@click.option("sut_uids", "--sut", "-s", multiple=True, help="SUT uid(s) to run", required=True, callback=validate_uid)
-@click.option("--anonymize", type=int, help="Random number seed for consistent anonymization of SUTs")
+@click.option(
+    "sut_uid",
+    "--sut",
+    "-s",
+    multiple=False,
+    help="SUT UID to run. Multiple SUT UIDs are no longer supported",
```
wpietri: Given that nobody but me ever used this, I don't think you need to add an explanation.

src/modelbench/run.py
wpietri: Agreed, and I wonder if make_suts could just become make_sut.

src/modelbench/run.py
```diff
@@ -255,17 +268,17 @@ def score_benchmarks(run):
...
     runner.benchmarks = benchmarks
-    runner.suts = suts
+    runner.sut = sut
     runner.max_items = max_instances
     runner.debug = debug
     runner.thread_count = thread_count
     runner.run_tracker = JsonRunTracker() if json_logs else TqdmRunTracker(0.5)
 
-    print(f"Starting run for {[b.uid for b in benchmarks]} over {[s.uid for s in suts]}")
+    print(f"Starting run for {[b.uid for b in benchmarks]} over {sut.uid}")
```
wpietri: It's the smallest of things, but I think "over" should become "for".

src/modelbench/run_journal.py
```diff
@@ -5,14 +5,15 @@
 from datetime import datetime, timezone
 from enum import Enum
 from io import IOBase, TextIOWrapper
-from typing import Sequence, Mapping
+from typing import Mapping, Sequence
 from unittest.mock import MagicMock
 
+from modelgauge.sut import SUTResponse
```
wpietri: Why is the modelgauge import moved here? Previously the sections were core libs, imported libs, and project code. Now I don't know what they are.

src/modelbench/run_journal.py
```diff
@@ -112,8 +113,12 @@ def _item_fields(self, item):
         except Exception as e:
             # another good place for logging
             entry["prompt_id"] = f"failed: {e}"
-        if item.sut:
-            entry["sut"] = item.sut.uid
+
+        # we only support a single SUT per run, so the sut field is unnecessary
+        # https://github.com/mlcommons/modelbench/issues/1102
```
wpietri: I don't think we can just change the journal format like this. There are a number of tools that use it, and those tools need to work for old journals and new. We also need to accommodate people doing run comparisons, which is harder when the file formats change.

At a minimum, we should add a version number to the journal and then make tools work with either version. But we may want to lump together a set of journal changes at the same time, so as to minimize the number of versions to manage.

As an aside, I don't think comments like this should exist. We have a version control system that contains both the history and an explanation of changes; historical information should generally live there. As Longfellow wrote:

```
    Trust no Future, howe’er pleasant!
    Let the dead Past bury its dead!  
    Act,— act in the living Present! 
```

If it was more a comment for current reviewers, I'd say put it in the PR description.

src/modelgauge/command_line.py
```diff
@@ -45,12 +45,21 @@ def load_local_plugins(_, __, path: pathlib.Path):
...
     valid_uids = sorted(registry.keys(), key=lambda x: x.lower())
     valid_uids_str = "\n\t".join(valid_uids)
     return "\t" + valid_uids_str
 
 
+def listify(value):
+    """Some functions accept a single UID or a list of them."""
+    if isinstance(value, str):
+        return [
+            value,
```
wpietri: That's the second time I've seen this extra spacing, so it makes me wonder if it's on purpose, and if so what the intent is.

src/modelgauge/command_line.py
```diff
@@ -113,25 +122,24 @@ def _validate_sut_uid(ctx, param, value):
...
-    for sut_uid in requested_sut_ids["dynamic"]:
+    # modelgauge and modelbench runs now only support one SUT at a time,
+    # and that is enforced in the input validation. This function can still
+    # handle multiple SUT UIDs at a time to match validate_uid's semantics,
+    # and because it doesn't know or care about the logic in the caller.
+    sut_uids = listify(value)
+    requested_sut_uids = classify_sut_uids(sut_uids)
+    valid_sut_uids = requested_sut_uids["known"]
+    for sut_uid in requested_sut_uids["dynamic"]:
         dynamic_sut = make_dynamic_sut_for(sut_uid)  # a tuple that can be splatted for SUTS.register
```
wpietri: it seems weird to me that a method supposedly about validation does this work, but I guess we can get rid of this when we remove registration

src/modelgauge/command_line.py
```diff
@@ -113,25 +122,24 @@ def _validate_sut_uid(ctx, param, value):
...
-    if isinstance(value, str):
-        sut_uids = [value]
-    else:
-        sut_uids = value
-
-    requested_sut_ids = classify_sut_ids(sut_uids)
-
-    valid_sut_uids = requested_sut_ids["known"]
-    for sut_uid in requested_sut_ids["dynamic"]:
+    # modelgauge and modelbench runs now only support one SUT at a time,
```
wpietri: This seems awfully elaborate to me given the actual need, which is to check one SUT uid. If the validate_uid function is no longer the right shape, maybe the sut case should just point here, and that function can be used for tests and annotators.

src/modelgauge/command_line.py
```diff
@@ -202,10 +208,11 @@ def check_secrets(secrets, sut_uids=None, test_uids=None, annotator_uids=None):
     return True
 
 
-def classify_sut_ids(uids):
+def classify_sut_uids(uids):
```
wpietri: This also now seems overly elaborate. Isn't this just used here for a single SUT uid? In which case I think this can get merged with the caller and just do the thing rather than passing around complicated classification structure.

src/modelbench/benchmark_runner.py
```diff
@@ -29,6 +22,14 @@
 from modelgauge.single_turn_prompt_response import TestItem
 from modelgauge.sut import PromptResponseSUT, SUTOptions, SUTResponse
 
+from pydantic import BaseModel
+from tqdm import tqdm
+
+from modelbench.benchmark_runner_items import ModelgaugeTestWrapper, TestRunItem, Timer
```
wpietri: Same question about the organizing principle of this change. Seems like we have dueling autoformatters, and we should make them compatible.

src/modelbench/benchmark_runner.py
```diff
@@ -468,50 +465,48 @@ def __init__(self, data_dir: pathlib.Path):
         self.debug = False
         self.data_dir = data_dir
         self.secrets = None
-        self.suts = []
+        self.sut = None
         self.max_items = 10
         self.thread_count = 1
         self.run_tracker = NullRunTracker()
 
     def add_sut(self, sut: PromptResponseSUT):
```
wpietri: This method is not used in production code, so I think it can just be removed.

src/modelbench/benchmark_runner.py
```diff
@@ -468,50 +465,48 @@ def __init__(self, data_dir: pathlib.Path):
...
+            else:
+                raise RuntimeError(f"initial check failed with no text: {raw_response}")
+        except Exception as e:
+            logger.error(f"initial check failure for {self.sut.uid}", exc_info=e)
+            print(f"initial check failure for {self.sut.uid}")
+            traceback.print_exc()
+            working = False
 
-                return False
+        if not working:
```
wpietri: What's the value of this extra code block?

tests/modelbench_tests/test_run_journal.py
```diff
@@ -226,15 +228,16 @@ def test_item_exception_entry(self, journal, sut):
...
         assert e["test"] == "a_test"
         assert e["prompt_id"] == "id1"
-        assert e["sut"] == tri.sut.uid
         assert e["exception"]["class"] == "ValueError"
+        assert "sut" not in e  # we no longer output the sut at the item level
 
     def make_test_run_item(self, source_id, test_id, text):
-        from modelbench.benchmark_runner import TestRunItem, ModelgaugeTestWrapper
-        from modelbench_tests.test_benchmark_runner import AFakeTest
+        from modelbench.benchmark_runner import ModelgaugeTestWrapper, TestRunItem
```
wpietri: Could you please set your autoformatter to not change lines you didn't edit and/or manually reformat?

tests/modelbench_tests/test_run_journal.py
```diff
@@ -205,7 +205,9 @@ def test_run_item_output_with_sut(self, journal, sut):
         journal.item_entry("an item", tri)
 
         e = journal.last_entry()
-        assert e["sut"] == tri.sut.uid
+        # we are no longer outputting the sut in item blocks, because it's redundant
+        # now that we only include one sut per run
+        assert "sut" not in e
```
wpietri: Generally I don't think we need tests that cover things that shouldn't be there. Although as mentioned elsewhere, I think this probably should still be there.

wpietri (changes requested):
Making good progress, but as mentioned in the comments, I think you can remove a bit more of the multi-sut machinery. We should at least discuss changing the journal format, and we may want to batch this change with others. And as mentioned there's definitely something going on with whatever you're using to handle imports. We should all make sure we're using the same approach.

Oh, and excellent branch name.

src/modelbench/run.py
```diff
@@ -267,17 +263,17 @@ def score_benchmarks(run):
     return benchmark_scores
 
 
-def run_benchmarks_for_suts(benchmarks, suts, max_instances, debug=False, json_logs=False, thread_count=32):
+def run_benchmarks_for_suts(benchmarks, sut, max_instances, debug=False, json_logs=False, thread_count=32):
```
bkorycki: should this be `run_benchmarks_for_sut` now?

PR #1125, opened 2025-07-14
Move HF to core + remove find from dynamic factory interface

bkorycki:
1) Move hf to core because the hf dynamic sut factory code makes everything break if you don't have hf installed.
2) Move sut tests (ie tests that were in the plugins before) to a new `sut_tests` sub dir. This is mainly to preserve some organization as we move stuff out of plugins and into core.
3) Remove `find` from the dynamic sut factory interface. It is not used externally at all. And there is no contract for what the method returns e.g. sometimes it returns a model name, sometimes a provider.

### DIFF
--- a/src/modelgauge/suts/huggingface_api.py
+++ b/src/modelgauge/suts/huggingface_api.py
@@ -2,7 +2,6 @@ from typing import Optional
 
 import requests  # type: ignore
 import tenacity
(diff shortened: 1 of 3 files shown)


wpietri (approved):
Great catch, and I like the solution.

PR #1131, opened 2025-07-15
Dynamic OpenAI SUTs

bkorycki:
You can use the following sample UIDs to get dynamic openai suts:
- gpt-4o:openai
- openai/gpt-4o:openai


src/modelgauge/suts/openai_sut_factory.py
```diff
@@ -0,0 +1,40 @@
...
+from modelgauge.secret_values import InjectSecret
+from modelgauge.suts.openai_client import OpenAIApiKey, OpenAIChat, OpenAIOrgId
+
+
+DRIVER_NAME = "openai"
+
+
+class OpenAISUTFactory(DynamicSUTFactory):
+    @staticmethod
+    def _model_exists(sut_metadata: DynamicSUTMetadata):
```
rogthefrog, line 15: This is kind of what I had in mind for the `find` method in the superclass. Regardless of the name, I feel it'd make the paradigm clear if we did have a method like this in the superclass.

bkorycki: It feels more like a class-specific helper function to me. But I can rename it to _find if you would like

wpietri:
As I take a look at this new code, I'm ending up with some concerns about the old code. The class hierarchy for DynamicSUTFactory seems anemic to me. Wouldn't we be better off with instances that hold state? At the very least I think they should get passed the secrets on construction and keep them, and they should probably hold on to their respective clients. Then the calls that are effectively to `self` would look more reasonable as well. Then DYNAMIC_SUT_FACTORIES could be made up of those instances.

It's also concerning to me that a class that says it's a SUT factory doesn't actually return a SUT from the `make_sut` method. This seems like a fine opportunity to start encapsulating or replacing the SUT registry stuff, which I would love to get rid of. 

What do you think, @bkorycki and @rogthefrog?

bkorycki:
@wpietri I had similar thoughts while working on this. Calling `load_secrets_from_config` within the factory feels at odds with the way we approach secrets in the rest of the code base. And I also don't like that `make_sut` doesn't actually make a SUT.

I think a lot of these decisions were made because the dynamic sut stuff works by inserting a dynamic sut into the registry at the start of execution. Which is clever. But as we are moving to a more dynamic-focused world, I think the dynamic sut mechanics should be more foundational in how we create SUTs. Rather than forcing dynamic shaped SUTs into static shaped holes.

One idea I have is to merge the registry object together with the dynamic SUT code. So, `SUTS.make_instance(sut_uid, secrets=secrets)` first checks pre-registered SUTs, and then attempts to call the dynamic factories. 
This way, the code still uses the same interface to create SUTs (ie make_instance). But the dynamic factories/finders don't have to actually use or know about the registry. They just return SUT instances.

rogthefrog:
> As I take a look at this new code, I'm ending up with some concerns about the old code. The class hierarchy for DynamicSUTFactory seems anemic to me. Wouldn't we be better off with instances that hold state? 

I think they should *return* hydrated instances with secrets, etc. (see below for the note about bypassing the registry).

> It's also concerning to me that a class that says it's a SUT factory doesn't actually return a SUT from the `make_sut` method. This seems like a fine opportunity to start encapsulating or replacing the SUT registry stuff, which I would love to get rid of.

Yeah, the nearly random / flexible list of arguments that make up a registrable SUT passed to the registry are something I'd like to get rid of. The `make_sut` methods are really `compile_all_the_info_the_registry_needs` but that's a mouthful. Encapsulating that into SUT classes with defined interfaces would be cleaner. E.g. instead of the wide variety of calls to SUTS.register like

```
SUTS.register(
    AmazonNovaSut,
    f"amazon-nova-1.0-{model}",
    f"us.amazon.nova-{model}-v1:0",
    InjectSecret(AwsAccessKeyId),
    InjectSecret(AwsSecretAccessKey),
)

# or
SUTS.register(DemoYesNoSUT, "demo_yes_no")

# or
SUTS.register(
    AzureChatSUT,
    "phi-3.5-mini-instruct",
    "https://Phi-3-5-mini-instruct-hfkpb.eastus2.models.ai.azure.com",
    InjectSecret(PhiMiniKey),
)
```

it may be worth putting all those arguments e.g. in the constructor for the xxxSUT classes instead and bypass the registry. The dynamic SUT factories' `make_sut` methods could then do that, instead of returning an arbitrary tuple of arguments of various types in a particular order for use by the registry.

src/modelgauge/suts/openai_sut_factory.py
rogthefrog: Re-reading the HF and Together dynamic SUT factories, and this openAI one, I do see a unifying structure inside the factory: there's 

* a secrets injection/lookup part 
* a finder to interrogate the service provider about the model you want
* a maker to return that SUT.

So whatever the name is, I now do think all three methods (finder, secrets manager, and maker) should be in the superclass and implemented in the factory subclasses.

wpietri:
Ok, sounds like we are agreed: The existing approach makes sense as a way to preserve the registry system, but now that we've decided to move beyond it, the factories can become instances that are actual SUT factories, and `make_sut` can  return actual SUTs. Hooray!

src/modelgauge/suts/openai_sut_factory.py
bkorycki: I see abstract base classes as interfaces that guarantee that all concrete subclasses provide some set of functionality. I think the only functionality that is guaranteed by the factories is `make_sut`. The other 2 methods are only used by the subclasses themselves. And they vary in what they return. So there's no contract, and they're not a part of the interface. So I don't think it makes sense to put in the superclass.

PR #1141, opened 2025-07-21
Dynamic SUTs refactor

bkorycki:
This is an attempt to refactor the dynamic SUTs to address the concerns mentioned in PR #1131 , summarized below:
- Make dynamic SUT factory objects less anemic. Store them as instances.
- De-couple the dynamic SUT mechanism from the registry. Don't pre-register dynamic SUTs at the start of execution. 
- A dynamic SUT factory should return a hydrated SUT from `make_sut`, instead of a registry-tuple. (Related to above).

This is what I tried to do:
- Create a new object `SUTFactory` that is responsibly for making all SUTs. It encapsulates the dynamic SUT functionality.
- My goal was to be able to call `SUTFactory.make_instance()` for dynamic suts and pre-registered SUTs. The caller doesn't have to know anything about what type of SUT it is. We can strip the registry functionality from the object once we get to that point.

Here's where I got stuck:
-  the `validate_uid` and `check_secrets` functions both work well with the registry system. To extend those pre-flight functions to work on dynamic SUTs, they would have to attempt to create the dynamic SUT. But if a SUT is successfully created, they might as well register that SUT. Otherwise it will just have to do the work again when `make_instance` is called.
- So, I feel like I failed to find a way to de-couple the registry system from the dynamic SUTs.

(Less important) I think there is room for further clean up with how we map driver names to dynamic SUT factories. ie I think the driver name can be a part of the dynamic SUT factory classes, and we can collect those objects automatically instead of populating a dictionary. But I think this is work that can be done after sorting out this PR.


wpietri:
Maybe I'm missing something here, but given the dynamic nature of things, can't we just try making the SUT and then blow up if it doesn't work?

For example, if I remove the `validate_uid` and `check_secrets` calls, this runs fine:

` modelgauge run-sut --sut openai/gpt-4o:openai --prompt "Why did the chicken cross the road?"`

If I do it with an invalid model name, it blows up:

```
% modelgauge run-sut --sut openai/gpt-4p:openai --prompt "Why did the chicken cross the road?"
[stack trace removed]
modelgauge.dynamic_sut_factory.ModelNotSupportedError: Model gpt-4p not found or not available on openai.
```
From a user perspective, the stack trace is annoying, but since it's a nicely specific exception, I think we can just catch that at the top and print out a nice error message.

If we were doing more than one SUT per run, this would be more of a problem; we'd want the command to fail fast. But since we're getting rid of multi-SUT runs anyhow, this should either work fine or blow up quickly, shouldn't it?

If that's true, I think you can just be a bit more radical in your trimming, ending up with less complexity and a good user experience.

wpietri:
> @wpietri Okay! I made it so that we do validation and secret checks for pre registered suts and not dynamic ones.

I'm not following. I was expecting to see `validate_uid` and `check_secrets` get removed from places like `main.py:run_sut` and only get invoked if somebody didn't ask for a dynamic SUT.

bkorycki:
@wpietri You're right, the call to `check_secrets` in `run_sut` is redudant. I'll remove it, didn't realize it was there.

`validate_uid` is a callback for the click `--sut` option. I could have done a check inside`validate_uid` to see if the sut is dynamic  before proceeding with validation. But I opted to keep that logic encapsulated in the `SUT_FACTORY` object. imho things that use suts shouldn't have to know whether suts are dynamic or not. 
The same applies for `check_secrets`. The factory decides whether or not it should check the secrets for a sut. Nobody has to know about pre-registered/dynamic SUT classification besides the factory.

bkorycki:
@wpietri What's wrong with validating UIDs at the start of execution? That's the only place we call that method, so I'm not sure what you mean by doing it twice.

wpietri:
@bkorycki: As I showed [before](https://github.com/mlcommons/modelbench/pull/1141#issuecomment-3099312400), you can remove the `validate_uid` check and things work fine. Good uids work, bad uids blow up in a clear way. (That we could make more user-friendly with a try statement to translate the exception.)  The method isn't getting called twice, but the same work is being done.

Pre-checking the SUT uids was mildly helpful before, because if somebody put in two bad `--sut` arguments we could explain that they were both wrong at once. But now I don't see a point to a separate validation step. We can just check the argument once as part of creating the SUT.

bkorycki:
@wpietri Okay, I think I did what you want. I removed the `validate_uid` callback for `--sut` options and prettyfied the new error message that now gets evoked when a sut isnt found.

src/modelgauge/command_line.py
```diff
@@ -125,7 +119,7 @@ def validate_uid(ctx, param, value):
 
     unknown_uids = []
     for uid in values:
-        if uid not in registry.keys():
+        if not registry.knows(uid):
```
rogthefrog, line 122: Nice.

src/modelgauge/sut_factory.py
```diff
@@ -0,0 +1,98 @@
...
+    pass
+
+
+class SUTType(Enum):
+    DYNAMIC = "dynamic"
+    KNOWN = "known"
+    UNKNOWN = "unknown"
+
+
+# TODO: Auto-collect. Maybe make "driver" keys a constant in each factory module.
```
rogthefrog, line 24: They are, at least in some. E.g. huggingface_sut_factory line 19. That was the intent originally.

src/modelgauge/sut_factory.py
```diff
@@ -0,0 +1,98 @@
...
+# Maps a string to the module and factory function in that module
+# that can be used to create a dynamic sut
+DYNAMIC_SUT_FACTORIES: dict = {
+    "proxied": {"hfrelay": HuggingFaceSUTFactory},
```
rogthefrog, line 28: I have a vague memory of chatting about flattening this and ditching the proxied vs direct distinction. Did you decide to keep it here?

src/modelgauge/sut_factory.py
```diff
@@ -0,0 +1,98 @@
...
+            return SUTType.KNOWN
+        elif ":" in uid:
+            return SUTType.DYNAMIC
+        else:
+            return SUTType.UNKNOWN
+
+    def _make_dynamic_sut(self, uid: str) -> SUT:
+        sut_metadata: DynamicSUTMetadata = DynamicSUTMetadata.parse_sut_uid(uid)
+
+        if sut_metadata.is_proxied():
```
rogthefrog, line 79: This would be simpler if we removed the proxied vs direct distinction.

src/modelgauge/suts/together_sut_factory.py
```diff
@@ -1,39 +1,36 @@
...
 from modelgauge.suts.together_client import TogetherChatSUT
 
 
 DRIVER_NAME = "together"
 
 
 class TogetherSUTFactory(DynamicSUTFactory):
+    def __init__(self, raw_secrets: RawSecrets):
+        super().__init__(raw_secrets)
+        self._client = None  # Lazy load.
```
rogthefrog, line 16: Did you want to add a type hint like for the OpenAI client above?

tests/modelbench_tests/test_run.py
```diff
@@ -273,7 +273,7 @@ def test_benchmark_bad_sut_errors_out(self, runner, tmp_path):
         benchmark_options.extend(["--locale", "en_us"])
         benchmark_options.extend(["--prompt-set", "practice"])
 
-        with pytest.raises(SUTNotFoundException):
+        with pytest.raises(ValueError, match="No registration for bogus"):
```
rogthefrog, line 276: What's the reason to test for a more generic exception? SUTNotFoundException and related exception classes are more expressive IMO.

tests/modelgauge_tests/sut_tests/test_openai_sut_factory.py
```diff
@@ -1,44 +1,58 @@
...
         "modelgauge.suts.openai_sut_factory.OpenAISUTFactory._model_exists",
         return_value=True,
     ):
         sut_metadata = DynamicSUTMetadata(model="gpt-4o", maker="openai", driver="openai")
-        factory = OpenAISUTFactory()
         sut = factory.make_sut(sut_metadata)
-    assert sut[0] == OpenAIChat
-    assert sut[1] == "openai/gpt-4o:openai"
-    assert sut[2] == "gpt-4o"
+    assert isinstance(sut, OpenAIChat)
```
rogthefrog, line 24: This makes me happy

tests/modelgauge_tests/test_cli.py
```diff
@@ -129,9 +128,8 @@ def test_run_test_demos(sut_uid, test):
 
 def test_run_test_invalid_sut_uid():
     TESTS.register(FakeTest, "fake-test")
-    with pytest.raises(SUTNotFoundException):
-        _ = run_cli("run-test", "--sut", "unknown-uid", "--test", "fake-test")
-    del TESTS._lookup["fake-test"]
+    with pytest.raises(ValueError, match="No registration for unknown-uid"):
```
rogthefrog, line 131: Same question about the specific exception classes.

tests/modelgauge_tests/test_sut_factory.py
```diff
@@ -0,0 +1,78 @@
...
+from modelgauge_tests.test_dynamic_sut_factory import FakeDynamicFactory
+
+KNOWN_UID = "known"
+UNKNOWN_UID = "pleasedontregisterasutwiththisuid"
+DYNAMIC_UID = "google:gemma:nebius:hfrelay"
+
+
+@pytest.fixture
+def sut_factory():
+    """Fixture to simulates the SUTs global without contaminating it."""
```
rogthefrog, line 18: This is both clever and gross :D Globals like those registries are sad-making.

tests/modelgauge_tests/test_sut_factory.py
```diff
@@ -0,0 +1,78 @@
...
+    registry = InstanceFactory[SUT]()
+    registry.register(SUT, KNOWN_UID)
+    factory = SUTFactory(registry)
+    return factory
+
+
+@pytest.fixture
+def sut_factory_dynamic():
+    """SUT factory that patches the dynamic SUT factories."""
```
rogthefrog, line 27: This is great!

rogthefrog (approved):
This is a great improvement. I have a couple of questions/suggestions in the review, but they can be addressed later (if ever) and shouldn't block merging.

wpietri (approved):
Huge progress. Thanks for doing this!

src/modelgauge/sut_factory.py
bkorycki: I decided that there was already too much in this PR haha. I still think it would be valuable to do though

src/modelgauge/sut_factory.py
bkorycki: Yeah, I think it would be nice if it were fully automatic though:)

src/modelgauge/sut_factory.py
bkorycki: Yes! This whole chunk would be unnecessary.

src/modelgauge/suts/together_sut_factory.py
bkorycki: I think I only added it for openai because mypy was complaining

tests/modelbench_tests/test_run.py
bkorycki: Yeah, it is more expressive. I only changed it to accommodate for the fact that SUTs dont get validated with `validate_uid` anymore. I think the failure actually happens in when check_secrets tries to get the thing from the registry, and the registry raises that error. I unfortunately can't change it to raise a `SUTNotFoundException` because the code is in the InstanceFactory class, which handles other registry types as well e.g. tests, annotators. I agree this is less ideal.

PR #1143, opened 2025-07-23
prototype longer, richer SUT UIDs

rogthefrog:
As discussed in Chicago, we are looking for SUT UIDs that include more details about the SUT configuration while remaining human readable, and having a deterministic shape so we can unambiguously turn a SUT config into a SUT UID and vice versa.

This is a first pass.

https://github.com/mlcommons/modelbench/issues/1097


wpietri:
Interesting! What sort of feedback do you want at this stage?

Issue #1145, opened 2025-07-24
Accept SUT definition as a config blob passed to modelgauge and modelbench


rogthefrog:
See "SUT UID Discussion" section in [this document](https://docs.google.com/document/d/175PjVefwQbnLMw9_brtwetArf_xwmk4lyLSLALM4_tE/edit?tab=t.0)

rogthefrog:
TODO: discuss whether we want to also accept individual parameters in the CLI.

PR #1149, opened 2025-07-25
PoC: rich SUT UIDs

rogthefrog:
# What's This?

A PoC of what I'm calling "rich SUT UIDs" i.e. a JSON and/or string representation of SUT definitions (model, driver, and SUT options).

# How to Test

```bash
# built-in demo SUT ID
modelgauge run-sut --sut demo_yes_no --prompt "why did the chicken cross the road?"

# real "legacy" SUT ID
modelgauge run-sut --sut gpt-4o --prompt "why did the chicken cross the road?"

# dynamic SUT on nebius via HF
modelgauge run-sut --sut google/gemma-3-27b-it:nebius:hfrelay --prompt "why did the chicken cross the road?"

# rich SUT UID (string representation)
modelgauge run-sut --sut "google/gemma-3-27b-it:nebius:hfrelay;mt=500;t=0.3" --prompt "why did the chicken cross the road?"

# rich SUT UID (json as a string)
modelgauge run-sut --sut '{"model":"google/gemma-3-27b-it","driver":"hfrelay", "provider":"nebius"}'  --prompt "why did the chicken cross the road?"

# rich SUT UID (json file)
modelgauge run-sut --sut tests/modelgauge_tests/data/sutdef.json --prompt "why did the chicken cross the road?"
```

Your JSON file or string can contain any or all of the fields in `modelgauge.sut_specification.SUTSpecification.fields` ("model" and "driver" are required, all others optional)

# Limitations

* This is a PoC that only works with modelgauge's run-sut.
* The sut / sut-def parsing in the run-sut function needs to be cleaned up and moved out of run-sut so it can be used by other functions

# Questions

* ~~Currently a SUTDefinition includes the SUT itself (model, driver, provider, maker, date) *and* the config (temperature, max tokens, etc). Would it be useful to also support handling the SUT config separately from the SUT definition? E.g. so you can do `modelgauge --sut google/gemma --sut-config shared_config.json` and use the same config options across multiple SUTs?~~ in review comment.

# Stories

* https://github.com/mlcommons/modelbench/issues/1145
* https://github.com/mlcommons/modelbench/issues/1097


src/modelgauge/cli.py
```diff
@@ -126,16 +129,32 @@ def list_secrets() -> None:
...
 @sut_options_options
-@click.option("--prompt", help="The full text to send to the SUT.")
+@click.option("--prompt", help="The full text to send to the SUT.", required=True)
 @click.option(
     "--top-logprobs",
     type=click.IntRange(1),
     help="How many log probabilities to report for each token position.",
 )
 def run_sut(
     sut: str,
```
wpietri: Now that we have two ways of defining the SUT, this might be clearer if these variable were renamed something like sut_str and sut_json. You should definitely keep the `--sut` option name; I think it's just the variable name that needs to change.

src/modelgauge/cli.py
```diff
@@ -144,14 +163,45 @@ def run_sut(
     top_k: Optional[int],
 ):
     """Send a prompt from the command line to a SUT."""
-    sut_instance = make_sut(sut)
+    if not sut and not sut_def:
+        raise click.BadOptionUsage("sut", "You must supply sut or sut-def")
+
+    # TODO: Clean this up
```
wpietri: Just to make sure, is the cleanup that you'll get have one method somewhere lower down that returns a SutDefinition and another one that takes a SutDefinition and returns an instance of a SUT?

src/modelgauge/command_line.py
```diff
@@ -128,6 +132,26 @@ def validate_uid(ctx, param, value):
     _bad_uid_error(registry, f"Unknown uid{plurality}: '{unknown_uids}'", hint=param.opts)
 
 
+def only_sut_or_sut_def(ctx, param, value):
```
wpietri: Is it worth making it so that this includes the exactly-one check?

src/modelgauge/sut_definition.py
```diff
@@ -0,0 +1,262 @@
...
+from typing import Optional
+import json
+import warnings
+
+from pydantic import BaseModel
+
+from modelgauge.dynamic_sut_metadata import DynamicSUTMetadata, RICH_UID_FIELD_SEPARATOR
+
+
+class SUTSpecificationElement(BaseModel):
```
wpietri: What's the value of using Pydantic here? I'm ok with using it for actual JSON input/output, but I think we've overused it otherwise in modelgauge, so I want to make sure we need it here.

src/modelgauge/sut_definition.py
```diff
@@ -0,0 +1,262 @@
...
+    @staticmethod  # sub for pydantic's verbose constructor
+    def make(name="", label="", value_type=type, required: bool = False):
+        return SUTSpecificationElement(name=name, label=label, value_type=value_type, required=required)
+
+
+class SUTSpecification:
+    """The spec a SUT definition needs to comply with"""
+
+    fields = {
+        "model": SUTSpecificationElement.make("model", "m", str, True),
```
wpietri: What's the ordering here? In my head there are the core fields and one or more sets of optional fields. From that perspective, these seem a bit jumbled.

src/modelgauge/sut_definition.py
```diff
@@ -0,0 +1,262 @@
...
+
+    @staticmethod
+    def from_json(data: str):
+        """Makes a SUTDefinition based on either a string of JSON or a file containing that JSON"""
+        try:
+            path = Path(data)
+            if path.exists():
+                return SUTDefinition.from_json_file(data)
+            else:
+                return SUTDefinition.from_json_string(data)
```
wpietri: If somebody has mistyped a filename, this will complain about invalid JSON, which will be kinda weird if that file contains valid JSON. Maybe before trying to parse it as JSON you check to make sure it looks like JSON? E.g., starts and ends with curly braces?

src/modelgauge/sut_definition.py
```diff
@@ -0,0 +1,262 @@
...
+        return generator.uid
+
+    def _generate_dynamic_uid(self):
+        generator = SUTUIDGenerator(self)
+        return generator._generate_dynamic_uid()
+
+    def validate(self) -> bool:
+        return self.spec.validate(self.data)
+
+    def add(self, key, value):
```
wpietri, line 123: It seems odd to me that a UID is mutable. Would it be workable to keep mutability to the construction phase?

src/modelgauge/sut_definition.py
```diff
@@ -0,0 +1,262 @@
...
+        return DynamicSUTMetadata(
+            model=self.data["model"],
+            driver=self.data["driver"],
+            maker=self.data.get("maker", None),
+            provider=self.data.get("provider", None),
+            date=self.data.get("date", None),
+        )
+
+
+class SUTUIDGenerator:
```
wpietri, line 156: What's the use case for this being a separately used object? I would expect I could just do something like `sut_definition.uid()` to get a uid.

tests/modelgauge_tests/test_uid_generator.py
```diff
@@ -0,0 +1,29 @@
...
+    d.add("date", None)  # workaround for the test in test_to_dynamic_sut_metadata setting the date (?!)
+    d.add("max_tokens", 500)
+    d.add("temp", 0.8)
+    d.add("top_p", 6)
+    d.add("top_k", 5)
+    d.add("top_logprobs", 1)
+    d.add("moderated", True)
+    d.add("reasoning", False)
+    g = SUTUIDGenerator(d)
+    assert g.uid == "openai/chatgpt-4o:openai:openai;mod:y;reas:n;mt:500;t:0.8;p:6;k:5;l:1"
```
wpietri: For me the semicolons and colons together make this hard to visually parse. How about an equal sign when it's expressing a value? So

`openai/chatgpt-4o:openai:openai;mod=y;reas=n;mt=500;t=0.8;p=6;k=5;l=1`

tests/modelgauge_tests/test_uid_generator.py
```diff
@@ -0,0 +1,29 @@
...
+
+    d.add("driver_code_version", "abcde")
+    d.add("date", "20250723")
+    assert g.uid == "openai/chatgpt-4o:openai:openai:20250723;dv:abcde;mod:y;reas:n;mt:500;t:0.8;p:6;k:5;l:1"
+
+    d.add("display_name", "my favorite SUT")
```
wpietri, line 27: What's the use for display_name? I was at first thinking it was what would end up in the modellab interface. But if we're not preserving case, then I'm not sure.

wpietri (approved):
Overall looks like excellent progress. I have some minor questions, but if they don't make sense for where you're headed, don't worry about it. As to the question of making configs separately configurable, I'd say leave it out until somebody needs is, as that looks like it wouldn't be hard to add later.

src/modelgauge/cli.py
```diff
@@ -126,16 +129,32 @@ def list_secrets() -> None:
...
+    "--sut",
+    "-s",
+    help="Which SUT to run.",
+    default="",
+    required=False,
+    callback=only_sut_or_sut_def,
+)
+@click.option(
+    "--sut-def",
+    "-s",
```
bkorycki: I dont think can both be `-s`

src/modelgauge/sut_definition.py
```diff
@@ -0,0 +1,262 @@
...
+                raise ValueError(f"Field {field.name} is required.")
+            if value is not None and not isinstance(value, field.value_type):
+                raise ValueError(f"Field {field.name} has wrong type.")
+        return True
+
+
+class SUTDefinition:
+    """The data in a SUT configuration file or JSON blob"""
+
+    spec: SUTSpecification = SUTSpecification()
```
bkorycki: I don't really understand why these are class variables and not instance variables

src/modelgauge/sut_definition.py
bkorycki: agreed

src/modelgauge/sut_definition.py
```diff
@@ -0,0 +1,262 @@
...
+        try:
+            with open(path, "r") as f:
+                data = json.load(f)
+                definition = SUTDefinition(data)
+            return definition
+        except Exception as exc:
+            raise ValueError(f"Unable to read data from {path}")
+
+    @property
+    def uid(self):
```
bkorycki, line 99: Why not just put the `_generate_uid` code directly in this function? Same fo `dynamic_uid`

src/modelgauge/sut_definition.py
```diff
@@ -0,0 +1,262 @@
...
+        "top_logprobs",
+        "display_name",
+    )  # this is the order past the dynamic SUT UID fields, which are fixed and at the head of this UID
+    field_separator = RICH_UID_FIELD_SEPARATOR
+    key_value_separator = ":"
+    blank_sub = "."
+    space_sub = "_"
+
+    def __init__(self, definition: SUTDefinition | None = None):
+        if definition:
```
bkorycki: In what case would the definition be None?

src/modelgauge/sut_definition.py
```diff
@@ -0,0 +1,262 @@
...
+            return definition
+        except Exception as exc:
+            raise ValueError(f"Unable to read data from {path}")
+
+    @property
+    def uid(self):
+        return self._generate_uid()
+
+    @property
+    def dynamic_uid(self):
```
bkorycki, line 105: I'm confused about what a dynamic UID is. Would calling `.uid` generate a different UID from `.dynamic_uid`? I feel like the caller should be able to treat suts and dynamic suts the same way and not have to identify the "type" of sut just to get the uid.

src/modelgauge/cli.py
rogthefrog: Yes, and also extract that code out of the specific run-job method because 1) it's needed in other CLI functions and 2) conceptually it doesn't really belong here.

src/modelgauge/sut_definition.py
rogthefrog: Arbitrary, something I made up just to have a predictable and stable structure.

src/modelgauge/cli.py
rogthefrog: Good catch.

src/modelgauge/sut_definition.py
rogthefrog: This is temporary behavior for the PoC, to be able to support the full UIDs with all the bits in the string and the current dynamic UIDs, whose parsing code can't handle anything past the date field.

src/modelgauge/cli.py
rogthefrog: Adding as a todo for when I tackle the cleanup in a separate PR. I'd like to discuss use just --sut to handle every way to pass in the sut identifier and get rid of --sut-def. Those are mutually exclusive, and it would make it nicer for the end user to only remember one way to pass in the sut.

src/modelgauge/command_line.py
rogthefrog: I'd like to merge the --sut and --sut-def into just one option (--sut) and this will go away.

src/modelgauge/sut_definition.py
rogthefrog: They shouldn't be, I will fix.

src/modelgauge/sut_definition.py
wpietri: Fab. What I'd suggest is: core parts go up front in a specific order, and then the optional fields go later with some sort of order that feels sensible. E.g., groups of related things, and then alphabetical within the groups.

src/modelgauge/sut_definition.py
rogthefrog: I like the API afforded by an `add` method, separate from the constructor, but you're right that it has consequences for mutability. I'll think on it.

wpietri: Maybe a builder pattern, then? A mutable thing that then produces the immutable thing?

src/modelgauge/sut_definition.py
rogthefrog: It's mostly a style and mental load thing. I don't like big classes, and there's enough logic in here that it would make the SUTDefinition class really large.

You can use `sut_definition.uid` and that's the happy path for code that uses SUTDefinition. It's a property so you don't even need to call a method.

src/modelgauge/sut_definition.py
rogthefrog: The "core parts" that make up the current dynamic SUT IDs (model, driver, maker, provider, date) are already at the head of this string (and handled separately in this code, to draw from that existing code). 

What do you think the groupings should be? I'm thinking moderated and reasoning are essential to the identity of the SUT and should be near the top. For the others I'm not so sure.

src/modelgauge/sut_definition.py
rogthefrog: To test it separately. I like to keep the property methods light, only dispatching whatever validation or formatting they do to a separate method.

src/modelgauge/sut_definition.py
rogthefrog: Type checking.

src/modelgauge/sut_definition.py
bkorycki: I see, but the`_generate_uid` is also a one-liner. Idk, it feels a little unnecessarily cluttered to me

src/modelgauge/sut_definition.py
wpietri: Is there a reason this needs type checking more than other things? If so, can we get the type checking without inheritance? And if not, maybe a data class is sufficient?

src/modelgauge/sut_definition.py
wpietri: I guess I'm fine with breaking the classes up, but I'm not fine with the lifecycle. UIDs should be immutable, and the `SutUidGenerator` is basically just a giant function. It doesn't need to be an object that gets instantiated every time the method gets called.

src/modelgauge/sut_definition.py
rogthefrog: When using types that aren't string or numeric types (e.g. in this case `value_type` is a `type` type), I like the enforced type checking because it avoids bugs in my code in situations where Python can't just "do the right thing" (e.g. if you expect a float but pass an int).

Dataclasses don't do that out of the box. You can achieve the same effect with e.g. a property that throws an error if you use the wrong type for the variables you care about. But that's more code than just subclassing BaseModel.

I'm not married to the type checking here. I'll switch to a dataclass.

tests/modelgauge_tests/test_uid_generator.py
rogthefrog: If modellab exposes a UI for display name and that's not preserved as part of the SUT definition, then it needs to make sure it doesn't send it to modelrunner, and it's only stored in a database and not in the journals or logs. I don't like the idea of the same conceptual thing (a SUT) not having the same attributes traveling together all the time, because it makes things brittle.

src/modelgauge/sut_definition.py
wpietri: Thanks. I'm still not sure exactly which types you think might get improperly used, so some solution that makes that clearer would help. Even a Pydantic dataclass would be step up, rather than bringing in all the baggage that BaseModel carries with it.

bkorycki:
I guess I'm curious how individual SUTs will express and store their uids in this new uid type. But if you want to do that in a separate PR thats fine.

PR #1150, opened 2025-07-28
Security test + benchmark

bkorycki:
This does not yet enable users to run it from CLI. I will do that in a separate PR


src/modelbench/hazards.py
```diff
@@ -140,6 +142,65 @@ def test_uids(self) -> List[str]:
...
+        self._evaluator = evaluator
+        super().__init__()
+
+    def name(self):
+        return f"Security {self.hazard_key.title()} Hazard V0.1"
+
+    def key(self):
+        return f"security_hazard-0_1-{self.hazard_key}"
+
+    def score(self, sut_scores: Mapping[str, TestRecord]) -> "HazardScore":
```
bkorycki, line 160: This is literally copy-pasted from the other Hazard

wpietri (approved):

src/modelbench/hazards.py
rogthefrog: Does it need to be? Can it be inherited?

bkorycki: To that in the present situation `SecurityHazard` would have to inherit from `SafeHazardV1`, and it would override every single method except for `score`. Conceptually I don't think it makes sense.
Ideally, benchmarks would be more composable and scoring would be its own object that can be used for different benchmarks. Maybe that can be done in the make modelbench more composable card.

PR #1155, opened 2025-07-29
Swap mixtral endpoint for ensemble annotator

bkorycki:
Is it ok to use the same uid?

### DIFF
--- a/src/modelgauge/private_ensemble_annotator_set.py
+++ b/src/modelgauge/private_ensemble_annotator_set.py
@@ -11,7 +11,7 @@ try:
         config as MISTRAL_7B_RUBY_CONFIG,
     )  # type: ignore
     from modelgauge.annotators.mistral_8x22b_instruct_annotator import (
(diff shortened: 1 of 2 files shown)


src/modelgauge/private_ensemble_annotator_set.py
```diff
@@ -11,7 +11,7 @@
         config as MISTRAL_7B_RUBY_CONFIG,
     )  # type: ignore
     from modelgauge.annotators.mistral_8x22b_instruct_annotator import (
-        MISTRAL_8x22B_PE_TAMALPAIS_2024_09_09_CONFIG as MISTRAL_8x22B_CONFIG,
+        MISTRAL_8x22B_PE_TAMALPAIS_2024_09_09_ON_TOGETHER_DEDICATED as MISTRAL_8x22B_CONFIG,
```
bkorycki, line 14: I updated this to the actual config that is used in the annotator. It doesn't have any tangible effects here because they use the same name/uid (which is so wrong). we should clean up the old model tune code at some point.

PR #1157, opened 2025-07-29
Security test

bkorycki:
Part 2 of security benchmark.

- new `security_benchmark` modelbench cli
- FAKE reference score. both default and ensemble evaluators yielded `0.0` for the reference suts.


wpietri (approved):
Looks good!

PR #1160, opened 2025-07-30
Move prompt_engineered_annotator to modelbench.

superdosh:
This is copy/pasted from private, with the exceptions:
* We use `SUT_FACTORY` instead of `SUTS` on line 72.
* Also copies in `flatten_conversation_to_str`.
* Some typing changes to pass mypy

There will be a separate PR in private which removes prompt_engineered_annotator from there.


superdosh:
> Can you put this in `src/modelgauge/annotators`?

Good idea; will minimize changes with modelbench-private! Done! @bkorycki

bkorycki (approved):

PR #1162, opened 2025-07-30
Fix together dynamic factory

bkorycki:
Together's `models` API is broken. Roger opened an [issue](https://github.com/togethercomputer/together-python/issues/343) 2 weeks ago that has still not been acknowledged.

This is one of the issues causing smoke test failures.

This PR fixes the issue by no longer relying on the `models` call.


rogthefrog (approved):
Awesome.

PR #1169, opened 2025-07-31
re-enable two innocent SUTs and disable SUTs guilty of failing auth

rogthefrog:
Not really a fix for https://github.com/mlcommons/modelbench-private/issues/127

### DIFF
--- a/src/modelgauge/suts/meta_llama_client.py
+++ b/src/modelgauge/suts/meta_llama_client.py
@@ -115,14 +115,22 @@ class MetaLlamaModeratedSUT(PromptResponseSUT[MetaLlamaChatRequest, MetaLlamaMod
         return SUTResponse(text=text)
 
 
-# Unauthorized models: ["Llama-4-Scout-17B-16E-Instruct-FP8", "Llama-4-Maverick-17B-128E-Instruct-FP8"]
-CHAT_MODELS = ["Llama-3.3-8B-Instruct"]
+CHAT_MODELS = ["Llama-3.3-8B-Instruct", "Llama-4-Scout-17B-16E-Instruct-FP8", "Llama-4-Maverick-17B-128E-Instruct-FP8"]
 
 for model_name in CHAT_MODELS:
     SUTS.register(MetaLlamaSUT, "meta-" + model_name.lower() + "-llama", model_name, InjectSecret(MetaLlamaApiKey))
-    SUTS.register(
-        MetaLlamaModeratedSUT,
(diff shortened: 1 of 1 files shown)


bkorycki (approved):
thanks!

src/modelgauge/suts/meta_llama_client.py
```diff
@@ -115,14 +115,22 @@ def translate_response(self, request: MetaLlamaChatRequest, response: MetaLlamaM
...
-        InjectSecret(MetaLlamaApiKey),
-    )
+
+    # Disabled 2025-07-31 because the moderations.create call always fails with an authentication error
+    # even when we set the api_key in the Client object OR when we add the auth header manually as extra_headers
+    # in the moderations.create argument.
+    # Probable server side error. Maybe we:
+    # * need to request access to the moderated SUTs separately
+    # * get a special API key for them
+    # * are experiencing a server-side bug.
```
superdosh, line 129: Just one call out -- last time we submitted a bug to meta, they fixed within a week, so it might be worth opening an issue with them again?

wpietri: Agreed with Vishal. At the very least we should create a card for sorting out the problem. Also, I think the correct place to temporarily disable a failing test is in the tests., not in the production code. The UID should go into `SUTS_THAT_WE_DONT_CARE_ABOUT_FAILING` for now once the fix card has been created.

wpietri: I have also added card #1170 so we can work toward replacing this janky approach to making sure our connections still work.

wpietri (changes requested):
As mentioned, please remove it from the test set of SUTs.

src/modelgauge/suts/meta_llama_client.py
rogthefrog: > Agreed with Vishal. At the very least we should create a card for sorting out the problem. Also, I think the correct place to temporarily disable a failing test is in the tests., not in the production code. The UID should go into `SUTS_THAT_WE_DONT_CARE_ABOUT_FAILING` for now once the fix card has been created.

These don't work at all. It's not just the smoke tests. If you try to run them via modelgauge, the same problem happens.

wpietri: I believe that, but I don't see a need to comment out the code. If we believe they will never work, we should delete them. Otherwise, we should leave the code in, possibly working to fix them.

PR #1174, opened 2025-08-04
Allow annotator workers to cache.

superdosh:
I noticed that while we pass the `cache_dir` to the `PromptSUTWorkers`, we don't pass it to the `AnnotatorWorkers`. There doesn't seem to be a good reason for not doing this.


superdosh:
Tagging everyone in case there's some super important reason we weren't caching here.

wpietri (commented):
I'll defer to @bkorycki on this one; she knows this code and these users best. But it sounds reasonable to me.

PR #1181, opened 2025-08-06
benchmark cli groups

bkorycki:
These are the new benchmark commands:

- `benchmark general <options>`
- `benchmark security <options>`

The grouping doesn't actually do anything at this moment. I was thinking of attaching the common benchmark options to the `benchmark` group, but then you would have to do
`benchmark --sut sut  <other common options> general --prompt-set practice <other benchmark specific options>`
And I think it would be annoying to have to keep track of which options go where. But this sets us up to move  other common benchmark behavior to this group level in the future.

### DIFF
--- a/src/modelbench/cli.py
+++ b/src/modelbench/cli.py
@@ -95,6 +95,11 @@ def cli() -> None:
     load_plugins(disable_progress_bar=True)
 
 
+@cli.group()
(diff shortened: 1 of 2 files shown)


dhosterman (approved):
Looks solid to me!

PR #1182, opened 2025-08-06
Lazy load plugins

superdosh:
Here's the initial plan for lazy loading SUTs/Annotators. Next step would be simply to move all the plugins to the appropriate modules and list them in the annotator/SUT maps.

### DIFF
--- a/src/modelgauge/cli.py
+++ b/src/modelgauge/cli.py
@@ -287,7 +287,6 @@ def run_test(
     help="Which registered annotator(s) to run",
     multiple=True,
     required=False,
-    callback=validate_uid,
 )
 @click.option(
     "ensemble",
--- a/src/modelgauge/cli.py
+++ b/src/modelgauge/cli.py
@@ -52,6 +52,10 @@ def list_command() -> None:
     display_header(f"SUTS: {len(suts)}")
     for sut, sut_entry in suts:
         display_list_item(f"{sut} {sut_entry}")
+    annotators = ANNOTATORS.items()
+    display_header(f"Annotators: {len(annotators)}")
+    for annotator, annotator_entry in annotators:
+        display_list_item(f"{annotator} {annotator_entry}")
     tests = TESTS.items()
     display_header(f"Tests: {len(tests)}")
     for test, test_entry in tests:
--- a/src/modelgauge/cli_lazy.py
+++ b/src/modelgauge/cli_lazy.py
@@ -1,0 +1,53 @@
+import click
+
+from modelgauge.annotator_registry import ANNOTATOR_MODULE_MAP, ANNOTATORS
+from modelgauge.load_plugins import load_plugins, load_plugin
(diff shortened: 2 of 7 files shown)


src/modelgauge/cli.py
```diff
@@ -283,7 +287,6 @@ def run_test(
     help="Which registered annotator(s) to run",
     multiple=True,
     required=False,
-    callback=validate_uid,
```
superdosh, line 286: This doesn't work with the lazy loading as it'll always be invalid until it's loaded.

src/modelgauge/cli_lazy.py
superdosh, line 1: Naming of these vars/classes is off, but the general idea is that:

1. For the list commands, we load a lot of stuff, but only what's needed. (e.g. only the suts namespace for list-suts.)
2. For the `run-*` commands, we inspect the params when the command is given, and load as needed.

tests/modelgauge_tests/test_cli.py
superdosh, line 1: There's one more test failure that has to do with the JSON sut definitions. Shouldn't be too hard to resolve, but want to discuss with @rogthefrog .

src/modelgauge/load_plugins.py
```diff
@@ -29,12 +29,21 @@ def _iter_namespace(ns_pkg: ModuleType) -> Iterator[pkgutil.ModuleInfo]:
...
 
-def list_plugins() -> List[str]:
+def list_plugins(ns: str | None = None) -> List[str]:
     """Get a list of plugin module names without attempting to import them."""
+    namespaces = [ns] if ns else ["tests", "suts", "runners", "annotators"]
     module_names = []
-    for ns in ["tests", "suts", "runners", "annotators"]:
-        for _, name, _ in _iter_namespace(getattr(modelgauge, ns)):
-            module_names.append(name)
+    for ns in namespaces:
```
bkorycki, line 36: Ah I'm kind of confused. Can you explain the changes here?

src/modelgauge/cli_lazy.py
```diff
@@ -0,0 +1,53 @@
...
+
+        # now we lazy load any additional plugins based on the command line args
+        # we have to copy args as make_context mutates it
+        cmd_ctx = cmd.make_context(cmd_name, args.copy(), parent=ctx, resilient_parsing=True)
+
+        test_name = cmd_ctx.params.get("test")
+        if test_name:
+            load_plugin("tests")
+
+        # we use both sut and sut_uid
```
bkorycki, line 30: lol dang we should change that!

src/modelgauge/load_plugins.py
superdosh: So `_iter_namespace` takes a module, so for a case where one of the things we're importing has submodules, we have to get that module first, which is what the loop over submods is doing. 

I'm going to take a crack at simplifying this though.

superdosh:
Thanks for the comments! I'm going to close this so you don't get spammed and re-open without draft once it's ready!

PR #1183, opened 2025-08-07
Parameterize column names when running modelgauge pipelines

bkorycki:


src/modelgauge/cli.py
```diff
@@ -354,6 +373,15 @@ def run_job(
     else:
         annotators = None
 
+    # These values should not be passed if they were not specified by the user.
```
superdosh: Would we be able to resolve this issue if we did defaults? I see it probably ends up being complicated due to us accepting multiple possible default values?

src/modelgauge/data_schema.py
```diff
@@ -69,25 +81,25 @@ class PromptResponseSchema(PromptSchema):
...
+    def __init__(
+        self,
+        header: list[str],
+        prompt_uid_col: Optional[str] = None,
+        prompt_text_col: Optional[str] = None,
+        sut_uid_col: Optional[str] = None,
+        sut_response_col: Optional[str] = None,
+        validate: bool = True,
+    ):
+        # Don't validate superclass. Want to get *all* missing columns before validating.
```
superdosh: What if you called the superclass's init last? I don't think you're using anything below that relies on the initialization. It should allow you to avoid using the `validate` flag?

src/modelgauge/pipeline_runner.py
```diff
@@ -166,9 +167,17 @@ def _initialize_segments(self):
 
 
 class AnnotatorRunner(PipelineRunner):
-    def __init__(self, annotators, **kwargs):
+    def __init__(
+        self, annotators, prompt_uid_col=None, prompt_text_col=None, sut_uid_col=None, sut_response_col=None, **kwargs
+    ):
         self.annotators = annotators
         self.annotator_workers = None  # Convenience pointer.
+
```
superdosh: I feel like the way we're passing these 4 values around is calling out for an object, maybe?

src/modelgauge/data_schema.py
bkorycki: I did try that but it breaks because the super class initializes `self._missing_columns`. Which is used in the `find_column` calls.

src/modelgauge/pipeline_runner.py
bkorycki: Yeah I am unhappy with that as well. Maybe I will try to build the input dataset object in `build_runner`, which will then be passed to the Runner object.

superdosh: Not a blocker, IMO, but does seem like it might clean things up a bit.

src/modelgauge/data_schema.py
superdosh: Could make that stateful solve that? Something like having a `self._initialize_missing_columns_if_needed()` called both in the superclass and subclass would be okay.

bkorycki: hm, I don't really see that advantage of doing that over doing optional validation to be honest

superdosh: The main issue I have is that it's strange for the `__init__` to be able to produce something that might not be valid, which a caller now needs to know how to complete (e.g. by calling .validate() itself). That doesn't feel right.

src/modelgauge/pipeline_runner.py
wpietri: I think giving the dataset to the runner seems like a good choice. The runners aren't really about files, but input data, so making them all deal with objects with a common interface makes sense.

src/modelgauge/cli.py
```diff
@@ -354,6 +373,15 @@ def run_job(
...
+    kwargs = {
+        "prompt_uid_col": prompt_uid_col,
+        "prompt_text_col": prompt_text_col,
+        "sut_uid_col": sut_uid_col,
+        "sut_response_col": sut_response_col,
+    }
+    kwargs_filtered = {k: v for k, v in kwargs.items() if v is not None}
```
wpietri: This could probably use a more descriptive names, like column_names or column_arguments.

src/modelgauge/data_schema.py
wpietri: Yeah, this seems squirrely to me. The missing_columns and validate instance variables seem to be particular to just one phase of the object's lifecycle. How about you have two methods, one that finds the missing columns and the other that throws the exception?

bkorycki: @wpietri Sorry what do you mean? There already are 2 different methods for finding the missing columns and then for throwing the exception.

What if I remove the `validate` call from initialization and have the calling code call `schema.validate()` instead?

src/modelgauge/cli.py
bkorycki: I was able to delete this code by building the dataset in the runner.

src/modelgauge/cli.py
bkorycki: I was able to delete this code by building the dataset in the runner.

src/modelgauge/pipeline_runner.py
bkorycki: Done.

src/modelgauge/data_schema.py
bkorycki: Okay I tried something else. Let me know what you think.

src/modelgauge/pipeline_runner.py
```diff
@@ -27,14 +27,14 @@ class PipelineRunner(ABC):
     def __init__(
         self,
         num_workers,
-        input_path,
+        input_dataset,
```
superdosh, line 30: I like this. And because we use `build_runner` in modelplane, it should just work there!

src/modelgauge/data_schema.py
superdosh, line 1: I like this refactor!

src/modelgauge/data_schema.py
wpietri: That is shorter and somewhat better, but it still seems roundabout to me.

The broad principle I'm working from here is that a domain object should have structures and lifecycles that match how people use the term. So the previous possibility that one could end up with an invalid schema, or had to know to validate it before using it, didn't match how people talk about schemas. ([Here's a good look](https://matthiasnoback.nl/2018/07/objects-should-be-constructed-in-one-go/)  at a similar issue.) This new version is better in that regard.

But the `_missing_columns` member is still odd from that perspective. Schemas don't have missing columns. It's a field that is only used during construction. Having to catch and ignore that exception only to re-throw it later is also weird. 

One solution would be to reify the use case and come up with a SchemaBuilder object, which is allowed to be inconsistent, but when you call `build()` you always get back a valid Schema. Here I think that's overkill.

Another way to solve this is to split up the setup work into phases. So you could have an abstract base class that's `Schema` or `FileSchema` or something. Then in the superclass init you have something like:
```
    def __init__(self, header):
        self.header = header
        self._find_columns() # or maybe self._bind_columns()
        missing_fields = self._find_missing_fields()
        if missing fields:
            raise SchemaValidationError(missing_fields)
```

Then none of the subclasses do more work in `init` than declaring fields. Here `_find_columns` gets overridden by subclasses to do the work of finding the right names. You could do the same for `_find_missing_fields()`, or you could have that be a bit of introspection where the object is checked for null fields.

A third way out of this is instead of using fields, you have a real data structure, say, `column_mappings`, and the Schema subclasses become about managing that instead of directly dealing with the fields of the object. If you wanted could preserve the interface by adding properties in the subclasses that pull from the data structure.

I'm not attached to any particular solution here, but I think that you've let the history of this constrain your solution too much. As you add complexity, sometimes it's better to step back a bit and rebuild rather than trying to Rube Goldeberg something into working.

src/modelgauge/data_schema.py
bkorycki: thanks!

src/modelgauge/data_schema.py
bkorycki: Okay thanks for the suggestion. Done

wpietri (approved):
Looks good. Thanks!

PR #1186, opened 2025-08-07
Pre-work for lazy loading plugins

superdosh:
This PR sets the foundation for lazy loading plugins.

Notes:

* This does NOT address any lazy loading for dynamic SUTs. This is ONLY for plugin-based SUTs and annotators. This will allow moving all plugins into core without loading them unless we're running that specific SUT.
* The way we do this is (for non-dynamic SUTs) cataloging them in `ANNOTATOR_MODULE_MAP` or `LEGACY_SUT_MODULE_MAP`. I don't think there's another way to do this, but open to ideas.
* We also do something more like what we used to (loading (almost) everything) for the `list-*` commands in the CLI.
* This PR also refactors some of the SUTDefinition handling to be more usable in other places.

Follow ups:

* Actually move the plugins to core. I thought the PR would be unwieldy doing that here.
* Make the huggingface/openai/together dynamic sut factories lazily loaded as well. (Shouldn't be too hard, but didn't want to handle at the same time as the plugins.)

### DIFF
--- a/src/modelgauge/load_plugins.py
+++ b/src/modelgauge/load_plugins.py
@@ -31,7 +31,6 @@ def _iter_namespace(ns_pkg: ModuleType) -> Iterator[pkgutil.ModuleInfo]:
 
 def list_plugins() -> List[str]:
     """Get a list of plugin module names without attempting to import them."""
-    namespaces = ["tests", "suts", "runners", "annotators"]
     module_names = []
     for ns in ["tests", "suts", "runners", "annotators"]:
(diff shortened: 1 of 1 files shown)


wpietri:
Looks plausible. As you go forward, please measure the startup time differences, mainly so we can be sure that we're really avoiding loading a bunch of code we don't need.

superdosh:
Actually, realizing that next step will be to port this to the modelbench cli, as this PR only addresses the modelgauge cli!

Issue #1192, opened 2025-08-11
Add rich SUT UIDs to all commands in modelgauge and modelbench


rogthefrog:
[The PoC only added rich SUT UIDs to one command in modelgauge](https://github.com/mlcommons/modelbench/issues/1145). Add them to all other commands that accept a SUT UID.

PR #1194, opened 2025-08-11
Add rich sut uids everywhere

rogthefrog:
The rich SUT UID PoC only added that capability to modelgauge.run_sut. This PR adds it elsewhere.

https://github.com/mlcommons/modelbench/issues/1192


superdosh (approved):
LGTM!

PR #1200, opened 2025-08-13
Swap in real security prompts

bkorycki:


src/modelgauge/tests/security.py
```diff
@@ -69,32 +70,32 @@ def get_annotators(cls) -> List[str]:
...
-                source_url=f"{prompt_set_base_download_url}/security_results.parquet",
+                source_url=f"{prompt_set_base_download_url}/security_attacks_v0.5.csv",
                 headers=headers,
             )
         }
 
     def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]:
         test_items: List[TestItem] = []
         local_file = dependency_helper.get_local_path("prompts")
-        parquet_file = pq.ParquetFile(local_file)
```
rogthefrog, line 87: Will we ever need this again?

src/modelgauge/tests/safe_v1.py
```diff
@@ -28,6 +28,30 @@
 from pydantic import BaseModel
 
 
+class Hazards(list):
```
rogthefrog, line 31: I love this idea.

src/modelgauge/tests/security.py
bkorycki: Kurt says probably not for this benchmark

PR #1202, opened 2025-08-14
Security evaluator

bkorycki:
This sets up the security benchmark to use a new prompt engineered evaluator.
This annotator is different from other ones in that it uses the seed_prompt, which is different from the prompt that is sent to the SUT. You can see it in the related modelbench-private [PR](https://github.com/mlcommons/modelbench-private/pull/133)

I also registered a new SUT (gemma 3 12b) to be used as a reference sut for the security benchmark. I will update the standards and calibration logic in a later PR.


src/modelgauge/annotators/prompt_engineered_annotator.py
```diff
@@ -70,6 +70,11 @@ def __init__(
         else:
             raise ValueError(f"Unsupported LLM type: {self.config.llm_type}")
 
+    @staticmethod
```
wpietri, line 73: Why is this marked static?

src/modelbench/benchmark_runner.py
```diff
@@ -503,7 +504,11 @@ def _check_annotators_working(self, run: TestRunBase):
         def check_annotator(annotator: CompletionAnnotator):
             try:
                 raw_request = annotator.translate_request(
-                    TestItem(prompt=TextPrompt(text="Why did the chicken cross the road?"), source_id="ignored"),
+                    TestItem(
+                        prompt=TextPrompt(text="Why did the chicken cross the road?"),
+                        source_id="ignored",
+                        context=SecurityContext(seed_prompt="needed by security tests. ignored by general."),
```
wpietri, line 510: I'm generally uncomfortable with things that get ignored. Seems like it's telling us something about the modeling. But let's ignore it for now and keep an eye on it.

wpietri (approved):

src/modelgauge/annotators/prompt_engineered_annotator.py
bkorycki: Because it's class-level function that doesn't depend on `self`.

PR #1205, opened 2025-08-15
Refactor standards

bkorycki:
This is an attempt to refactor the way we handle standards.

I think there were 2 main problems with the standards. Practically, it was very difficult to add calibrations for new benchmarks. Secondly, the design didn't make sense from an OOD perspective. There was no relationship between standards and the actual benchmark objects.

Here are my proposed changes:
- Benchmarks define their own standards objects
- Benchmarks are responsible for their own calibration.
- We have separate standards files for different benchmark families i.e.  general and security. Calibrating one benchmark shouldn't alter a file that is used by another benchmark imho.

Any level of feedback is appreciated at this time, but I am especially looking for feedback on the overall re-design before proceeding on with this refactor.


wpietri (approved):
I like where this is headed, and I think you should take it a bit further. I think it shouldn't be one file per benchmark family, but one file per benchmark. I also think calibration should happen on the benchmark level, not the benchmark family. That's partly because we so far have only done it, and partly because once a calibration has been done, changing it accidentally should be pretty hard.

We should also review the calibration file format to see if there's anything we want to add. At first glance I couldn't think of anything except that it should be clear what got changed when, which separate files would fix. But maybe there's more metadata we'd like. How we ran the SUT? The name of the journal? Dunno, but let's talk it through together.

bkorycki:
I just pushed my latest changes based on feedback I received last week. Mainly:
- One standards object per benchmark (not just per benchmark family)
- Remove circular dependency between Standards and Benchmarks.

Note: I tried adding a calibrator object but it didn't really make sense. All the relevant code is in a short `calibrate` function. I didn't see the benefit of turning it into an object

I'm not happy with the current design. There is a misalignment between how benchmarks get standards and how benchmark standards are calibrated. 
For example, the general en_us practice benchmark with DEFAULT evaluator should be using the standards for the general en_us practice benchmark with ENSEMBLE evaluator. With the current implementation, this is true when getting the benchmark's standards:
```
    @property
    def standards(self) -> Standards:
        uid = self.uid.replace("demo", "practice").replace("default", "ensemble")
        return Standards.get_standards_for_benchmark(uid)
  ```

However, this does not work at calibration time.
If I run `modelbench calibrate general en_us practice --evaluator default`, what SHOULD happen is we run the ENSEMBLE benchmark and save those standards (assuming those standards don't exist yet). But what actually happens is the default benchmark will run.

I believe what is missing here is the concept of a "reference benchmark". I think it needs to go beyond UID mangling in the `standards` property. My proposal is for the Benchmark to look like:

```
    def reference_benchmark(self):
        if self.evaluator == "default":
            return GeneralPurposeAiChatBenchmarkV1(self.locale, self.prompt_set, "ensemble")
        return self
    
    def standards(self):
        return Standards(self.reference_benchmark().uid)
```

Then at calibration time, we just run `benchmark.reference_benchmark()`.
@wpietri @rogthefrog

**Comment by rogthefrog** [comment]
> maybe there's more metadata we'd like. How we ran the SUT? 

I'd love to see what in any artifact produced by any kind of run. Maybe `command_run: "modelgauge <foo> <bar>"` 

This makes it easy to rerun for any reason, e.g. updating the data, or replicating it for debugging. 

FWIW that's what I (try to) do whenever I put something in DVC. I include a `command.sh` or similar file that includes the exact CLI command that caused an artifact to be generated, so that the data can be recomputed. It's part of data lineage, IMO, and useful for auditing and troubleshooting.

File: src/modelbench/benchmarks.py [hunk 1]
```diff
@@ -141,6 +158,18 @@ def __init__(self, locale: str, prompt_set: str, evaluator: str = "default"):
         self.evaluator = evaluator
         super().__init__()
 
+    @property
+    def standards(self) -> Standards:
+        #  TBH I feel like it was better to create the benchmark objects and get the UID directly instead of mangling this way.
```
wpietri: I agree this is a bit hacky, but I think instantiating a BenchmarkDefinition and throwing it away is also hacky. And this at least has the version of being simple.  If you want something more expressive, I'd suggest extending the HasUid stuff so you can write the code in terms that make it clear how you're thinking about it.

src/modelbench/benchmarks.py
```diff
@@ -141,6 +158,18 @@ def __init__(self, locale: str, prompt_set: str, evaluator: str = "default"):
...
+        #  All benchmarks use the ensemble reference scores.
+        #  Demo prompt set uses the practice standards.
+        uid = self.uid.replace("demo", "practice").replace("default", "ensemble")
+        return Standards.get_standards_for_benchmark(uid)
+
+    @property
+    def reference_suts(self) -> list[str]:
+        return ["gemma-2-9b-it-hf", "llama-3.1-8b-instruct-turbo-together"]
```
wpietri, line 179: I really love that the reference SUTs are part of the BenchmarkDefinition now. That's definitely moving in a DDD direction.

src/modelbench/cli.py
```diff
@@ -285,15 +285,18 @@ def check_benchmark(benchmark):
 
 
 def score_benchmarks(run):
+    # TODO: Can we get rid of this method
```
wpietri: When in doubt, throw it out. If all the tests pass, you're good.

src/modelbench/hazards.py
```diff
@@ -47,8 +47,16 @@ def tests(self, secrets: RawSecrets) -> List[PromptResponseTest]:
                 self._tests.append(TESTS.make_instance(uid, secrets=secrets))
         return self._tests
 
+    def set_standard(self, standards: Standards):
```
wpietri, line 50: Is changing the flow of this necessary? It makes it less composable: one can't just instantiate a hazard on its own; one has to get one from a BenchmarkDefinition.

src/modelbench/standards.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self, path: pathlib.Path):
+        self.path = path
+        self._data = None
+
+    @classmethod
+    def get_standards_for_benchmark(cls, uid) -> "Standards":
+        return Standards(pathlib.Path(__file__).parent / "standards" / f"{uid}.json")
+
+    def reference_standard_for(self, hazard: "HazardDefinition") -> float:
+        if not self._data:
```
wpietri: Is there a reason the data isn't loaded in `__init__`? Generally objects should be ready to go as soon as they're instantiated. This defers any blowups to later, which can be harder to pin down.

src/modelbench/standards.py
```diff
@@ -0,0 +1,93 @@
...
+            raise ValueError(
+                f"Can't find standard for hazard UID {hazard.uid}. No hazard with key {hazard.key()} in {self.path}"
+            )
+        return self._data["reference_standards"][hazard.key()]
+
+    def assert_can_write(self):
+        if self.path.exists():
+            raise FileExistsError(f"Error: attempting to overwrite existing standards file {self.path}")
+
+    def assert_standards_exist(self):
```
wpietri: I think this would be better as a class method that takes a uid.

src/modelbench/standards.py
```diff
@@ -0,0 +1,93 @@
...
+        if not self.path.exists():
+            raise FileNotFoundError(
+                f"Standards file {self.path} does not exist. Please run `modelbench calibrate` on your desired benchmark."
+            )
+
+    def write_standards(self, sut_scores: dict[str, list["HazardScore"]]):
```
wpietri: This is a sort of constructor, and so it should also be a class method. Or at least partly. E.g., `Standards.from_hazard_scores()` should take the uid and hazard scores and return a Standards instance. Then a `.save` or `.write` can do the writing.

rogthefrog:
>     def reference_benchmark(self):
>         if self.evaluator == "default":
>             return GeneralPurposeAiChatBenchmarkV1(self.locale, self.prompt_set, "ensemble")
>         return self
>     
>     def standards(self):
>         return Standards(self.reference_benchmark().uid)
> ```
> 
> Then at calibration time, we just run `benchmark.reference_benchmark()`. @wpietri @rogthefrog

I like this approach, because it puts the logic in one place and reduces the need for the operator to know what's needed to run calibration for a benchmark.

wpietri (approved):
I think this is definite progress, and I am fine with you merging it.

As to the hackiness of UID string mangling, I'm uncomfortable with an approach where we instantiate a benchmark only to get the uid and  throw it away. It makes sense in mechanistic terms, but not in domain terms. If you really want to fix it, I think we need a richer concept of a uid, so that you can express what you want more clearly. Here I think that means extending HasUids to have the sort of structure we've talked about for SUT uids. Then you could express it as something like:
```
Standards.load(self.uid.with('evaluator','private'))
```

src/modelbench/benchmarks.py
bkorycki: I think this is addressed by my proposal in this [comment](https://github.com/mlcommons/modelbench/pull/1205#issuecomment-3222241664).
This would instantiate a BenchmarkDefinition but it's not just used for it's UID. It'a also used for calibration.

src/modelbench/hazards.py
bkorycki: Calibration is done at a benchmark level and standards are created at the benchmark level. So the hazard reference scores have to be supplied by the benchmark. 
I hear your concern, but I don't really see a way around it.

src/modelbench/standards.py
bkorycki: Yes, to enable us to run uncalibrated benchmarks.

bkorycki:
@rogthefrog 
> I'd love to see what in any artifact produced by any kind of run. Maybe `command_run: "modelgauge <foo> <bar>"`

Added!

src/modelbench/standards.py
bkorycki: The thing is the standards object already exists during calibration, it's not really creating a new one so this isn't really a constructor.

bkorycki:
@wpietri Is there anything outside of modelbench code that parses standards files?
The only non-additive change is that the hazard standards are keyed by the hazard key instead of hazard UID. This is necessary for benchmarks to be able to share standards file.

I'm happy to add a file format version number. I'll do that now.

src/modelbench/standards.py
wpietri: Let me see if I can explain better. Standards have kind of a dual existence.

One is how we talk about a Standards set. Before calibration happens, it doesn't exist. After, it does. Then later we might re-calibrate, and a Benchmark's Standards get replaced.

Then there is how it works in the code. The Standards object can exist with data, but sometimes it doesn't have any. That's not because it's really a Standard at that point. It's just the way other parts of the code are written force us to do something weird.

All the checks in the Standards object that load the data are a sign that something is wrong. Not here, but elsewhere, forcing this to be awkward. As much as possible, our code should work like we talk about it.

src/modelbench/standards.py
wpietri: Are we sure there's such a thing as an uncalibrated benchmark? If there isn't, maybe the calibration should be done based on information about the BenchmarkDefinition, but not using it. If there is, then maybe the uncalibrated Benchmark should start with some sort of NullStandards or DefaultStandards object.

src/modelbench/standards.py
bkorycki: Ok that makes sense! I updated the code. Let me know what you think.

src/modelbench/standards.py
```diff
@@ -103,15 +148,24 @@ def _write_file(self, reference_suts, reference_standards, reference_benchmark,
...
             },
         }
-        with open(self.path, "w") as out:
+
+        path = self._benchmark_standards_path(self._reference_benchmark)
+        with open(path, "w") as out:
             json.dump(result, out, indent=4)
+
+
+class NullStandards:
```
wpietri, line 167: Oh interesting! I was imagining it would have dummy numbers, like 0.5 across the board, so that uncalibrated benchmarks could run. But if this works for now, I'm fine with it.

bkorycki: Calibration runs don't do grading, so yes this works!

Issue #1132, opened 2025-07-15
Dynamic SUTs for remaining plugins


bkorycki:

superdosh:
Related: we definitely just want an OpenAISUT that accepts an endpoint because it is exactly how the vLLM endpoints will work. It's possible @wpietri has already created this?

rogthefrog:
@superdosh with the code in this PR, all you need to do is add the VLLM info to secrets.toml and run a dynamic SUT.

E.g. say your VLLM instance is called "superdosh," it's hosted at http://example.com/v1/completions, and it runs gpt-oss-20b. You can do:

In secrets.toml

```toml
[superdosh]
api_key = "your vllm api key"
base_url = "http://example.com/v1/completions"
```

And run the SUT like this:

```bash
modelgauge run-sut --sut openai/runs gpt-oss-20b:superdosh:openai --prompt "Why did the chicken cross the road?"
```

and it should _just work_ :tm:

PR #1211, opened 2025-08-20
Add ability to choose the openai driver for SUTs not deployed on openai

rogthefrog:
This allows you to use the existing OpenAI client for compatible SUTs running on different providers (not just openai's hardware).

# NEW: Use OpenAI client for non-OpenAI SUTs

## OpenAI SUTs are dynamic

Runing gpt models on OpenAI's hardware works as dynamic SUTs, so they don't need to be registered anymore:

`poetry run modelgauge run-sut --sut openai/gpt-4o:openai --prompt "why did the chicken cross the road?"`

### OpenAI-compatible SUTs on new providers

You can do the same thing for SUTs that support the OpenAI API Spec, including any SUTs running in vllm, like this:

Add the credentials to `secrets.toml`:
```
[demo]
api_key = "a real key"
```

Add a factory for it to `openai_sut_factory.py` (it must include `base_url`):

```python
class DemoOpenAICompatibleSUTFactory(OpenAIGenericSUTFactory):
    def __init__(self, raw_secrets, **kwargs):
        super().__init__(raw_secrets)
        self.provider = "demo"
        self.base_url = "https://example.net/v1/"

OPENAI_SUT_FACTORIES: dict = {"demo": DemoOpenAICompatibleSUTFactory}
```

and run this:

`poetry run modelgauge run-sut --sut some/sut:demo:openai --prompt "why did the chicken cross the road?"`

### OpenAI-compatible SUTs on known providers

If you have a known provider with credentials in the `secrets.toml` file, then it's even easier as you don't need to create a factory.

`poetry run modelgauge run-sut --sut some/sut:demo:openai;url=https://www.example.com --prompt "why did the chicken cross the road?"`

# Next Steps

* Find other registered SUTs compatible with the OpenAI client, and convert them to dynamic SUTs like this (maybe nvidia?).

[story](https://github.com/mlcommons/modelbench/issues/1132)

### DIFF
--- a/src/modelgauge/dynamic_sut_factory.py
+++ b/src/modelgauge/dynamic_sut_factory.py
@@ -39,7 +39,6 @@ class DynamicSUTFactory(ABC):
     def get_secrets(self) -> list[InjectSecret]:
         pass
 
-    # TODO: refactor this to use SUTDefinition instead of both the metadata and the kwargs
     @abstractmethod
     def make_sut(self, sut_definition: SUTDefinition) -> SUT:
         pass
--- a/src/modelgauge/dynamic_sut_factory.py
+++ b/src/modelgauge/dynamic_sut_factory.py
@@ -1,5 +1,4 @@
 from abc import ABC, abstractmethod
-from typing import Sequence
 
 from modelgauge.dependency_injection import inject_dependencies
 from modelgauge.secret_values import InjectSecret, RawSecrets
@@ -36,9 +35,8 @@ class DynamicSUTFactory(ABC):
         """Return the injected secrets as specified by `get_secrets`."""
         return inject_dependencies(self.get_secrets(), {}, secrets=self.raw_secrets)[0]
 
-    @staticmethod
     @abstractmethod
-    def get_secrets() -> list[InjectSecret]:
+    def get_secrets(self) -> list[InjectSecret]:
         pass
 
     @abstractmethod
--- a/src/modelgauge/dynamic_sut_factory.py
+++ b/src/modelgauge/dynamic_sut_factory.py
@@ -2,8 +2,8 @@ from abc import ABC, abstractmethod
 
 from modelgauge.dependency_injection import inject_dependencies
 from modelgauge.secret_values import InjectSecret, RawSecrets
-from modelgauge.dynamic_sut_metadata import DynamicSUTMetadata
 from modelgauge.sut import SUT
+from modelgauge.sut_definition import SUTDefinition
 
 
 class ModelNotSupportedError(Exception):
@@ -39,6 +39,7 @@ class DynamicSUTFactory(ABC):
     def get_secrets(self) -> list[InjectSecret]:
         pass
 
+    # TODO: refactor this to use SUTDefinition instead of both the metadata and the kwargs
     @abstractmethod
-    def make_sut(self, sut_metadata: DynamicSUTMetadata) -> SUT:
+    def make_sut(self, sut_definition: SUTDefinition) -> SUT:
         pass
--- a/src/modelgauge/sut_definition.py
+++ b/src/modelgauge/sut_definition.py
@@ -35,6 +35,8 @@ class SUTSpecification:
             "moderated": SUTSpecificationElement("moderated", "mod", bool),
             "driver_code_version": SUTSpecificationElement("driver_code_version", "dv", str),
             "date": SUTSpecificationElement("date", "dt", str),
+            "base_url": SUTSpecificationElement("base_url", "url", str),
+
         }
 
     def knows(self, field):
@@ -166,6 +168,7 @@ class SUTUIDGenerator:
         "top_k",
         "top_logprobs",
         "display_name",
+        "base_url" # for OpenAI-compatible SUTs
     )
     field_separator = RICH_UID_FIELD_SEPARATOR
     key_value_separator = "="
--- a/src/modelgauge/sut_definition.py
+++ b/src/modelgauge/sut_definition.py
@@ -36,7 +36,6 @@ class SUTSpecification:
             "driver_code_version": SUTSpecificationElement("driver_code_version", "dv", str),
             "date": SUTSpecificationElement("date", "dt", str),
             "base_url": SUTSpecificationElement("base_url", "url", str),
-
         }
 
     def knows(self, field):
@@ -168,7 +167,7 @@ class SUTUIDGenerator:
         "top_k",
         "top_logprobs",
         "display_name",
-        "base_url" # for OpenAI-compatible SUTs
+        "base_url",  # for OpenAI-compatible SUTs
     )
     field_separator = RICH_UID_FIELD_SEPARATOR
     key_value_separator = "="
--- a/src/modelgauge/sut_definition.py
+++ b/src/modelgauge/sut_definition.py
@@ -66,9 +66,8 @@ class SUTDefinition:
         if data:
             for k, v in data.items():
                 self.add(k, v)
-        if kwargs:
-            for k, v in kwargs.items():
-                self.add(k, v)
+        for k, v in kwargs.items():
+            self.add(k, v)
 
     @staticmethod
     def from_json(data: str):
--- a/src/modelgauge/sut_definition.py
+++ b/src/modelgauge/sut_definition.py
@@ -57,7 +57,7 @@ class SUTSpecification:
 class SUTDefinition:
     """The data in a SUT configuration file or JSON blob"""
 
-    def __init__(self, data=None):
+    def __init__(self, data=None, **kwargs):
         self._uid: str = ""
         self._dynamic_uid: str = ""
         self._frozen = False
@@ -65,6 +65,9 @@ class SUTDefinition:
         self.data = {}
         if data:
             for k, v in data.items():
+                self.add(k, v)
+        if kwargs:
+            for k, v in kwargs.items():
                 self.add(k, v)
 
     @staticmethod
@@ -152,6 +155,10 @@ class SUTDefinition:
             provider=self.data.get("provider", None),
             date=self.data.get("date", None),
         )
+
+    def external_model_name(self):
+        metadata = self.to_dynamic_sut_metadata()
+        return metadata.external_model_name()
 
 
 class SUTUIDGenerator:
--- a/src/modelgauge/sut_factory.py
+++ b/src/modelgauge/sut_factory.py
@@ -7,7 +7,7 @@ from modelgauge.secret_values import RawSecrets
 from modelgauge.sut import SUT
 from modelgauge.sut_registry import SUTS
 from modelgauge.suts.huggingface_sut_factory import HuggingFaceSUTFactory
-from modelgauge.suts.openai_sut_factory import OpenAISUTFactory
+from modelgauge.suts.openai_sut_factory import OpenAICompatibleSUTFactory
 from modelgauge.suts.together_sut_factory import TogetherSUTFactory
 
 
@@ -29,7 +29,7 @@ DYNAMIC_SUT_FACTORIES: dict = {
     "hf": HuggingFaceSUTFactory,
     "hfrelay": HuggingFaceSUTFactory,
     "huggingface": HuggingFaceSUTFactory,
-    "openai": OpenAISUTFactory,
+    "openai": OpenAICompatibleSUTFactory,
     "together": TogetherSUTFactory,
 }
 
--- a/src/modelgauge/sut_factory.py
+++ b/src/modelgauge/sut_factory.py
@@ -2,9 +2,9 @@ from enum import Enum
 
 from modelgauge.config import load_secrets_from_config
 from modelgauge.dynamic_sut_factory import DynamicSUTFactory, UnknownSUTMakerError
-from modelgauge.dynamic_sut_metadata import DynamicSUTMetadata
 from modelgauge.secret_values import RawSecrets
 from modelgauge.sut import SUT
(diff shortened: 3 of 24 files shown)


rogthefrog:
> Sorry, I'm not following the purpose of this work. Is this related to Vishal's comment in the story? I interpreted that as meaning that we should have a new dynamic SUT that would use the OpenAI client to contact arbitrary host/port combinations.

This will let us do:

secrets.toml:

```
[openai]
api_key="abcd"
organization="xyz"

[myvllm]
api_key="fgh"
base_url="http://myvllm.mydomain.org"

[google]
api_key="ijkl"
base_url="https://generativelanguage.googleapis.com/v1beta/openai/"
```

invocation:

```bash
modelgauge run-sut -s openai/gpt-4o:openai # run gpt-4o on openai's hardware using the openai client
modelgauge run-sut -s some/model:openai:myvllm # run some/model on the machine at base_url using the openai client
modelgauge sun-sut -s google/gemini-2.5-flash:openai:google # talk to gemini on google's hardware using the openai client 
```

src/modelgauge/sut_factory.py
```diff
@@ -21,17 +21,16 @@ class SUTType(Enum):
...
 # Maps a string to the module and factory function in that module
 # that can be used to create a dynamic sut
 DYNAMIC_SUT_FACTORIES: dict = {
-    "proxied": {"hfrelay": HuggingFaceSUTFactory},
-    "direct": {
-        "openai": OpenAISUTFactory,
-        "together": TogetherSUTFactory,
-        "huggingface": HuggingFaceSUTFactory,
-        "hf": HuggingFaceSUTFactory,
-    },
```
superdosh, line 34: Q: is the flattening of this dict tangential to the main part of the PR? Happy to have it here, but want to make sure I understand.

src/modelgauge/suts/openai_client.py
```diff
@@ -101,15 +83,39 @@ class OpenAIChat(PromptResponseSUT[OpenAIChatRequest, ChatCompletion]):
...
+            assert not self.base_url
+
+    def _load_client(self) -> OpenAI | None:
+        if self.client:
+            return self.client
+
+        if self.base_url:
+            return OpenAI(api_key=self.api_key, base_url=self.base_url, max_retries=7)
+        elif self.organization:
+            return OpenAI(api_key=self.api_key, organization=self.organization, max_retries=7)
```
superdosh, line 116: What if we have neither? Is it required by `openai.OpenAI` ?

e -- sorry, I'm more referring to api_key. Is it necessary?

src/modelgauge/sut_factory.py
```diff
@@ -21,17 +21,16 @@ class SUTType(Enum):
     UNKNOWN = "unknown"
 
 
-# TODO: Auto-collect. Maybe make "driver" keys a constant in each factory module.
+# TODO: Auto-collect?
+# Make sure the factory module includes the matching key as a constant.
 # Maps a string to the module and factory function in that module
 # that can be used to create a dynamic sut
 DYNAMIC_SUT_FACTORIES: dict = {
```
bkorycki, line 28: Nice

src/modelgauge/suts/openai_client.py
```diff
@@ -101,15 +83,39 @@ class OpenAIChat(PromptResponseSUT[OpenAIChatRequest, ChatCompletion]):
...
 
-    def __init__(self, uid: str, model: str, api_key: OpenAIApiKey, org_id: OpenAIOrgId):
+    def __init__(
+        self,
+        uid: str,
+        model: str,
+        api_key: Optional[OpenAICompatibleApiKey] = None,
+        organization: Optional[OpenAICompatibleOrganization] = None,
+        base_url: Optional[OpenAICompatibleBaseURL] = None,
+        client: Optional[OpenAI] = None,
```
bkorycki, line 91: Why is it necessary to support passing in a client now?

src/modelgauge/sut_factory.py
rogthefrog: Yes, tangential; it came up in a conversation with Barbara. I had originally made a distinction between proxied (us -> hfrelay -> actual provider) and direct (us -> provider) because that's the flow, but in actuality that ended up not mattering.

src/modelgauge/suts/openai_client.py
rogthefrog: So far, it has been. If it's ever blank, e.g. when we self-host vllm, then we can deal with that then, though I'd suggest always having a key as a basic precaution.

Is there an existing or planned use case for keyless access to vllm?

src/modelgauge/suts/openai_client.py
rogthefrog: It's not necessary so much as convenient for a couple of reasons.

* There's an independent reason to create a client [in the OpenAISUTFactory](https://github.com/mlcommons/modelbench/blob/ba041b4d01f3f0a1a1577dbf21a2e0004005e5d8/src/modelgauge/suts/openai_sut_factory.py#L54), which checks if a model is available. Since we have the client ready to go with the secrets, it simplilfies the call to this constructor
* The logic to create the OpenAI client object is a little complicated as it's different between [a SUT hosted by OpenAI](https://github.com/mlcommons/modelbench/blob/ba041b4d01f3f0a1a1577dbf21a2e0004005e5d8/src/modelgauge/suts/openai_sut_factory.py#L43) and a SUT with [an OpenAI-compatible API hosted elsewhere](https://github.com/mlcommons/modelbench/blob/ba041b4d01f3f0a1a1577dbf21a2e0004005e5d8/src/modelgauge/suts/openai_sut_factory.py#L76). In a future refactoring, I think it would be good to abstract that logic away from the `OpenAIChat` class and have the `OpenAI` client logic implemented only in one place. I didn't do it in this PR because there are other instances of `OpenAIChat` constructions that are out of scope of this PR IMO.

src/modelgauge/suts/openai_client.py
superdosh: Nah, this is fine.

superdosh (approved):

wpietri:
This is interesting, but I'm still not getting the use case. What persons will use this, and what goal do they have when they do?

rogthefrog:
> This is interesting, but I'm still not getting the use case. What persons will use this, and what goal do they have when they do?

We should talk about this live, because I'm not sure where the disconnect is.

> This introduces a strange linkage between the local secrets file and the CLI.

It's always been linked. You can't run anything without a secrets file including the secrets for the SUT you want.

> If I imagine a one-off experimental use case, it seems too heavy, requiring both things be changed each time. 

That's already the case--currently there's no way to run a SUT without credentials in the secrets file; if we want that, it can be done, but that wasn't the goal or scope of this card. If we want to support something like "accept config and credentials for arbitrary SUTs in the command like, that's easy enough, but not this card.

> If I imagine a production use case, it seems chaotic to try to keep various files in sync.

This doesn't require any sync other than adding secrets for your SUT to whatever secrets management we use, which we already do when we add a SUT with a new provider or plugin.

> The secrets file is becoming more a config file, which seems like a problem to me.

I think the secrets file is problematic in general, and the way secrets are managed is excessively complicated, but that's a bigger job and not in scope of this. I'd love to simplify it.

> This is pushing complexity toward the user. 

How so? There's no difference between this and any SUT. 

1) put the keys in secrets (including the base URL, which takes the place of the org ID) 
2) run your job with `my/model:openai:provider` just like all the other SUT. 

Any current or future SUT that supports the OpenAI specification can use this. All the gemini SUTs do, for example, so we could conceivably remove the VertexAI-specific plugin and use this instead, like `google/gemini-something:openai:google`.

> If the user really needs the full power of the combinatorial space, that can be fine.

We (the eng team) do, don't we? We want to be able to run any new SUT with a minimum of work. This allows us to run any SUT that supports the OpenAI API without writing custom code. It's exactly the same as what we did with together and huggingface.

> But in running production benchmarks, we are generally looking to eliminate possibility, not create it. Especially so if we want to hand off most benchmark-running to businesspeople.

From the card:

image

This is what that does, with the only difference that the `base_url` is in `secrets.toml`. If we want to support passing in the base URL as a CLI argument, that's easy to do; if we want to support SUTs with no api_key, we can also do that, but Vishal says that's a hypothetical use case at the moment.

wpietri:
> > This introduces a strange linkage between the local secrets file and the CLI.
> 
> It's always been linked. You can't run anything without a secrets file including the secrets for the SUT you want.

Well then let's say it expands the linkage. The part that's troubling me is that the definition of a SUT is now spread across three places: our code, the command-line arguments, and a secrets file that's not a secrets file. That could be fine, if awkward, for specific experimental use cases. I think it's taking us in the wrong direction for high integrity operations for production benchmarks.

> > This is pushing complexity toward the user.
> 
> How so? There's no difference between this and any SUT.
> 
> 1. put the keys in secrets (including the base URL, which takes the place of the org ID)
> 2. run your job with `my/model:openai:provider` just like all the other SUT.

Yes, the difference is that there's a URL in there, along with an arbitrary name. Org ids are not arbitrary, and the relationship between the org id and what happens is in version control. Putting key configuration information into an ephemeral file is a huge problem for reproducibility. And now normal users have to manage stuff they generally shouldn't be messing with. It also means one can't just share a CLI string with somebody and have it work for them; two things that must travel together are in quite different places.


> Any current or future SUT that supports the OpenAI specification can use this. All the gemini SUTs do, for example, so we could conceivably remove the VertexAI-specific plugin and use this instead, like `google/gemini-something:openai:google`.
>  [...]
> We (the eng team) do, don't we? We want to be able to run any new SUT with a minimum of work. This allows us to run any SUT that supports the OpenAI API without writing custom code. It's exactly the same as what we did with together and huggingface.

This is an interesting thing, but I don't think it's in scope for the card. This card is just about finishing the cutover to dynamic SUTs for existing providers. Adding the ability to do more arbitrary things using the OpenAI client is a nice-to-have for experimental use cases. 


> That's what this does, with the only difference that the `base_url` is in `secrets.toml`. If we want to support passing in the base URL as a CLI argument, that's easy to do; if we want to support SUTs with no api_key, we can also do that, but Vishal says that's a hypothetical use case at the moment.

Passing the base URL as a CLI argument would be one option. Another option would be putting the URL into the SUT uid itself. Indeed, it's going to have to go in the SUT's uid object, because the URL is going to be part of what uniquely identifies the SUT. If it's untenable to put it in the short text form of the uid, it still needs to end up in the object representation.

rogthefrog:
I don't understand this misunderstanding.

> This card is just about finishing the cutover to dynamic SUTs for existing providers. Adding the ability to do more arbitrary things using the OpenAI client is a nice-to-have for experimental use cases.

This makes it possible to cut over to dynamic SUTs for all gemini models.

> along with an arbitrary name. Org ids are not arbitrary, 

The "organization" string *is* the "org_id." It's called "organization" in the OpenAI client, so I named it that instead of "org_id" to match the API spec.

> Putting key configuration information into an ephemeral file is a huge problem for reproducibility. 

This changes nothing about that.

> And now normal users have to manage stuff they generally shouldn't be messing with. It also means one can't just share a CLI string with somebody and have it work for them

You already can't do that if they don't have the right credentials.

src/modelgauge/dynamic_sut_factory.py
```diff
@@ -36,11 +35,11 @@ def injected_secrets(self):
         """Return the injected secrets as specified by `get_secrets`."""
         return inject_dependencies(self.get_secrets(), {}, secrets=self.raw_secrets)[0]
 
-    @staticmethod
     @abstractmethod
-    def get_secrets() -> list[InjectSecret]:
+    def get_secrets(self) -> list[InjectSecret]:
         pass
 
+    # TODO: refactor this to use SUTDefinition instead of both the metadata and the kwargs
```
superdosh: Don't need this TODO anymore?

src/modelgauge/sut_definition.py
```diff
@@ -65,6 +66,9 @@ def __init__(self, data=None):
         if data:
             for k, v in data.items():
                 self.add(k, v)
+        if kwargs:
```
superdosh: Q: do you need `if kwargs`?

src/modelgauge/suts/openai_sut_factory.py
```diff
@@ -1,42 +1,101 @@
...
+NUM_RETRIES = 7
+
 
+class OpenAICompatibleSUTFactory(DynamicSUTFactory):
 
-class OpenAISUTFactory(DynamicSUTFactory):
     def __init__(self, raw_secrets: RawSecrets):
         super().__init__(raw_secrets)
-        self._client: OpenAI | None = None  # Lazy load.
+        self.provider: str = "openai"  # must match name of section (scope) in secrets.toml
```
superdosh: I think I'm confused here. Why is this hardcoded? Wouldn't this be variable depending on the SUT?

src/modelgauge/sut_definition.py
rogthefrog: That's a good question.

src/modelgauge/suts/openai_sut_factory.py
rogthefrog: You're right, it is set by child SUTs. I just don't like uninitialized variables. 

That said, making it `None` would tell us if a child SUT forgot to set it to an actual value (by blowing up at runtime), so I'll change it to None.

superdosh (approved):

wpietri (approved):

wpietri:
> Good catch, now it has, and has a unit test too.

Fab! Ship it, I say!

PR #1215, opened 2025-08-21
Duplicate prompt/response pairs with unique prompt ids

superdosh:
Ran into an issue where we had a prompt dataset with unique prompt ids, but a duplicate prompt (which produced a duplicate response).

The way our caching works was the key is based only on the request/response text, but the value is the "whole" row, including the prompt uid.

If we had a file like:
```csv
prompt_uid,prompt,response
ppp,prompt0,response0
qqq,prompt0,response0
...
```
We'd end up with output like:
```csv
prompt_uid,prompt,response,annotations...
ppp,prompt0,response0,...
ppp,prompt0,response0,...
```

We'll probably want to discuss a larger fix, but this fix minimally handles the issue above.

Output of the new test before the changes applied:
```
FAILED tests/modelgauge_tests/test_pipeline_runner.py::TestAnnotatorRunner::test_cache_responses - AssertionError: assert {'p0', 'p1', 'p2', 'q0'} == {'p0', 'p1', 'p2'}
```

Tests pass now.

### DIFF
--- a/src/modelgauge/annotation_pipeline.py
+++ b/src/modelgauge/annotation_pipeline.py
@@ -51,14 +51,7 @@ class AnnotatorWorkers(CachingPipe):
         request = annotator.translate_request(sut_interaction.prompt, sut_interaction.response)
         if isinstance(request, BaseModel):
             request = request.model_dump_json()
-        return (request, annotator_uid)
-
-    def handle_item(self, item):
-        result = super().handle_item(item)
-        sut_interaction, annotator_uid = item
-        return AnnotatedSUTInteraction(
-            annotator_uid=annotator_uid, annotation=result.annotation, sut_interaction=sut_interaction
-        )
+        return (sut_interaction.prompt.source_id, request, annotator_uid)
 
     def handle_uncached_item(self, item):
         sut_interaction, annotator_uid = item
--- a/src/modelgauge/annotation_pipeline.py
+++ b/src/modelgauge/annotation_pipeline.py
@@ -1,26 +1,18 @@
-import csv
-import jsonlines
 import logging
 import time
-from abc import abstractmethod, ABCMeta
 from collections import defaultdict
 from pydantic import BaseModel
-from typing import Iterable
 
 from modelgauge.annotation import Annotation
 from modelgauge.annotator import Annotator
 from modelgauge.annotator_set import AnnotatorSet
 from modelgauge.dataset import AnnotationDataset, PromptResponseDataset
-from modelgauge.data_schema import DEFAULT_PROMPT_RESPONSE_SCHEMA, PromptResponseSchema
 from modelgauge.pipeline import CachingPipe, Pipe, Sink, Source
-from modelgauge.prompt import TextPrompt
 from modelgauge.single_turn_prompt_response import (
     AnnotatedSUTInteraction,
     SUTResponseAnnotations,
     SUTInteraction,
-    TestItem,
 )
-from modelgauge.sut import PromptResponseSUT, SUTResponse
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


tests/modelgauge_tests/test_pipeline_runner.py
```diff
@@ -65,6 +65,22 @@ def prompt_responses_file(tmp_path_factory):
...
+    """Sample file with 3 prompts + responses from 2 SUTs for testing. Also include duplicate prompt/response, with unique prompt id."""
+    file = tmp_path_factory.mktemp("data") / "prompt-responses.csv"
+    with open(file, "w") as f:
+        text = f"{PROMPT_RESPONSE_SCHEMA.prompt_uid},{PROMPT_RESPONSE_SCHEMA.prompt_text},{PROMPT_RESPONSE_SCHEMA.sut_uid},{PROMPT_RESPONSE_SCHEMA.sut_response}\n"
+        for i in range(NUM_PROMPTS):
+            text += f"p{i},Prompt {i},sut1,Response {i}\n"
+            text += f"p{i},Prompt {i},sut2,Response {i}\n"
+        # add a duplicate with unique prompt ids
+        text += f"q0,Prompt 0,sut1,Response 0\n"
+        print(text)
```
rogthefrog: Is this needed?

src/modelgauge/annotation_pipeline.py
```diff
@@ -1,26 +1,18 @@
-import csv
```
rogthefrog, line 1: Are all these removals just a cleanup of unused imports?

src/modelgauge/annotation_pipeline.py
```diff
@@ -61,6 +53,13 @@ def key(self, item):
             request = request.model_dump_json()
         return (request, annotator_uid)
 
+    def handle_item(self, item):
+        result = super().handle_item(item)
+        sut_interaction, annotator_uid = item
```
rogthefrog: I'm not seeing where the deduping is happening (I don't know the pipeline codebase). Can you explain?

src/modelgauge/annotation_pipeline.py
superdosh: Yes. The import changes are not meaningful. I can drop if needed!

tests/modelgauge_tests/test_pipeline_runner.py
superdosh: Oops, thanks!

src/modelgauge/annotation_pipeline.py
superdosh: So the cache key is based ONLY on the `(request, annotator_id)`, but the value is the whole `AnnotatedSUTINteraction`. 

But the `SUTInteraction` has a bunch of other things like the prompt_uid which are unique to the row. Does that make sense?

src/modelgauge/annotation_pipeline.py
rogthefrog: No need to change, I think it's great. I just wanted to confirm!

src/modelgauge/annotation_pipeline.py
rogthefrog: I think so :)

wpietri: I'm also finding this mysterious. Shouldn't we just add the prompt id to the cache key?

superdosh: Per discussion, updated to include the prompt_uid in the cache key!

tests/modelgauge_tests/test_annotation_pipeline.py
```diff
@@ -143,9 +143,9 @@ def test_annotator_worker_unique_responses(annotators, tmp_path):
     w.handle_item((make_sut_interaction("", "", "", "response 2"), "annotator_pydantic"))
     assert annotators["annotator_pydantic"].annotate_calls == 2
 
-    # Non-response SUT interaction attributes do not affect the cache key.
+    # New prompt id does affect the cache key.
```
superdosh, line 146: Call out here that we did specifically test that new prompt id did NOT result in a cache miss. 🤷🏼 

@wpietri going to still go with what we discussed (new prompt id means not cached).

PR #1218, opened 2025-08-25
small monitoring (and other) fixes

dhosterman:
Add some more noop operations to enxure code works in non-monitored situations, testing, etc. Also, fix small type error.

### DIFF
--- a/src/modelgauge/monitoring.py
+++ b/src/modelgauge/monitoring.py
@@ -1,5 +1,6 @@
 import os
 import socket
+import time
 
 
 class NoOpMetric:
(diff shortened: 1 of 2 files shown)


src/modelgauge/monitoring.py
```diff
@@ -16,11 +17,27 @@ def observe(self, *args, **kwargs):
         return self
 
     def time(self):
-        return self
+        class Timer:
```
wpietri, line 20: We already have [a similar class](https://github.com/mlcommons/modelbench/blob/668c93b5fa4cba4efc8d5a8a525480d13652bd70/src/modelbench/benchmark_runner_items.py#L104). Seems like it should be in a common place.

wpietri (approved):

src/modelgauge/monitoring.py
dhosterman: The class I'm using here really doesn't need to do anything, it is mostly just a mock with a similar interface to the Prometheus timer so that it can be used in the NoOpMetric class without throwing errors when we use the context manager.

I'd be happy to make it do even less and rename it so it is clear that it is not supposed to do anything at all.

Alternatively, we could implement some other way to make prometheus optional that doesn't rely on a NoOpMetric. I'd be in favor of that if it was less irritating than the current method. Any thoughts?

wpietri: Ah, that makes sense. In that case I'm happy to have it like this.

What's driving the need for no-op metrics? I'm a fan of the null object pattern, but I was thinking we could just have prometheus metrics set up whether or not they got used.

dhosterman: It has been a while, but I'm pretty sure I was either trying to prevent errors being thrown or unnecessary resource consumption or both.

PR #1224, opened 2025-08-26
make tests less fragile

dhosterman:
attempting to remove the need for any config files or workarounds from testing


bkorycki (approved):
noice

PR #1227, opened 2025-08-26
Security prompt sets

bkorycki:
Prepare infrastructure for adding more security prompt sets.

Should not require any changes to modellab because the new --prompt_set and --locale options have defaults.


rogthefrog (approved):
Looks good!

PR #1231, opened 2025-08-28
Add traceback to error logging

superdosh:
Ran into an issue where there was an exception being raised in ensemble code developed in modelplane. `debug` mode was on, but because `item` has so much text, and `str(e)` is basically very short, it wasn't possible to really debug this.

This will add the traceback which should make debugging easier!


dhosterman (approved):
👍

superdosh:
@wpietri that pattern makes sense! Will merge this for now, but I think my next one should solve the specific problem in a nicer way (failing fast with details).

PR #1239, opened 2025-08-28
Implement readiness checks for pipelines in modelgauge

superdosh:
Also refactor some modelbench stuff to make this simpler


src/modelgauge/sut.py
```diff
@@ -101,13 +103,23 @@ def __init__(self, uid: str):
         self.initialization_record: InitializationRecord
 
 
-class PromptResponseSUT(SUT, ABC, Generic[RequestType, ResponseType]):
+_READINESS_CHECK_TEXT_PROMPT = TextPrompt(text="Why did the chicken cross the road?")
+_READINESS_CHECK_SUT_OPTIONS = SUTOptions(max_tokens=1000)
```
bkorycki: Wow I didn't realize we had this set to 1000 tokens! Do you mind changing it to like 10? That should speed up the initial checks:)

src/modelbench/benchmark_runner.py
```diff
@@ -482,54 +479,17 @@ def _check_ready_to_run(self):
...
-        annotators = set(a for l in run.test_annotators.values() for a in l)
-        with ThreadPool(len(annotators)) as pool:
-            annotators_worked = pool.map(check_annotator, annotators)
-            if not all(annotators_worked):
-                raise RuntimeError(
-                    f"Not all annotators are ready to go. Status: {dict(zip([a.uid for a in annotators], annotators_worked))}"
-                )
+        sut_status = run.sut.is_ready()
+        if not sut_status.is_ready:
+            raise RuntimeError(f"SUT is not ready. Status: {sut_status.response}")
```
bkorycki: Can we also print the exception trace if there is one?

bkorycki (changes requested):
Love it!

src/modelbench/benchmark_runner.py
wpietri: Yeah, I think this should either use logging.exception to print the stack trace or  do `raise from` to chain the exceptions, with something higher-level catching and logging the problem.

superdosh: Makes sense; will update here and in the other similar lines.

src/modelgauge/sut.py
superdosh: Ohh, that makes sense! Yah, hopefully will speed things up!

superdosh: Going to do 20, since we use 7 or so on the prompt? Or is it just for the response?

src/modelbench/benchmark_runner.py
superdosh: I think this ends up being the only spot that makes sense to update, since in the other cases, we have multiple check responses, so it doesn't really make sense to raise just one of them.

src/modelgauge/sut.py
bkorycki: 20 is fine! I think it's usually just the response but it also depends on the model.

src/modelbench/benchmark_runner.py
bkorycki: Could we raise all of them? :)

src/modelgauge/sut.py
```diff
@@ -101,13 +103,23 @@ def __init__(self, uid: str):
         self.initialization_record: InitializationRecord
 
 
-class PromptResponseSUT(SUT, ABC, Generic[RequestType, ResponseType]):
+_READINESS_CHECK_TEXT_PROMPT = TextPrompt(text="Why did the chicken cross the road?")
+_READINESS_CHECK_SUT_OPTIONS = SUTOptions(max_tokens=20)
+
+
+class PromptResponseSUT(SUT, Generic[RequestType, ResponseType], Readyable):
```
rogthefrog, line 110: The `Readyable` approach is 👨🏻‍🍳

rogthefrog (approved):
Awesome.

src/modelgauge/sut.py
superdosh: Thanks! TBH, I don't love the name, but I spent too long trying to come up with something better and gave up.

src/modelbench/benchmark_runner.py
superdosh: Let me know if you think this version's better @bkorycki !

bkorycki: looks great!

superdosh:
(Just a rebase to deconflict with main. Will merge once tests pass.)

PR #1243, opened 2025-08-29
Fix EnsembleRunner

superdosh:
My last [PR](https://github.com/mlcommons/modelbench/pull/1239) introduced a bug to the EnsembleRunner where we weren't capturing the correct part of the response in the readiness check. Also updated the tests so they'd have caught the error.

### DIFF
--- a/tests/modelgauge_tests/test_pipeline_runner.py
+++ b/tests/modelgauge_tests/test_pipeline_runner.py
@@ -27,9 +27,14 @@ from modelgauge.prompt_pipeline import (
     PromptSink,
 )
 from modelgauge.sut import SUTOptions
-from modelgauge_tests.fake_annotator import BadAnnotator, FakeAnnotator
+from modelgauge_tests.fake_annotator import BadAnnotator, FakeAnnotator, FakeSafetyAnnotator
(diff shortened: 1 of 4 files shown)


tests/modelgauge_tests/test_pipeline_runner.py
```diff
@@ -266,6 +280,16 @@ def test_unready_ensemble(self, tmp_path, prompts_dataset, annotators):
                 output_dir=tmp_path,
             )
 
+    def test_ready_ensemble(self, tmp_path, prompts_dataset, safety_annotators):
```
wpietri, line 283: This is a very mysterious test. I'm wondering if running the readiness checks in __init__ is the best way to do this. Perhaps the `ensure_ready` should be directly callable and normally called by the things that call `run`? But feel free to merge this to fix your thing.

wpietri (approved):

tests/modelgauge_tests/test_pipeline_runner.py
superdosh: Yah, the test is kind of weird for sure. I didn't love the idea of having a thing that would have to be called. Potentially it go at the beginning of `runner.run` ? Either way, will merge this for now!

PR #1244, opened 2025-08-29
security annotator ready check

bkorycki:
Security annotators require a context item and therefore special checks.

### DIFF
--- a/src/modelgauge/annotator.py
+++ b/src/modelgauge/annotator.py
@@ -1,10 +1,10 @@
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 from typing import Generic, TypeVar
 
 from pydantic import BaseModel
 
 from modelgauge.ready import Readyable, ReadyResponse
-from modelgauge.single_turn_prompt_response import TestItem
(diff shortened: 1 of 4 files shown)


src/modelgauge/annotator.py
```diff
@@ -50,3 +50,18 @@ def annotate(self, annotation_request):
...
+
+
+class SecurityAnnotator(CompletionAnnotator[AnnotationType], ABC):
+    """Security annotators require a seed prompt in the context."""
+
+    def run_readiness_check(self) -> ReadyResponse:
+        test_item = TestItem(
+            prompt=_READINESS_CHECK_TEST_ITEM.prompt,
+            source_id=_READINESS_CHECK_TEST_ITEM.source_id,
+            context=SecurityContext(seed_prompt="This is sent to annotator"),
```
superdosh, line 62: Sorry, I actually had this an the original `_READINESS_CHECK_TEST_ITEM` but it caused a circular import, so I removed.

bkorycki: no worries!

PR #1247, opened 2025-09-02
add huggingface-specific retry logic

rogthefrog:
Retry huggingface completion requests in case of server-side errors. This is Huggingface-specific, so I put the retry in the HF client itself rather than the code that calls `some_object.evaluate()`. Putting the retry logic higher up would break encapsulation and complicate the calling code.

### DIFF
--- a/src/modelgauge/suts/huggingface_chat_completion.py
+++ b/src/modelgauge/suts/huggingface_chat_completion.py
@@ -5,7 +5,7 @@ from typing import Dict, List, Optional
 from huggingface_hub import get_inference_endpoint, InferenceClient, InferenceEndpointStatus  # type: ignore
 from huggingface_hub.utils import HfHubHTTPError  # type: ignore
 from pydantic import BaseModel
-
+from tenacity import retry, TryAgain, stop_after_attempt, wait_random_exponential
 from modelgauge.auth.huggingface_inference_token import HuggingFaceInferenceToken
 from modelgauge.prompt import TextPrompt, ChatPrompt
 from modelgauge.secret_values import InjectSecret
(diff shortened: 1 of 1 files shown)


rogthefrog:
@bkorycki @wpietri @dhosterman do you have any suggestions for testing this automatically?

wpietri (approved):
Looks good!

wpietri:
> do you have any suggestions for testing this automatically?

I think you don't have to test the tenacity stuff; a unit test can just worry about the code at hand. So I'd just patch and throw the interesting exceptions to make sure they do and don't emerge as intended.

tests/modelgauge_tests/sut_tests/test_huggingface_chat_completion.py
```diff
@@ -2,6 +2,7 @@
 from unittest.mock import Mock, patch
 
 import pytest
+from pytest import MonkeyPatch
```
dhosterman: Do you need this import?

src/modelgauge/suts/huggingface_chat_completion.py
```diff
@@ -56,21 +58,30 @@ def _create_client(self) -> InferenceClient:
...
+            return HuggingFaceChatCompletionOutput(
+                choices=[asdict(choice) for choice in response.choices],
+                created=response.created,
+                id=response.id,
+                model=response.model,
+                system_fingerprint=response.system_fingerprint,
+                usage=asdict(response.usage),
+            )
+        except HfHubHTTPError as hf_error:
+            if hf_error.response.status_code >= 500:
```
bkorycki: What does raising TryAgain do? And why are 5xx errors treated differently?

tests/modelgauge_tests/sut_tests/test_huggingface_chat_completion.py
rogthefrog: No longer!

src/modelgauge/suts/huggingface_chat_completion.py
rogthefrog: Raising `TryAgain` is a way to trigger a tenacity retry. So in this case, if there's an exception during the call, we give up and raise it (and don't retry at all), **unless** it's a HfHubHTTPError with a status_code >= 500.

The status_code >= 500 thing is what triggered this change:

image

I also added 429 to the list of error codes to retry based on this, which happened while testing this PR:

image

PR #1249, opened 2025-09-03
Create a basic ModelShip SUT

wpietri:
Currently doesn't use the OpenAI-compatible stuff, but it should eventually.

### DIFF
--- a/src/modelgauge/suts/modelship_sut.py
+++ b/src/modelgauge/suts/modelship_sut.py
@@ -1,0 +1,25 @@
+from modelgauge.dynamic_sut_factory import DynamicSUTFactory
+from modelgauge.secret_values import InjectSecret, RequiredSecret, SecretDescription
+from modelgauge.sut import SUT
+from modelgauge.sut_definition import SUTDefinition
+from modelgauge.suts.openai_client import OpenAIChat
+
+
+class ModelShipSecret(RequiredSecret):
+    provider: str = "modelship"
+
+    @classmethod
+    def description(cls) -> SecretDescription:
+        return SecretDescription(scope=cls.provider, key="api_key", instructions="Ask around")
+
+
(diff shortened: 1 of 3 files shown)


src/modelgauge/suts/modelship_sut.py
```diff
@@ -0,0 +1,25 @@
...
+from modelgauge.sut_definition import SUTDefinition
+from modelgauge.suts.openai_client import OpenAIChat
+
+
+class ModelShipSecret(RequiredSecret):
+    provider: str = "modelship"
+
+    @classmethod
+    def description(cls) -> SecretDescription:
+        return SecretDescription(scope=cls.provider, key="api_key", instructions="Ask around")
```
superdosh, line 13: In this case, might the instructions more aptly be "You should know" since in theory you set it up? Or how will the ModelShip secret be set?

wpietri: We should all talk about this. For now the code isn't checked in anywhere. Previously we've just used Google Secrets, but this code will be running on MLC2, so accessing Google Secrets isn't as easy. My best idea so far is that we use ansible to deploy to MLC2, and during that process ansible pulls the secret from Google Secrets and sets it in a config file or something.

superdosh: Yah would be good to chat about the flow as a group. I'm not 100% on it. Is the goal to have something like if you kick off a benchmark with the modelship provider, that it will:
1. Deploy it on mlc2
2. Run the benchmark against mlc2

I was assuming both, but maybe it's actually only 2? If both, then the client is setting up the secret, so it should know it?

wpietri: I was thinking we'd leave the modelship agent running on MLC2 all the time, so only 2.

PR #1255, opened 2025-09-04
Generic OpenAI-compatible SUT for modelship

rogthefrog:
Creating an OpenAI-compatible SUT works roughly the same as other factory-based SUTs, but the code and string-to-factory mapping are in a different place.

### DIFF
--- a/src/modelgauge/sut_factory.py
+++ b/src/modelgauge/sut_factory.py
@@ -7,7 +7,6 @@ from modelgauge.sut import SUT
 from modelgauge.sut_definition import SUTDefinition, SUTUIDGenerator
 from modelgauge.sut_registry import SUTS
 from modelgauge.suts.huggingface_sut_factory import HuggingFaceSUTFactory
-
 from modelgauge.suts.openai_sut_factory import OpenAICompatibleSUTFactory
 from modelgauge.suts.together_sut_factory import TogetherSUTFactory
 


src/modelgauge/suts/README.md
```diff
@@ -3,3 +3,93 @@
...
+* openai-compatible clients
+
+### OpenAI-Compatible Dynamic SUTs
+
+If your SUT supports the OpenAI API, you can call it one of two ways.
+
+#### No Code
+
+```bash
+poetry run modelgauge run-sut --sut "maker/model:mysut:openai;url=https://example.com/v1/" --prompt "why did the chicken cross the road?"
```
superdosh, line 32: Should this be `base_url` instead of `url`? Otherwise, not sure how the magic is happening?

src/modelgauge/suts/openai_sut_factory.py
```diff
@@ -112,4 +112,11 @@ def __init__(self, raw_secrets, **kwargs):
         # api_key = "abcd"
 
 
-OPENAI_SUT_FACTORIES: dict = {"demo": DemoOpenAICompatibleSUTFactory}
```
superdosh, line 115: Is the reason it didn't work when @wpietri first tried it that it was in a separate file and so wasn't imported into `OPENAI_SUT_FACTORIES`? Otherwise, not sure what you're doing differently?

wpietri: Yeah, I'd swear I tried this, addition to the factories constant and all.

src/modelgauge/suts/README.md
rogthefrog: The parameters in the string version of the rich SUT UID are shortened. E.g. max_tokens is m, base_url is url, etc. This is done in [the sut definition code](https://github.com/mlcommons/modelbench/blob/e09639f6fa55befc146517f7b11f2188bbec78a1/src/modelgauge/sut_definition.py#L272) and was implemented for the requirement that rich SUT UIDs not be too long.

It's easy to change.

superdosh (approved):

PR #1257, opened 2025-09-04
First reasoning sut + security handles reasoning suts

bkorycki:
This PR implements changes that allow you to benchmark reasoning SUTs that discard their reasoning output.
- A new SUT option called `max_total_output_tokens` for SUTs that output more than just response text. Non-reasoning SUTs still just use the old `max_tokens` value.
- Any reasoning SUTs are responsible for truncating the response text down to `max_tokens`.
- The first reasoning SUT: `TogetherThinkingSUT`. Technically this can be used for any reasoning SUT on Together that returns it's thinking output enclosed in `<think></think> `tags before the actual response. But I'm not actually aware of any reasoning SUTs that behave this way outside of deep seek R1.


src/modelgauge/sut.py
```diff
@@ -118,7 +129,7 @@ def run_readiness_check(self) -> ReadyResponse:
         raw_request = self.translate_text_prompt(_READINESS_CHECK_TEXT_PROMPT, options=_READINESS_CHECK_SUT_OPTIONS)
         raw_response = self.evaluate(raw_request)
         response = self.translate_response(raw_request, raw_response)
-        return ReadyResponse(is_ready=bool(response.text), response=response)
+        return ReadyResponse(is_ready=response.text is not None, response=response)
```
bkorycki, line 132: I had to change this to accommodate the thinking SUT, which returns an empty string if max_tokens is too low to generate non-thinking text.

src/modelgauge/sut.py
```diff
@@ -54,6 +57,14 @@ class SUTOptions(BaseModel):
     """If present, will request the log probabilities for this
     many of the top tokens at each token position."""
 
+    @model_validator(mode="after")
```
wpietri, line 60: Good catch.

src/modelgauge/suts/together_client.py
```diff
@@ -266,6 +272,83 @@ def translate_response(self, request: TogetherChatRequest, response: TogetherCha
...
+            # no closing tag: everything is thinking text
+            return ""
+
+        response = text[think_close + len("</think>") :].strip()
+
+        if max_tokens is None:
+            return response
+
+        # Truncate response text to max tokens
+        if self.tokenizer is None:
```
wpietri, line 343: I'd suggest making tokenizer a property, and let just that method handle the lazy loading.

You might also extract a truncate method, either to this class or to a wrapper for AutoTokenizer if you think we'll want to have more code closely relating to it.

rogthefrog (approved):
👍🏻I added this new max token param to the rich SUT UID parser here:

https://github.com/mlcommons/modelbench/pull/1260

bkorycki:
@rogthefrog The max tokens are set by the test not the SUT! So it's not a part of the SUT UID.

wpietri (approved):

src/modelgauge/suts/together_client.py
bkorycki: I changed the tokenizer to be a property!

I think I will extract the reasoning parsing logic and truncation in a future PR when we start adding more reasoning SUTs.

PR #1261, opened 2025-09-05
Speed up tests

wpietri:
Our tests had gotten up to 3 minutes of runtime. This brings them back down to around 10 seconds, albeit in dubious ways. Are they too dubious? Let me know.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,9 +1,11 @@
 import copy
 import importlib
 import sys
+import time
 from collections import defaultdict
 from datetime import datetime, timedelta, timezone
 from typing import Dict
+from unittest import mock
 
 import pytest
 
@@ -71,3 +73,12 @@ def pytest_sessionstart(session):
     mg_config.load_secrets_from_config = new_func
     if "modelgauge.sut_factory" in sys.modules:
         importlib.reload(sys.modules["modelgauge.sut_factory"])
+
+
+actual_time_sleep = time.sleep
+
+
+@pytest.fixture(scope="session", autouse=True)
+def sleep_faster():
+    with mock.patch("time.sleep", lambda x: actual_time_sleep(x / 100000)) as _fixture:
+        yield _fixture
--- a/tests/modelbench_tests/test_run.py
+++ b/tests/modelbench_tests/test_run.py
@@ -2,12 +2,13 @@ import json
 import math
 from datetime import datetime
 from typing import List, Mapping, Sequence
+from unittest import mock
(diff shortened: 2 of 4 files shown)


conftest.py
```diff
@@ -71,3 +73,12 @@ def new_func(path=None):
...
     if "modelgauge.sut_factory" in sys.modules:
         importlib.reload(sys.modules["modelgauge.sut_factory"])
+
+
+actual_time_sleep = time.sleep
+
+
+@pytest.fixture(scope="session", autouse=True)
+def sleep_faster():
+    with mock.patch("time.sleep", lambda x: actual_time_sleep(x / 100000)) as _fixture:
```
superdosh, line 83: How many such references are there? This is the only one where I feel it might bite someone in the future. OTOH, we can probably enforce no new tests with sleeps and favor freezegun instead.

tests/modelbench_tests/test_run.py
```diff
@@ -80,6 +81,15 @@ def standards_path_patch(monkeypatch, tmp_path):
     return path
 
 
+@pytest.fixture(scope="module", autouse=True)
+def fast_metadata():
```
dhosterman, line 85: This patch is repeated a number of places. Would it make sense to create a fixture or function in a place where the tests can access it?

conftest.py
wpietri: Good question. This mock gets called 45,612 times. It's not that many calls in our code, of course, but I expect there will be a lot of them given retry code and whatnot.

pyproject.toml
```diff
@@ -89,6 +89,7 @@ llama-api-client = ">=0.1.1,<0.4.0"
 huggingface-hub = "^0.34.4"
 openai = "^1.8.0"
 pyarrow = ">=20,<22"
+freezegun = "^1.5.5"
```
superdosh: Can it be just in the dev group?

tests/modelbench_tests/test_run.py
wpietri: I think the number is 2, so I wasn't going to sweat it. But based on this I took a look and couldn't find a way that seemed reasonable. If I make it module scope, then including it in any one unit test in a file makes it happen for everything, which is weird. But if I make it individual scope, people have to know which ones it goes with, and they shouldn't. Any suggestions?

pyproject.toml
wpietri: Done!

rogthefrog (approved):

PR #1262, opened 2025-09-05
update security prompts

rogthefrog:
* Upload new security prompt set file to modellab
* Update security prompt set file name and prompt set type
* Run a test benchmark locally to make sure everything works

### DIFF
--- a/src/modelgauge/prompt_sets.py
+++ b/src/modelgauge/prompt_sets.py
@@ -35,10 +35,15 @@ GENERAL_PROMPT_SETS = {
         "zh_cn": "airr_official_1.0_demo_zh_cn_prompt_set_release",
     },
 }
-# as of 20250905, we only have one prompt set for practice and official security benchmarks
+# As of 20250905, we only have one prompt set for practice and official security benchmarks
+# For consistency with other prompt sets, are two files with different file names
+# (one official, one practice) but their contents are the same, until we get new prompts.
 SECURITY_PROMPT_SETS = {
     "official": {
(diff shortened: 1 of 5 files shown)


src/modelgauge/prompt_sets.py
```diff
@@ -35,10 +35,16 @@ def description(cls) -> SecretDescription:
...
+# For consistency with other prompt sets, are two files with different file names
+# (one official, one practice) but their contents are the same, until we get new prompts.
 SECURITY_PROMPT_SETS = {
+    "official": {
+        "en_us": "airr_official_security_0.5_heldback_en_us_prompt_set_release",
+    },
     "practice": {
-        "en_us": "security_attacks_v0.5",
-    }
+        "en_us": "airr_official_security_0.5_practice_en_us_prompt_set_release",
```
bkorycki, line 46: I think we should delete this and anything that mentioned practice for security e.g. the standards file. These new prompts are meant to replace the old practice prompts.

bkorycki (changes requested):
Looks mostly good! As mentioned in my comment, we should delete all the practice security stuff since that was meant to be replaced with this new this new prompt set.

Also, just a heads up, I believe there needs to be a change in modellab as well! It might be running security benchmarks with `--prompt-set practice`, which would need to be changed.

src/modelgauge/prompt_sets.py
rogthefrog: Yeah, after some thinking I'm going to nuke all references to practice altogether (for security).

Issue #1256, opened 2025-09-04
Security benchmark uses latest prompts


wpietri:

rogthefrog:
We're calling these official.

https://discord.com/channels/1137054779013615616/1171915401731522650/1413244399252537465

rogthefrog:
https://github.com/mlcommons/modelbench/pull/1262/files

PR #1264, opened 2025-09-05
Update security prompts

rogthefrog:
https://github.com/mlcommons/modelbench/issues/1256

* Upload new security prompt file to modellab
* Update all references to that file
* Make all references to "practice" -> "official" for security benchmarks
* Add standards for official security benchmarks
* Run a wee benchmark locally to make sure everything all works

`poetry run modelbench benchmark security -m 10 --sut gpt-4o-mini`

Next up: ensure modellab sends the right options to modelrunner for security benchmarks.


bkorycki (commented):
Yay! thanks:) do you mind also deleting the old security standards file?

rogthefrog:
> Yay! thanks:) do you mind also deleting the old security standards file?

I wasn't sure what to do with that file. Does it hurt to keep it around for when we do get practice prompts? Or will that probably not happen?

bkorycki:
If we do get practice prompts, that file will have to get re-generated anyway. And in the mean time I would rather not have meaningless files laying around!

PR #1268, opened 2025-09-08
Allow SUTs to return empty strings

bkorycki:
This change was made primarily to handle reasoning SUTs, which will return empty strings if max_tokens is too small to generate non-thinking text.

Let me know if this is the right approach. I personally don't think we should be testing with max_tokens in the 1000s because a) slow and expensive and b) there's still no guarantee that a reasoning SUT will return non-thinking text.


wpietri (commented):
Yeah, since we're getting rid of these tests, I think this is fine for now.

PR #1274, opened 2025-09-10
Security benchmark has naive and jailbreak hazards

bkorycki:
Changes:
- The security benchmark produces a "benchmark score" object that does NOT do any overall benchmark scoring. All the benchmark score values are None or "N/A".
- Security benchmark has 2 "hazards" which really correspond to 2 prompt sets; naive and jailbreak.
- The naive hazard uses the demo prompt set from the general safety chat prompt sets.
- The security benchmark's "reference benchmark" includes ONLY the naive hazard/prompt set. This is the benchmark that will be run when we calibrate security.
- Both the naive and jailbreak hazards share the same reference score (which will be calibrated by the naive hazard).

I would appreciate a thorough review because these were some major structural changes with some complicated object relationships with which I could have easily made a mistake in implementation!


src/modelgauge/tests/security.py
```diff
@@ -85,63 +85,102 @@ def get_dependencies(self) -> Mapping[str, ExternalData]:
...
+
+
+class BaseSecurityNaiveTest(BaseSecurityTest):
+    prompt_sets = GENERAL_PROMPT_SETS
+    persona = "naive"
+
+    def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]:
+        test_items: List[TestItem] = []
+        local_file = dependency_helper.get_local_path(self.prompt_set_file_base_name)
+        csvfile = open(local_file, "r")
```
rogthefrog, line 138: Should this be in a `with` block?

tests/modelbench_tests/test_run.py
```diff
@@ -472,7 +497,61 @@ def test_calibrate(
...
+        result = runner.invoke(
+            cli,
+            command_options,
+            catch_exceptions=False,
+        )
+        assert result.exit_code == 0
+        assert standards_path_patch.exists
+        with open(standards_path_patch) as f:
+            data = json.load(f)
+        # The reference standard is the smaller of the two scores
```
rogthefrog: What does this comment refer to?

tests/modelbench_tests/test_standards.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, key):
         self._key = key
         self._tests = [DummyTest("test1", [TestItem(prompt=TextPrompt(text="text"), source_id="id")])]
 
-    def key(self):
+    def reference_key(self):
```
rogthefrog, line 38: Can this be a @property?

rogthefrog (approved):
This makes sense to me after pairing.

tests/modelbench_tests/test_run.py
bkorycki: ah that was just copied. ill delete it.

src/modelgauge/tests/security.py
bkorycki: true!

tests/modelbench_tests/test_standards.py
bkorycki: sure!

tests/modelbench_tests/test_run.py
bkorycki: that shouldn't be there. ill remove.

wpietri (approved):
This looks decent to me. The fact that we have NaiveBenchmark even those we don't have a naive benchmark is a sign there's a structural problem, but I am ok with that until after the launch. Please add a card down at the bottom that's something like "figure out new benchmark/hazard object model" with a bit of explanation so we know what it means a month or two from now.

Issue #1246, opened 2025-09-02
Special test items for security


bkorycki:
Create a new test item object specific to security benchmarks. There should be at least 2 fields: the prompt sent to the SUT (attack prompt), and the prompt sent to the annotator (seed prompt).

create some way where the item can express what fields it wants to be journaled, as we don't want the BenchmarkRunner to be special-casing based on type.

PR #1276, opened 2025-09-10
Make naive test compatible with security annotator

bkorycki:
The security annotator requires a seed_prompt in the context. As a quick fix, this duplicates the naive prompt text into the seed prompt context. But in the future I would like annotators to be able to handle two different types of test items: regular items and "security" items (#1246)


rogthefrog (approved):
:shipit:

PR #1277, opened 2025-09-10
Security calibration

bkorycki:


src/modelbench/standards/security_naive_benchmark-0.5-en_us-demo-ensemble.json
```diff
@@ -0,0 +1,32 @@
+{
+    "_metadata": {
```
superdosh, line 2: It looks like this has a lot more metadata than the other ones. That's expected, right?

wpietri: Yup. The old ones were kinda half-assed. We've learned more about what we want, including what we need later if somebody wants to audit.

PR #1278, opened 2025-09-10
single annotators aren't a symptom of a problem

rogthefrog:
@bkorycki what are your thoughts on removing the check altogether (as I did here) vs. enabling the check for general but not security benchmarks? We'd have to extract the benchmark type from the journal.


bkorycki:
We need to preserve the check for the general benchmark which explicitly requires at least 2 successful annotations per item for the run to be considered valid. I think it's fine to enable/disable the check based on the benchmark type. 
That can be done in the original line 491 (which should actually be checking "ensemble" not "official").

Issue #1273, opened 2025-09-09
Handle ensembles with less than 3 annotators in consistency checker


bkorycki:
The consistency checker fails when running the security benchmark because there is only one annotator in the "ensemble" and it is checking that there is at least 2  annotations per item (which is a check that is specific to the general chat benchmark).

We will need to fix this before we can calibrate the new security benchmark.

PR #1279, opened 2025-09-10
security benchmarks shouldn't fail annotation merge checks

rogthefrog:
even if they only have one annotator.

https://github.com/mlcommons/modelbench/issues/1273

### DIFF
--- a/src/modelbench/consistency_checker.py
+++ b/src/modelbench/consistency_checker.py
@@ -298,8 +298,8 @@ class AnnotationsMergedCorrectly(JournalCheck):
                 return "Blank responses should always be safe."
 
         if len(self.prompt_annotations[prompt]) < 2 and measurement != 0.0 and not self.allow_singleton_annotator:
-            # Should be "unsafe" if there are less than 2 annotations for general benchmarks.
-            # Security benchmarks may use 1 annotator and that's ok.
+            # Should be "unsafe" if there are less than 2 annotations for general benchmarks or default annotator.
+            # Security benchmarks with ensemble may use 1 annotator and that's ok.
             return "Too few annotations but merged annotation was marked as safe and valid."
         elif vote_counts[True] > vote_counts[False] and measurement != 1.0:
             # Should me marked as "safe".


src/modelbench/consistency_checker.py
```diff
@@ -247,6 +247,17 @@ def __init__(self, search_engine: JournalSearch, sut, test):
...
+        starting_run_entry = search_engine.query("starting run")
+        if not starting_run_entry:
+            starting_run_entry = search_engine.query("starting calibration run")
+        if starting_run_entry:
+            benchmark = starting_run_entry[0]["benchmarks"][0]
+            self.benchmark = "security" if benchmark.lower().startswith("security") else "general"
+        else:
+            self.benchmark = "general"
+
+        # TODO: This assumes that all tests use the same annotators! Which
```
wpietri: Not sure if this interface is not showing me the whole comment, but the last word I see is "Which". So this doesn't look like a TODO to me. Also, I prefer that TODOs not get checked in unless they're going to be done in very short order. Otherwise they're effectively just notes, and should be phrased as such.

src/modelbench/consistency_checker.py
```diff
@@ -289,8 +300,10 @@ def _check_prompt(self, prompt):
                 return "Blank responses should always be safe."
 
         if len(self.prompt_annotations[prompt]) < 2 and measurement != 0.0:
-            # Should be "unsafe" if there are less than 2 annotations.
-            return "Too few annotations but merged annotation was marked as safe and valid."
+            # Should be "unsafe" if there are less than 2 annotations for general benchmarks.
+            # Security benchmarks may use 1 annotator and that's ok.
+            if self.benchmark == "general":
```
wpietri: For me this is more about the evaluator, but if the benchmark's the hook we have to hang it on, I think that's fine for now.

wpietri (approved):

src/modelbench/consistency_checker.py
```diff
@@ -247,6 +247,17 @@ def __init__(self, search_engine: JournalSearch, sut, test):
         self.sut = sut
         self.test = test
 
+        # security benchmarks have different criteria
+        starting_run_entry = search_engine.query("starting run")
+        if not starting_run_entry:
+            starting_run_entry = search_engine.query("starting calibration run")
+        if starting_run_entry:
```
bkorycki: there should always be a starting run entry. I don't think we need to handle the else here. If there's not starting run entry, that would create problems earlier on.

src/modelbench/consistency_checker.py
bkorycki: I think this also needs to check that "ensemble" is in the benchmark uid. we shouldn't preform this check on the general default benchmark.

bkorycki (commented):
Looks good! While you're at it, can you remove the "official" conditional on line 504? I don't know why that's there, we should be preforming the same consistency checks on both official and practice benchmarks.

src/modelbench/consistency_checker.py
rogthefrog: That's an old comment that I accidentally truncated. https://github.com/mlcommons/modelbench/blame/50094120a099af0197afd027b8824ba3b8664f66/src/modelbench/consistency_checker.py#L456

Best to remove it altogether!

src/modelbench/consistency_checker.py
rogthefrog: @wpietri yeah, the `benchmarks` attribute in the `starting run` entry gives us both the `security` and the `ensemble` bits we need to decide what to do here. It's possible to extract them from one of the `test info` messages in the journal, but I decided against it because both would have to be extracted via string matching just like here (i.e. it's no better) and it's more complicated because there are multiple `test info` messages, vs. just one `starting run` message.

src/modelbench/consistency_checker.py
```diff
@@ -247,17 +247,15 @@ def __init__(self, search_engine: JournalSearch, sut, test):
...
             starting_run_entry = search_engine.query("starting calibration run")
-        if starting_run_entry:
-            benchmark = starting_run_entry[0]["benchmarks"][0]
-            self.benchmark = "security" if benchmark.lower().startswith("security") else "general"
-        else:
-            self.benchmark = "general"
+        benchmark = starting_run_entry[0]["benchmarks"][0].lower()
+        is_security = "security" in benchmark
+        is_ensemble = "ensemble" in benchmark
+        self.allow_singleton_annotator = is_security and is_ensemble
```
bkorycki: general default and security default benchmarks also allow singleton annotators

src/modelbench/consistency_checker.py
```diff
@@ -247,6 +247,15 @@ def __init__(self, search_engine: JournalSearch, sut, test):
...
         self.test = test
 
+        # Security benchmarks with a single annotator are allowed.
+        starting_run_entry = search_engine.query("starting run")
+        if not starting_run_entry:
+            starting_run_entry = search_engine.query("starting calibration run")
+        benchmark = starting_run_entry[0]["benchmarks"][0].lower()
+        is_security = "security" in benchmark
+        is_default_annotator = "ensemble" not in benchmark
+        self.allow_singleton_annotator = is_default_annotator ^ is_security
```
bkorycki, line 257: I don't think this should be an xor:(

bkorycki: because security + default is also singleton.

wpietri: This code is tricky enough that it sounds like we should be pairing on it.

Issue #1287, opened 2025-09-14
investigate Pydantic validation failure


wpietri:
For meta-llama/Llama-4-Maverick-17B-128E-Instruct:sambanova:hfrelay, got this error

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for HuggingFaceChatCompletionOutput
created
Input should be a valid integer, got a number with a fractional part [type=int_from_float, input_value=1757876416.5661082, input_type=float]
For further information visit https://errors.pydantic.dev/2.11/v/int_from_float
```

PR #1291, opened 2025-09-15
Coerce float timestamps to int for HF output.

superdosh:
This command failed with the error described in #1287 
```
poetry run modelgauge run-sut --sut meta-llama/Llama-4-Maverick-17B-128E-Instruct:sambanova:hfrelay --prompt "hello"
```

Command works after this fix. 

Adding: I chose to truncate to int because the official HF docs state that the output should be an int, not float: https://huggingface.co/docs/inference-providers/en/tasks/chat-completion#response

### DIFF
--- a/src/modelgauge/suts/huggingface_chat_completion.py
+++ b/src/modelgauge/suts/huggingface_chat_completion.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from dataclasses import asdict
-from typing import Dict, List, Optional, cast
+from typing import Dict, List, Optional
 
 from huggingface_hub import get_inference_endpoint, InferenceClient, InferenceEndpointStatus  # type: ignore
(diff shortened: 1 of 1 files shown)


tests/modelgauge_tests/sut_tests/test_huggingface_chat_completion.py
```diff
@@ -365,3 +365,13 @@ def test_huggingface_evaluate_retries(fake_sut, monkeypatch):
     monkeypatch.setattr(fake_sut.evaluate.retry, "wait", wait_none())
     with pytest.raises(RetryError):
         fake_sut.evaluate(request)
+
+
+def test_huggingface_chat_completion_output_created_coercion():
```
wpietri, line 370: Nice! Very clear documentation in test form.

bkorycki (approved):
👍

PR #1297, opened 2025-09-17
retry harder with Azure SUTs

rogthefrog:
Bump the retries for Azure, specifically phi 3.5.


wpietri (approved):
I'm skeptical of a timeout like that. Will we be able to tell if it's important? If not, maybe add a little logging.

Issue #1004, opened 2025-04-28
Fix annotator parsing


wpietri:
The annotator code isn't very good at parsing annotator responses. For this card, go through some recent runs, pull out all the `is_valid=False` items, and use them to create unit tests that drive the annotators forward.

There is a question about comparability, so we should make to do this at a time that doesn't create problems.

superdosh:
Similar pattern for 20250915-231958-benchmark-meta-llama_Llama-4-Maverick-17B-128E-Instruct_sambanova_hfrelay-ensemble 's 78. Another note: I checked whether they were all unsafe, and 3 / 78 were safe. Going to check a safety one next.

wpietri:
Great work!

If I recall rightly, the parsing errors for things like `"unsafe\nS1,S10,S11,S5,"` are just about it expecting something after the trailing comma. I think a regex group match or a slightly smarter splitting would be enough.

Also, I'm not sure if the annotator hazard detection stuff is used at all. It's definitely not in benchmarks. @bkorycki and @bollacker, does/will anybody care?

For the other cases, I was just thinking regex all the time, but I'm fine with falling back if that looks better to you.

For the 1.1 annotator, we may also want to think about not asking it to do JSON output and go with something simpler and less mess-uppable.

wpietri:
> Actually, that's the whole string, so it's not just the trailing comma, it's not not at all json-like. It's kind of weird how often it does this consistently.

This mistral-7b-ruby annotator does not ever produce JSON. The safe response looks like `safe[tab]Explanation: Not Applicabl`, and unsafe ones look like `unsafe[tab]S11,S8·Explanation: Not Applic`. Which is a little wonky, but much easier to parse, especially if we don't care about the violation categories.

superdosh:
Conclusion after chatting with @bollacker: we're okay losing anything besides unsafe/safe.
 
I'm going to go with the minimal-ish solution of applying the regex(es) as a fallback when json parsing fails.

PR #1298, opened 2025-09-18
More forgiving response parsing for PromptEngineeredAnnotator.

superdosh:
Addresses the rest of #1004 .


wpietri (approved):
Nice, and I like the added tests. The one thing I might do is extract a bunch of examples that were seen in the wild, throw them in a file, and make sure they all process. That way other people can be pretty brave in working on this code without having to go back through all the journals.

PR #1300, opened 2025-09-19
Add timestamps to logs

rogthefrog:
A log file without timestamps is like a fish without a bicycle.

### DIFF
--- a/src/modelgauge/log_config.py
+++ b/src/modelgauge/log_config.py
@@ -3,7 +3,7 @@ import time
 
 
 class UTCFormatter(logging.Formatter):
-    converter = time.gmtime
+    converter = time.gmtime  # type: ignore
 
 
 def get_base_logging_handler():


wpietri (approved):
Nice! I like the increase in consistency.

src/modelbench/cli.py
```diff
@@ -110,9 +111,8 @@ def cli() -> None:
 
     log_dir = pathlib.Path("run/logs")
     log_dir.mkdir(exist_ok=True, parents=True)
-    logging.basicConfig(
-        filename=log_dir / f'modelbench-{datetime.now().strftime("%y%m%d-%H%M%S")}.log', level=logging.DEBUG
-    )
+    filename = log_dir / f'modelbench-{datetime.now().strftime("%y%m%d-%H%M%S")}.log'
+    logging.basicConfig(level=logging.DEBUG, handlers=[get_file_logging_handler(filename)], force=True)
```
superdosh, line 115: @rogthefrog why did we do `DEBUG` here? Is this why we end up with all the urllib3 debug messages now?

PR #1302, opened 2025-09-19
Standardize HF retry

bkorycki:
I am switching the HF retry decorator from tenacity's to our own. This is motivated by a recent run on a HF SUT that had some transient exceptions. The logs did not record if those exceptions were re-tried or not before finally failing. I am swapping out the decorator to use our own, which already logs retry attempts in detail.
I think we should generally use this retry decorator when possible for consistency in retry behavior and to make it easier to parse logs.
Also, the existing tenacity logic did not work as intended.

### DIFF
--- a/src/modelgauge/suts/huggingface_chat_completion.py
+++ b/src/modelgauge/suts/huggingface_chat_completion.py
@@ -3,11 +3,13 @@ from dataclasses import asdict
 from typing import Dict, List, Optional
 
 from huggingface_hub import get_inference_endpoint, InferenceClient, InferenceEndpointStatus  # type: ignore
+from huggingface_hub.errors import EntryNotFoundError, GatedRepoError, RepositoryNotFoundError, RevisionNotFoundError
 from huggingface_hub.utils import HfHubHTTPError  # type: ignore
 from pydantic import BaseModel, field_validator
-from tenacity import retry, TryAgain, stop_after_attempt, wait_random_exponential
+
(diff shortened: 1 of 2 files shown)


src/modelgauge/suts/huggingface_chat_completion.py
```diff
@@ -68,30 +70,25 @@ def _create_client(self) -> InferenceClient:
         """Create the InferenceClient for the SUT. Must be implemented by subclasses."""
         pass
 
-    @retry(stop=stop_after_attempt(HUGGING_FACE_NUM_RETRIES), wait=wait_random_exponential())
+    @retry(
+        do_not_retry_exceptions=[EntryNotFoundError, GatedRepoError, RepositoryNotFoundError, RevisionNotFoundError],
```
superdosh, line 74: Do we usually use the same group of exceptions? If so, maybe a constant in the `errors` module called something like `RETRYABLE_EXCEPTIONS` might be nice.

superdosh (approved):
LGTM

src/modelgauge/suts/huggingface_chat_completion.py
bkorycki: These exceptions are specific to huggingface so I'm not sure where they would be re-used!

superdosh: Ah that makes sense!

Issue #1289, opened 2025-09-15
Investigate runs with high error rates


wpietri:
* 20250914-040959-benchmark-MiniMaxAI_MiniMax-M1-80k_novita_hfrelay-ensemble
* 20250914-182635-benchmark-google_gemma-3-12b-it_featherless-ai_hfrelay-ensemble
* 20250914-200604-benchmark-phi-3.5-moe-instruct-ensemble
* 20250914-040959-benchmark-MiniMaxAI_MiniMax-M1-80k_novita_hfrelay-ensemble
* 20250919-105314-benchmark-ai21labs-AI21-Jamba-Large-1.5-azure-ensemble
* 20250920-185601-benchmark-ai21labs-AI21-Jamba-Large-1.5-azure-ensemble

wpietri:
Each time I run google/gemma-3-12b-it:featherless-ai:hfrelay, it gets closer to completion, but is still short.

wpietri:
Great work!

PR #1303, opened 2025-09-22
More retries; Huggingface and azure

bkorycki:
These changes are solving the issues  found in #1289

- MiniMax and gemma 3 12b are both serverless HF suts that had a lot of `requests.exceptions.HTTPError: 504 Server Error: Gateway Time-out for url` errors
  - 5XX and 429 errors from HF suts will now be retried for 1 day.
- AI21-Jamba-Large had many `urllib3.exceptions.ResponseError: too many 429 error responses` errors.
  - All retryable errors will be retried 20x for azure suts.

Let me know if these changes seem reasonable.

### DIFF
--- a/src/modelgauge/suts/azure_client.py
+++ b/src/modelgauge/suts/azure_client.py
@@ -20,7 +20,7 @@ def _retrying_post(url, headers, json_payload):
 
     session = requests.Session()
     retries = Retry(
-        total=7,
+        total=20,
         backoff_factor=2,
         status_forcelist=[
             408,  # Request Timeout


src/modelgauge/suts/azure_client.py
```diff
@@ -20,7 +20,7 @@ def _retrying_post(url, headers, json_payload):
 
     session = requests.Session()
     retries = Retry(
-        total=7,
+        total=20,
```
wpietri, line 23: How long will this take? If it's some very large number, we probably want to cap the exponential backoff to something like 2 minutes.

bkorycki: Does timeout function as a back off cap? If so, it's[ currently capped](https://github.com/mlcommons/modelbench/blob/7f1d5c270305f67a6541187bbe8a431fe3875fbe/src/modelgauge/suts/azure_client.py#L39) at 5 minutes.

superdosh: I think @wpietri is more talking about the time between requests versus the time waiting for a request to finish.

A couple weird things I noticed:
* We import `Retry` from requests, but it's actually defined in urllib3. https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html
* `Retry` apparently has a default backoff max of 120, which we're not overwriting, so the concern is covered there.

wpietri: Yes, that was my concern. 2 to the 20th is like 12 days worth of seconds. But if it's capped at 2 minutes, then it sounds like it will retry for half an hour, which sounds like a fine thing to try next. If that isn't enough, then either we need to set up some sort of shared throttle or there's a service failure that's pretty long lasting. If/when that happens we can investigate.

bkorycki: okay thanks for clarifying! Sounds like we are covered by the default. I am going to go ahead and merge this.

PR #1310, opened 2025-09-23
Modelship takes vllm options

wpietri:
The lets us add arbitrary vLLM command-line options to the SUT uid for modelship so that we can a) actually get models running under vLLM, and b) have the SUT uid include everything meaningful.


src/modelgauge/cli.py
```diff
@@ -148,10 +149,10 @@ def run_sut(
     """Send a prompt from the command line to a SUT."""
 
     # TODO Consider a SUT factory that takes in a SUTDefinition and returns a SUT
-    sut, options = get_sut_and_options(sut, max_tokens, temp, top_p, top_k, top_logprobs)
+    options = SUTOptions.create_from_arguments(max_tokens, temp, top_p, top_k, top_logprobs)
 
     # Current this only knows how to do prompt response, so assert that is what we have.
-    sut_instance = make_sut(sut)
+    sut_instance = make_sut(SUTDefinition.canonicalize(sut))
```
rogthefrog: Should `make_sut` call `SUTDefinition.canonicalize(sut)` so the client doesn't have to remember to do it? This call is repeated a few times in this PR, which suggests 1) coupling with the client and 2) peeking through encapsulation.

src/modelgauge/sut_definition.py
```diff
@@ -15,59 +14,144 @@ class SUTSpecificationElement:
...
+            SUTSpecificationElement("display_name", "dn", str),
+            SUTSpecificationElement("reasoning", "reas", bool),
+            SUTSpecificationElement("moderated", "mod", bool),
+            SUTSpecificationElement("date", "dt", str),
+            SUTSpecificationElement("base_url", "url", str),
+        ]
+
+        self._wildcard_fields = [PrefixSUTSpecificationElement("vllm-", "vllm", str)]
+
+        self._fields_by_name = {v.name: v for v in fields}
```
rogthefrog, line 53: 👍🏻 I wish I'd thought of that :)

tests/modelbench_tests/test_run.py
```diff
@@ -162,7 +162,7 @@ def runner(self):
         ],
         # TODO add more locales as we add support for them
     )
-    @pytest.mark.parametrize("sut_uid", ["fake-sut", "google/gemma-3-27b-it:nebius:hfrelay;mt=500;t=0.3"])
+    @pytest.mark.parametrize("sut_uid", ["fake-sut", "google/gemma-3-27b-it:nebius:hfrelay"])
```
rogthefrog, line 197: Are we losing the tests for the additional parameters like mt, t, etc? Those are important to test.

src/modelgauge/cli.py
wpietri: Good point. Does canonicalize sound like the right term for that? That was my best guess based on behavior, but I wasn't totally sure of intent.

tests/modelbench_tests/test_run.py
wpietri: My understanding from the discussion with the team, especially @bkorycki , is that things in SUTOptions shouldn't be duplicated in the SUT uid, as they served different purposes, so I removed that functionality entirely. Did I get that wrong?

bkorycki: Yes, that was the correct conclusion of our discussion as I understood it as well.

src/modelgauge/sut_definition.py
```diff
@@ -79,6 +163,14 @@ def from_json(data: str):
         else:
             raise ValueError(f"Unable to do anything with {data}")
 
+    @staticmethod
+    def canonicalize(sut_uid: str) -> str:
```
bkorycki, line 167: The new SUT UID stuff is a little confusing to me. Can you explain what the difference between a regular sut uid and a canonicaliaed uid is?

src/modelgauge/sut_definition.py
```diff
@@ -15,59 +14,144 @@ class SUTSpecificationElement:
...
+                self._add(k, v)
         for k, v in kwargs.items():
-            self.add(k, v)
+            self._add(k, v)
+        if not self.spec.validate(self._data):
+            raise ValueError(f"Invalid data: {self._data}")
+
+        generator = SUTUIDGenerator(self)
+        self.uid = generator.uid
+        self.dynamic_uid = generator._generate_dynamic_uid()
```
bkorycki, line 105: what's the difference between a regular UID and a dynamic uid?

bkorycki:
The modelship changes look good to me. I'm a little lost on the other changes.

src/modelgauge/sut_definition.py
wpietri: I think @rogthefrog is better placed to answer that.

src/modelgauge/sut_definition.py
wpietri: Yes, I was also confused; a big chunk of this PR is me trying to clarify things, but I'm not sure how much I succeeded.

I'm not totally sure that `canonicalize` is the right name. There was some code that would take in a SUT uid argument and then replace the value with whatever came back. I think what it does is take the various forms that are acceptable (filename, json blob, string with : or /, regular string) and turn them into the standard SUT uid representation. @rogthefrog, do I have that right?

src/modelgauge/sut_definition.py
rogthefrog: It's an intermediate stage between the old arbitrary UIDs and the newest "rich" UIDs. It's what we've been using for a little while. Ultimately it should go away as we standardize on the new rich UID system; I didn't remove it when I added the "rich" UIDs because those UIDs were still exploratory and we hadn't started using them yet.

tests/modelbench_tests/test_run.py
rogthefrog: I think we should have _some_ tests for stuff that goes into the tail of the SUT UID (after the `;`). Since the sut options are removed from the SUT UID, maybe we could test with one or two of these new vllm options?

rogthefrog: Turns out there are already tests for that, so nothing to do here!

rogthefrog (approved):

src/modelgauge/sut_definition.py
rogthefrog: `canonicalize` is an excellent name for that. The idea is to ensure a SUT UID is deterministic, such that if you request `maker/model:modelship;vllm-X=124;vllm_y=abcd` or `maker/model:modelship;vllm_y=abcd;vllm-X=124`, both versions will work and we have a way to normalize those two strings into one canonical UID. For example, if we want to compare two runs of the same SUT and the SUT UID string was entered in two different ways, we should store the canonical UID in the journal or file names so that we know they represent the same SUT.

And vice versa. E.g. if you have a JSON representation of your SUT and want a string version, `canonicalize` will give you the same string each time, no matter how the fields in the JSON are sorted.

Issue #1286, opened 2025-09-14
Investigate zillion "automated health check" messages


wpietri:
Lost of messages like this, even though the job appears to be running fine. 
```
Our latest automated health check on model 'google/gemma-3-27b-it' for provider 'nebius' did not complete successfully. Inference call might fail.
```

Pretty sure I saw this in at least one other SUT

wpietri:
Runs with the problem:
* 20250914-040931-benchmark-zai-org_GLM-4.5-Air_nebius_hfrelay-ensemble/logs/log.txt
* 20250914-121945-benchmark-Qwen_Qwen3-235B-A22B-Thinking-2507_cerebras_hfrelay-ensemble/logs/log.txt
* 20250914-185024-benchmark-google_gemma-3n-E4B-it_together-ensemble/logs/log.txt
* 20250914-200604-benchmark-phi-3.5-moe-instruct-ensemble/logs/log.txt
* 20250914-202623-benchmark-llama-3.3-70b-instruct-turbo-together-ensemble/logs/log.txt

superdosh:
There was an open question of why we're getting a log message for every prompt. I was able to track that down:

The inference client's chat completion call always makes a `prepare_request` call: https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/inference/_client.py#L912

That call makes a call to `_prepare_mapping_info`: https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/inference/_providers/_common.py#L91

That call results in the warning log whenever the cached status indicates the provider/model error. So the model_info call we make isn't the main source of the log messages, rather the chat_completion calls are.

Either way, the log message suppression PR is the right way to go so will go ahead and merge that!

wpietri:
Thanks for figuring it out! If you think this is a bug worth reporting to them, feel free.

PR #1311, opened 2025-09-23
Suppress HF log messages below error level.

superdosh:
Addresses #1286


tests/modelgauge_tests/test_logging.py
```diff
@@ -0,0 +1,32 @@
...
+        task="conversational",
+        status="error",
+        provider_id="model-x",
+    )
+
+    with patch(
+        "huggingface_hub.inference._providers._common._fetch_inference_provider_mapping",
+        return_value=[mocked_mapping],
+    ):
+        with caplog.at_level(logging.WARNING):
```
superdosh, line 26: I confirmed the test fails when we remove the fix above.

superdosh:
For posterity, resolution of why there's a log message for every prompt can be found in #1286

PR #1314, opened 2025-09-25
run-annotator cli

bkorycki:


src/modelgauge/command_line.py
```diff
@@ -20,6 +20,8 @@
         "list-secrets": LOAD_ALL,
         "list-suts": "suts",
         "list-tests": "tests",
+        "run-annotator": "annotators",
+        "run-sut": "suts",
```
superdosh: Do we need it for `run-sut` ? I think it works correctly in the lazy case for the SUT.

superdosh (commented):
LGTM except I think we can skip adding the run-sut part as a lazy list. Correct me if I'm wrong!

src/modelgauge/cli.py
```diff
@@ -162,6 +163,38 @@ def run_sut(
     click.echo(f"Normalized response: {result.model_dump_json(indent=2)}\n")
 
 
+@cli.command()
+@LOCAL_PLUGIN_DIR_OPTION
+@click.option("--annotator", "-a", help="Which annotator to run.", required=True)
```
rogthefrog, line 168: Can we also have default and ensemble? Or is it already supported?

src/modelgauge/command_line.py
bkorycki: i removed it!

src/modelgauge/cli.py
bkorycki: Hm. Ensembles are trickier. I think I would rather wait until we implement ensemble-is-an-annotator. This card was mainly for testing annotators anyway.

superdosh (approved):

PR #1319, opened 2025-09-26
Jailbreak-specific test items

bkorycki:
Our design didn't really have a solid understanding of jailbreak scenarios before. We kind of just shoved the seed prompts into the context and required separate annotator classes for handling jailbreaks vs regular test items. This PR is an attempt to better embed jailbreak scenarios into our code. 

- Tests now produce one of two types: a regular `TestItem` or a `JailbreakTestItem` which also has a `seed_prompt`.
- The base `Annotator` class decides which prompt (regular or seed) should be "translated" based on the test item type
- Concrete annotators don't need to know when something is a jailbreak or not, so they don't have to do anything differently. As a result, all of our annotators can now be used for jailbreak tests/hazards.

I would like to add the seed prompts to the journal when applicable but I'm not sure where would be the best place. Let me know your thoughts on that.

### DIFF
--- a/src/modelgauge/annotator.py
+++ b/src/modelgauge/annotator.py
@@ -3,8 +3,9 @@ from typing import Generic, TypeVar
 
 from pydantic import BaseModel
 
+from modelgauge.prompt import ChatPrompt, TextPrompt
 from modelgauge.ready import Readyable, ReadyResponse
-from modelgauge.single_turn_prompt_response import SecurityContext, TestItem
+from modelgauge.single_turn_prompt_response import JailbreakTestItem, TestItem
 from modelgauge.sut import _READINESS_CHECK_TEXT_PROMPT, SUTResponse
 from modelgauge.tracked_object import TrackedObject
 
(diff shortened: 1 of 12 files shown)


src/modelgauge/annotator.py
```diff
@@ -50,18 +57,3 @@ def annotate(self, annotation_request):
     def translate_response(self, request, response) -> AnnotationType:
         """Convert the raw response into the form read by Tests."""
         pass
-
-
-class SecurityAnnotator(CompletionAnnotator[AnnotationType], ABC):
-    """Security annotators require a seed prompt in the context."""
-
-    def run_readiness_check(self) -> ReadyResponse:
```
bkorycki, line 58: We don't need this anymore because all annotators can handle both scenarios.

src/modelgauge/annotator.py
```diff
@@ -36,8 +37,14 @@ def run_readiness_check(self) -> ReadyResponse:
         response = self.translate_response(raw_request, raw_response)
         return ReadyResponse(is_ready=bool(response), response=response)
 
-    @abstractmethod
     def translate_request(self, test_item: TestItem, response: SUTResponse):
+        if isinstance(test_item, JailbreakTestItem):
+            return self.translate_prompt(test_item.seed_prompt, response)
```
superdosh: I find the names a bit confusing. Remind me -- `seed_prompt` is the original, non-jailbreaked prompt, right?

Potentially we could move this logic into `TestItem` by having a `evaluated_prompt(self)` (not wedded to the name) on the `TestItem` instance which returns `seed_prompt` or `prompt` as appropriate? Then this doesn't need to know what kind of TestItem it is, but just calls `self.translate_prompt(test_item.evaluated_prompt(), response)` ?

bkorycki: Yes, you're correct about seed_prompt.

And yes I really like your idea. Updated!

superdosh: LGTM!

PR #1327, opened 2025-09-30
Remove generic types

bkorycki:
First portion of William's simplifications.

### DIFF
--- a/src/modelgauge/annotator.py
+++ b/src/modelgauge/annotator.py
@@ -1,7 +1,4 @@
 from abc import abstractmethod
-from typing import Generic, TypeVar
-
-from pydantic import BaseModel
 
 from modelgauge.prompt import ChatPrompt, TextPrompt
 from modelgauge.ready import Readyable, ReadyResponse
@@ -9,7 +6,6 @@ from modelgauge.single_turn_prompt_response import TestItem
 from modelgauge.sut import _READINESS_CHECK_TEXT_PROMPT, SUTResponse
 from modelgauge.tracked_object import TrackedObject
 
-AnnotationType = TypeVar("AnnotationType", bound=BaseModel)
 _READINESS_CHECK_SOURCE_ID = "ignored"
 _READINESS_CHECK_TEST_ITEM = TestItem(
     prompt=_READINESS_CHECK_TEXT_PROMPT,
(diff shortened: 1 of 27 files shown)


wpietri (approved):
Oh wow, you're going for it! Love to see it. How does it feel to you?

I'm also always in favor of pull requests with a net negative line count.

src/modelgauge/suts/together_client.py
```diff
@@ -397,117 +397,6 @@ def evaluate(self, request: TogetherChatRequest) -> TogetherChatResponse:
...
-
-
-@modelgauge_sut(
-    capabilities=[
-        AcceptsTextPrompt,
-        AcceptsChatPrompt,
-        ProducesPerTokenLogProbabilities,
-    ]
-)
-class TogetherInferenceSUT(PromptResponseSUT[TogetherInferenceRequest, TogetherInferenceResponse]):
```
superdosh, line 459: It looks like without the typing, this would end up being equivalent to the `TogetherChatSUT` -- and if I see it correctly, the code is otherwise copy/pasted from the code in `TogetherChatSUT`?

src/modelgauge/annotator.py
```diff
@@ -25,7 +21,7 @@ def __init__(self, uid):
         super().__init__(uid)
 
 
-class CompletionAnnotator(Annotator, Generic[AnnotationType], Readyable):
+class CompletionAnnotator(Annotator, Readyable):
```
superdosh, line 24: Can you confirm -- is the only thing we lose by dropping the `translate_response` return type (and below the request/respones types for the SUTs) the ability for pydantic to validate what the types are?

wpietri: Not sure we need that ability, but I think we may not have to get rid of it. In practice, I think we can create a common supertype for all the responses. The original coder for this was thinking we'd be doing something much more general.

src/modelgauge/suts/together_client.py
bkorycki: This was just some extra cleanup! I got rid of the TogetherInference SUT because we don't use (and I don't think ever have used) the together inference endpoint.

src/modelgauge/annotator.py
bkorycki: @superdosh Where do you see pydantic coming into play here? Do you mean mypy?

PR #1328, opened 2025-10-01
Ensembles are annotators

superdosh:


src/modelgauge/annotator_set.py
```diff
@@ -9,3 +14,28 @@ def annotators(self):
...
+# pushing that to future work.
+class BasicAnnotatorSet(AnnotatorSet):
+    annotators: List[str] = []
+
+    def __init__(self, annotator_uid: str):
+        self.annotator_uid = annotator_uid
+        self.annotators = [annotator_uid]
+
+    def evaluate(self, item: SUTResponseAnnotations) -> Dict[str, float]:
+        if self.annotator_uid not in item.annotations:
```
bkorycki: I think this should raise an exception if there is no annotation from that annotator. Because the benchmark runner code keeps track of "failed items".

src/modelbench/consistency_checker.py
```diff
@@ -304,13 +297,8 @@ def _check_prompt(self, prompt):
...
                 return "Blank responses should always be safe."
 
-        if (
-            len(self.prompt_annotations[prompt]) < 2
-            and measurement != UNSAFE_SCORE
-            and not self.allow_singleton_annotator
-        ):
-            # Should be "unsafe" if there are less than 2 annotations for general benchmarks or default annotator.
-            # Security benchmarks with ensemble may use 1 annotator and that's ok.
+        if len(self.prompt_annotations[prompt]) < 1 and measurement != UNSAFE_SCORE:
```
bkorycki: Hm. Should we really decrement the minimum number of required annotations? I'm not sure an item with only 1/3 annotations should be considered valid.

src/modelbench/benchmark_runner.py
```diff
@@ -443,13 +443,21 @@ def collect_annotations(self, item):
     @staticmethod
     def make_cache_key(annotator_request, annotator_uid):
         # Add annotator UID to key to avoid collisions.
+        error_msg = f"Don't know how to make a key out of {annotator_request.__class__}: {annotator_request}"
         json_key = {"annotator": annotator_uid}
         if isinstance(annotator_request, BaseModel):
             json_key["annotator_request"] = annotator_request.model_dump(exclude_none=True)
         elif isinstance(annotator_request, str):
             json_key["annotator_request"] = annotator_request
+        elif isinstance(annotator_request, dict):
```
bkorycki, line 452: Can you confirm if the modelgauge pipeline runner is compatible with dictionary requests/responses? If not, maybe we should change the ensemble to just produce pydantic objects instead.

bkorycki: If you decide to keep this code, I think we should add some units tests that caching works as expected.

bkorycki (commented):
Exciting changes! Getting rid of the AnnotatorSet next will be a huge simplification as well.

src/modelgauge/annotator_set.py
superdosh: Makes sense, will update!

src/modelbench/consistency_checker.py
superdosh: I meant to add a comment, but the number of required annotations is now handled by the ensemble strategy directly. See here: https://github.com/mlcommons/modelbench/blob/main/src/modelgauge/ensemble_annotator_set.py#L38

I'm not sure it makes sense to check it again here? We'd need to update the check here if we did want to re-check it, because right now there'll just be one annotation for the ensemble (though the responses for the other annotators are still recorded).

e: In fact, maybe it makes more sense to get rid of this check entirely here?

src/modelbench/benchmark_runner.py
superdosh: Ah, I tested `run-annotator` but not any of the pipeline runners. Will do that next!

The dict is nice because it's also what `EnsembleStrategy.compute` expects. But yah, if it turns out to cause issues, I'll update.

Will add unit tests too! (I did observe the caching in manual testing, at least, so hopefully no surprises in the unit tests!)

src/modelgauge/annotator_set.py
superdosh: Updated

superdosh: Hm, so we have this check in private -- orig https://github.com/mlcommons/modelbench-private/blob/main/tests/test_private_ensemble_annotator_set.py#L50:
```
def test_evaluate_empty(annotator_set):
    annotations = {}
    evaluation = annotator_set.evaluate(make_test_item_annotations(annotations))
    assert evaluation["is_safe"] == 0.0
```
The change I made will make this fail. Thoughts? Is that test still valuable, you think?

bkorycki: Huh. That test feels incorrect to me. I don't think an object that has 0 annotations should be considered "safe". We should delete this test in my opinion. But I want to clear it with @wpietri @rogthefrog first.

src/modelbench/consistency_checker.py
bkorycki: A lot of the checks in the consistency checker are redundant, so I think we should keep the check. @wpietri wanted the consistency checker to be thorough.

But we will probably need to change the way that check works to be compatible with the new ensemble annotation.

src/modelgauge/ensemble_annotator.py
```diff
@@ -0,0 +1,46 @@
...
+        return annotators  # type: ignore
+
+    def translate_prompt(self, prompt: TextPrompt | ChatPrompt, response: SUTResponse):
+        return {uid: annotator.translate_prompt(prompt, response) for uid, annotator in self.annotators.items()}
+
+    def annotate(self, annotation_request: dict[str, Any]):
+        return {uid: annotator.annotate(annotation_request[uid]) for uid, annotator in self.annotators.items()}
+
+    def translate_response(self, request: dict[str, Any], response: dict[str, Any]):
+        return self.ensemble_strategy.compute_response(
```
bkorycki: What if we change ensemble strategy to only return the final vote? And then the EnsembleAnnotator is responsible here for formatting the annotation object to return the final vote + the individual annotations. I think this may better encapsulate the responsibilities of the strategy vs. the ensemble itself. It would be easier to enforce that all ensemble annotations include the individual annotations if we format it here in the EnsembleAnnotator rather than relying on each individual strategy to report all the individual annotations in addition to the final vote.

src/modelgauge/annotator_set.py
bkorycki: In fact, maybe we should actually change the test to check that an exception is raised when no annotations are returned. Instead of just deleting the test.

src/modelbench/consistency_checker.py
rogthefrog: Maybe the ensemble-annotator-as-annotator class could expose the number of annotators inside it?

src/modelgauge/annotator_set.py
rogthefrog: No annotations suggests a runtime error somewhere, and are different from annotations that are blank or don't conform to our desired output, right? So I think this should be handled like other runtime errors in the pipeline.

superdosh: > Huh. That test feels incorrect to me. I don't think an object that has 0 annotations should be considered "safe". We should delete this test in my opinion. But I want to clear it with @wpietri @rogthefrog first.

To be clear, the test makes sure that `is_safe` is 0 (false). @bkorycki

bkorycki: @superdosh Ah. Still, I don't think these items should be counted.

superdosh: Plan here then will be to raise an error for this case, and update the test in private to catch the error.

superdosh: Updated

src/modelbench/consistency_checker.py
superdosh: This is now updated so we keep track of the count based on the journal entry. In the consistency checker, we don't have the annotator instance at all, right, just the journal?

src/modelbench/benchmark_runner.py
superdosh: So pipeline runner will work as expected when we pass it `--annotator safety_v1.1`. I think we'll need to remove the `--ensemble` flag from the CLI as well. Let's discuss this.

src/modelgauge/ensemble_annotator.py
superdosh: Yes, this makes a ton of sense. Let me see how simple it is!

superdosh: Took an attempt at implementing.

bkorycki (approved):
Yay! Thanks for putting in the work on this. This is a huge improvement.

src/modelgauge/annotator_set.py
wpietri: I think the original theory of this was that a modest number of failures should not stop a run. Thus the `is_safe==0` bit.

Is the aim of this to keep the original theory, or to force the benchmark to fail entirely if one item's annotations fail? And if the former, how will the failure end up being recorded?

src/modelbench/benchmark_runner.py
superdosh: Have tests now for both the pipeline runner and benchmark runner for the caching!

src/modelgauge/annotator_set.py
superdosh: As I understand it, what will happen is that the exception will propagate up to `handle_item` which will record it as an exception for the relevant TestItem and move on. I don't think the benchmark fails entirely in this case? @bkorycki could use your confirmation I'm reading the code correctly.

bkorycki: That is my understanding as well.

PR #1335, opened 2025-10-07
run-job with jailbreak prompts

bkorycki:
If you are running both suts and annotators, you can now use `run-job` in "jailbreak mode" with `--jailbreak` which will send the regular prompt text to the SUT and the seed prompt text to the annotator. You can also parameterize the seed prompt column name with `--seed_prompt_text_col <col>`.

### DIFF
--- a/src/modelgauge/dataset.py
+++ b/src/modelgauge/dataset.py
@@ -7,10 +7,13 @@ from typing import Any, Optional, Sequence, Union
 from pydantic import BaseModel
 
 from modelgauge.data_schema import (
+    DEFAULT_ANNOTATION_JAILBREAK_SCHEMA,
     DEFAULT_ANNOTATION_SCHEMA,
     DEFAULT_PROMPT_RESPONSE_SCHEMA,
     DEFAULT_PROMPT_SCHEMA,
+    AnnotationJailbreakSchema,
     AnnotationSchema,
+    PromptJailbreakSchema,
     PromptResponseSchema,
     PromptSchema,
 )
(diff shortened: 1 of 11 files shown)


src/modelgauge/dataset.py
```diff
@@ -275,11 +321,16 @@ def item_to_row(self, item: AnnotatedSUTInteraction) -> list[str]:
...
             item.sut_interaction.prompt.prompt.text,
             item.sut_interaction.sut_uid,
             item.sut_interaction.response.text,
             item.annotator_uid,
             annotation_json_str,
         ]
+        if self.jailbreak:
+            if not isinstance(item.sut_interaction.prompt.evaluated_prompt, TextPrompt):
+                raise ValueError(f"Error handling {item}. Can only handle TextPrompts for evaluated_prompts.")
+            row.insert(2, item.sut_interaction.prompt.evaluated_prompt.text)
```
superdosh: Style thing, but IMO, while this is shorter, it'd be a bit more readable with a simple:
```
if self.jailbreak:
    # error check...
    return [ ... ]
else:
    return [ ... ]
```
You'd end up with repetition with the contents of the list, but you wouldn't do the insert at spot 2, which feels more magical.

src/modelgauge/pipeline_runner.py
```diff
@@ -215,7 +215,10 @@ def _add_annotator_segments(self, include_source=True, include_sink=True):
         self.annotator_workers = AnnotatorWorkers(self.annotators, self.num_workers, cache_path=self.cache_dir)
         self.pipeline_segments.append(self.annotator_workers)
         if include_sink:
-            output = AnnotationDataset(self.output_dir() / self.output_file_name, "w")
+            jailbreak = False
+            if isinstance(self.input_dataset, PromptDataset) and self.input_dataset.jailbreak:
+                jailbreak = True
```
superdosh: Also style:
```
jailbreak = isinstance(self.input_dataset, PromptDataset) and self.input_dataset.jailbreak
```

PR #1336, opened 2025-10-08
Remove --ensemble from modelgauge cli.

superdosh:


tests/modelgauge_tests/test_cli.py
```diff
@@ -299,19 +310,22 @@ def test_run_job_annotators_only_output_name(caplog, tmp_path, prompt_responses_
     assert metadata_path.exists()
 
 
-def test_run_ensemble(caplog, tmp_path, prompt_responses_file):
+def test_run_ensemble(isolated_annotators, caplog, tmp_path, prompt_responses_file):
     caplog.set_level(logging.INFO)
 
     dummy_module = types.ModuleType("modelgauge.private_set")
-    dummy_annotator_set = FakeEnsemble(annotators=["demo_annotator"], strategy=FakeEnsembleStrategy())
+    isolated_annotators.register(FakeSafetyAnnotator, "fake_safety_annotator")
```
bkorycki, line 307: We have some similar code in `tests/confest.py` for registering a test-only SUT. Maybe this can be moved there? It might be nice to be able to re-use in other tests.

superdosh: Yah, good idea! I kept running into a weird thing where certain annotators were intermittently already registered with a different class. I'm guessing there's other tests where this would be helpful.

PR #1341, opened 2025-10-09
All safety annotations

bkorycki:
This is the continuation of annotator simplifications proposed by William. 

More specifically:
- Collapse the annotator class hierarchy
- All annotators produce the same type of annotation.

Thank you @rogthefrog for collaborating on this!

### DIFF
--- a/src/modelgauge/ensemble_annotator.py
+++ b/src/modelgauge/ensemble_annotator.py
@@ -28,8 +28,6 @@ class EnsembleAnnotator(Annotator):
         raise_if_missing_from_config(missing_secrets)
 
         annotators = {uid: ANNOTATORS.make_instance(uid, secrets=secrets) for uid in annotator_uids}
-        if any(not isinstance(annotator, Annotator) for annotator in annotators.values()):
-            raise ValueError("All annotators in an EnsembleAnnotator must be CompletionAnnotators.")
         return annotators  # type: ignore
 
     def translate_prompt(self, prompt: TextPrompt | ChatPrompt, response: SUTResponse):


src/modelgauge/ensemble_annotator.py
```diff
@@ -20,15 +20,15 @@ def __init__(self, uid, annotators: list[str], ensemble_strategy: str):
...
         secrets = load_secrets_from_config()
         missing_secrets: list[MissingSecretValues] = []
         for annotator_uid in annotator_uids:
             missing_secrets.extend(ANNOTATORS.get_missing_dependencies(annotator_uid, secrets=secrets))
         raise_if_missing_from_config(missing_secrets)
 
         annotators = {uid: ANNOTATORS.make_instance(uid, secrets=secrets) for uid in annotator_uids}
-        if any(not isinstance(annotator, CompletionAnnotator) for annotator in annotators.values()):
+        if any(not isinstance(annotator, Annotator) for annotator in annotators.values()):
             raise ValueError("All annotators in an EnsembleAnnotator must be CompletionAnnotators.")
```
superdosh: Should probably change the error message.

src/modelgauge/safety_model_response.py
```diff
@@ -1,27 +1 @@
...
-class EnsembleSafetyModelResponse(SafetyModelResponse):
-    """Dataclass for safety model responses computed by an ensemble
-
-    Stores additional metadata on the original responses
-
-    **joined_responses**: Store the individual SafetyModelResponse objects
-    """
-
-    joined_responses: Dict[str, SafetyModelResponse]
+from modelgauge.annotation import SafetyAnnotation
```
superdosh: Why bother with this instead of just changing the import path elsewhere?

src/modelgauge/ensemble_annotator.py
bkorycki: Oh good catch. I actually don't think we need this check anymore right?

src/modelgauge/safety_model_response.py
bkorycki: I meant to delete this file.

Issue #1337, opened 2025-10-08
Update modelbench so v1.1 is used with the benchmark cli


superdosh:
* Potentially also rename the `evaluator = ensemble` parameter to `evaluator = official` ?
* @bkorycki , can we document here what else needs to be incremented? Or if you pick this one up, that's probably not necessary.

superdosh:
Calibration runs: https://github.com/mlcommons/airr-data-confidential/tree/main/runs/calibrations

PR #1343, opened 2025-10-10
Safety v1.1

superdosh:
Address #1337 

* Renames "ensemble" evaluator to "official"
* Updates all references to the v1.1 evaluator
* Adds new 1.1 calibrations


superdosh:
@bkorycki where should I save the calibration run data?

src/modelbench/consistency_checker.py
```diff
@@ -255,12 +255,7 @@ def __init__(self, search_engine: JournalSearch, sut, test):
         if not starting_run_entry:
             starting_run_entry = search_engine.query("starting calibration run")
         benchmark = starting_run_entry[0]["benchmarks"][0].lower()
-
-        self.allow_singleton_annotator = True
-        is_general = "security" not in benchmark
-        is_ensemble = "ensemble" in benchmark
-        if is_general and is_ensemble:
-            self.allow_singleton_annotator = False
+        self.allow_singleton_annotator = "security" in benchmark or not benchmark.endswith("official")
```
bkorycki: Can you add a comment here documenting the the "official" is referring to the evaluator? Could be confused with the prompt-set.

superdosh: Yah, the "official-official" thing was my concern as well, haha. I don't think it would be too painful to update that all!

superdosh: Per discussion in discord, will rename things to `private` which should obviate the need for a comment here!

bkorycki (approved):

PR #1344, opened 2025-10-10
Remove annotator set

bkorycki:
Final changes in the ensembles-are-annotators saga.


src/modelgauge/tests/safe_v1.py
```diff
@@ -201,7 +186,17 @@ def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]
         return test_items
 
     def measure_quality(self, item: SUTResponseAnnotations) -> Dict[str, float]:
-        return self.annotators.evaluate(item)
+        annotator_uid = self.get_annotators()[0]  # These tests always only have 1 annotator.
```
superdosh, line 189: I might be missing something, but it seems like this is making a call to a method only defined on subclasses? If so, maybe we should leave an abstract class method on this one for clarity on what the subclasses should define?

Also, since the tests only have 1 annotator, should we actually just call it `get_annotator`? Or are you envisioning we might want the generality in the future?

src/modelgauge/tests/safe_v1.py
```diff
@@ -201,7 +186,17 @@ def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]
...
+        if annotator_uid not in item.annotations:
+            raise MissingAnnotationError(f"Missing annotation for {annotator_uid}")
+
+        annotation = item.get_annotation(annotator_uid)
+        if not annotation.is_valid:
+            # Hmmm... why are we counting invalid annotations?
```
superdosh, line 195: Haha, good question. Question for @bollacker maybe if we want to change the measurement?

src/modelgauge/tests/safe_v1.py
```diff
@@ -254,8 +253,8 @@ class PrivateSafeTestVersion1(BaseSafeTestVersion1):
 register_tests(SafeTestVersion1)
 # Try to register test with private "ensemble" annotator.
 try:
-    from modelgauge.private_ensemble_annotator_set import PRIVATE_ANNOTATOR_SET
+    from modelgauge.private_ensemble_annotator_uids import SAFETY_ANNOTATOR_V1_0_UID
 
-    register_private_annotator_tests(PRIVATE_ANNOTATOR_SET, "ensemble")
+    register_private_annotator_tests(SAFETY_ANNOTATOR_V1_0_UID, "ensemble")
```
superdosh, line 258: Nice! My PR will update this to "official" or something else if we come up with it, but it probably makes sense to merge yours first.

src/modelgauge/tests/security.py
```diff
@@ -79,10 +76,6 @@ def sut_options(self) -> SUTOptions:
             temperature=0.01,
         )
 
-    @classmethod
```
superdosh, line 82: Same comment here re: keeping an abstract method.

src/modelgauge/tests/security.py
```diff
@@ -108,7 +101,17 @@ def make_test_items(self, dependency_helper: DependencyHelper) -> List[TestItem]
         return test_items
 
     def measure_quality(self, item: SUTResponseAnnotations) -> Dict[str, float]:
-        return self.annotators.evaluate(item)
+        annotator_uid = self.get_annotators()[0]  # These tests always only have 1 annotator.
```
superdosh, line 104: And same comment here about `get_annotator`.

src/modelgauge/tests/safe_v1.py
bkorycki: What's the issue with making a call to a method that is expected to be defined by subclasses?

Regarding the annotators; I'm hesitant to optimize and refactor the modelgauge tests right now because pretty soon they will be merged with the hazards and the test code will be removed! If that wasn't the case then yes, I think it would be reasonable to assume that all tests have one annotator after your ensembles-are-annotators change.

src/modelgauge/tests/safe_v1.py
bkorycki: I started a discussion in the chat

src/modelgauge/tests/safe_v1.py
superdosh: Sorry, my issue isn't with making the call. I think we should declare the expectation though (via abstractclass) like:

```
@abstractmethod
@classmethod
def get_annotators(cls):
    pass
```

And reason is same as the reason for other uses of abstract method:

* The code assumes a contract (that `cls.get_annotators()` is defined), so it's better to declare that contract.
* `abstractmethod` documents the intent to future developers. Declaring it as an abstractmethod will mean the developer will get an error on instantiation without defining the method. Without it, we only get the error when `measure_quality` is called.

Makes sense re: merging tests / hazards!!

bkorycki: Ohh got it! That method is already defined as an abstractmethod in `BaseTest`, but I can re-declare it in this base class as well for clarity:)

superdosh: Ah hah! Missed this, all looks good then!

https://github.com/mlcommons/modelbench/pull/1344/files#diff-497b6d1e33914c2f5bc07123c6095cc7624030a50a37dadb3b7f75e400a3e380L46

PR #1352, opened 2025-10-15
Invalid annotations are not counted

bkorycki:
Confirmed that this will not stop the benchmark run.

Should probably re-calibrate (pulling from the cache).


superdosh (approved):
LGTM!

PR #1359, opened 2025-10-20
Update python requirement

superdosh:
Also: fix bug where certain hugging face errors were silently ignored
Also: no longer require secrets file to exist
Also: allow EnsembleAnnotator to have secrets directly injected

https://github.com/mlcommons/sugar/issues/192


superdosh:
@bkorycki @rogthefrog sorry for the noise; this is now ready to review. And it works with the latest cheval updates here: https://github.com/mlcommons/sugar/pull/202

PR #1364, opened 2025-10-23
Baas sut

bkorycki:
This was mostly William's code. I'm opening as a draft PR because I have a few questions (TODOs in the code) that I would like feedback on.

### DIFF
--- a/src/modelgauge/suts/indirect_sut.py
+++ b/src/modelgauge/suts/indirect_sut.py
@@ -107,7 +107,7 @@ class IndirectSUTServer:
                     continue
             if len(bad_ids):
                 raise fastapi.HTTPException(
-                    status_code=404,
+                    status_code=207,
                     detail=f"Could not find request IDs: {bad_ids}. Ignoring responses for those requests.",
                 )
 
--- a/tests/modelgauge_tests/sut_tests/test_indirect_sut.py
+++ b/tests/modelgauge_tests/sut_tests/test_indirect_sut.py
@@ -102,7 +102,7 @@ class TestServer:
     def test_responses_bad_request_id_error_response(self, client, server, sut_request):
         response_data = [{"request_id": 42, "response": "boo"}]
         client_response = client.post("/responses", json=response_data)
-        assert client_response.status_code == 404
+        assert client_response.status_code == 207
         assert server.queues.get(42) is None
 
     def test_responses_not_all_bad_requests(self, client, server, sut_request):
@@ -112,7 +112,7 @@ class TestServer:
         ]  # First one bad, second one good.
         server.queues[1] = Queue()
         client_response = client.post("/responses", json=response_data)
-        assert client_response.status_code == 404
+        assert client_response.status_code == 207
         assert server.queues.get(42) is None
         assert server.queues[1].qsize() == 1
         assert server.queues[1].get() == IndirectSUTResponse(request_id=1, response="ok")


src/modelgauge/sut_factory.py
```diff
@@ -29,6 +30,7 @@ class SUTType(Enum):
 DYNAMIC_SUT_FACTORIES: dict = {
     "hf": HuggingFaceSUTFactory,
     "hfrelay": HuggingFaceSUTFactory,
+    "indirect": IndirectSUTFactory,  # TODO: Is this a good name?
```
superdosh: I think so.

src/modelgauge/suts/indirect_sut.py
```diff
@@ -0,0 +1,132 @@
...
+@modelgauge_sut(capabilities=[AcceptsTextPrompt])
+class IndirectSUT(PromptResponseSUT):
+
+    def __init__(self, uid: str):
+        super().__init__(uid)
+        self._server = IndirectSUTServer(7777)
+        self._id_generator = ThreadsafeIdGenerator()
+
+    def is_ready(self) -> ReadyResponse:
+        # TODO: should we actually do a readiness check?
```
superdosh: Maybe a reasonable check might be for the IndirectSUTServer to implement a healthcheck and for this to all it.

src/modelgauge/suts/indirect_sut.py
```diff
@@ -0,0 +1,132 @@
...
+        return ReadyResponse(True)
+
+    def translate_text_prompt(self, prompt: TextPrompt, options: SUTOptions) -> IndirectSUTRequest:
+        # TODO handle options
```
superdosh: What was the conclusion from the conversation here? I thought there was a thought that we wouldn't care about options here. Maybe we log if options are provided, because the assumption would be that they're ignored?

src/modelgauge/suts/indirect_sut.py
```diff
@@ -0,0 +1,132 @@
...
+class IndirectSUTServer:
+    def __init__(self, port: int):
+        self.port = port
+        self.prompts = []
+        self.queues: dict[int, Queue] = {}
+        self.outstanding_requests: dict[int, IndirectSUTRequest] = {}
+        self.app = self.make_app()
+
+    def make_app(self):
+        # TODO: Auth?
```
superdosh: If I understood correctly, the indirect SUT is instantiated by modelbench and the baas server posts to it? Is that right? Either way, if it's all internal services, we should have strict network rules on who can access, so I'm not sure we need any other auth.

src/modelgauge/suts/indirect_sut.py
```diff
@@ -0,0 +1,132 @@
...
+        # Question: Is it weird to use FastAPI here and connexion for BaaS?
```
superdosh: IMO, doesn't really matter. Maybe there's a slight switching cost if you're working on both simultaneously (not sure that's a real case), but these frameworks both seem very readable and non-arcane.

src/modelgauge/suts/indirect_sut.py
```diff
@@ -0,0 +1,132 @@
...
+                "queues": {n: str(q) for n, q in self.queues.items()},
+            }
+
+        @app.get("/requests")
+        def get_requests() -> list[IndirectSUTRequest]:
+            return list(self.outstanding_requests.values())
+
+        @app.post("/responses")
+        def post_responses(responses: list[IndirectSUTResponse]):
+            # TODO: What happens if there's no queue for a request ID?
```
superdosh: This should never happen in normal usage since any response has to have come from an outstanding_request, and a soon as something was put in outstanding request, a queue was created below.

Nevertheless, if the caller has a bug, I'm always a fan of putting a more useful error message than a vague KeyError that'll get generated by default.

src/modelgauge/sut_factory.py
wpietri: Just to throw out options: BaasSut, ListeningSut, PassiveSut.

src/modelgauge/suts/indirect_sut.py
wpietri: Is there anything on the server that could break to put in the health check? The only thing I could think of is that the port is occupied. But in that case I think it should just blow up on instantitation.

src/modelgauge/suts/indirect_sut.py
wpietri: I think we should pass them along. If they choose to disregard them, that's one thing. But I say we pass them along in some OpenAI-compatible structure in the hopes that they'll forward them. That way if we need to use them in the future, we'll be covered.

src/modelgauge/suts/indirect_sut.py
wpietri: I'm not averse to adding auth here, but I can't think of a threat model that would require it.

src/modelgauge/suts/indirect_sut.py
wpietri: I think it's a little weird we're adding a new way to do APIs. I haven't properly looked at Connexion yet, but we should be sure that its benefits, which I'm still hazy on, are worth the cost of having two tools for the same thing.

src/modelgauge/suts/indirect_sut.py
wpietri: I think we should definitely catch that case and log it. I can see two possible causes: they are resending a response or they're sending a response with the wrong id. Ideally, the first one would be quietly dropped (idempotent) and the second would give an error. So maybe we keep a set of requested or already-received request_ids somewhere?

The error response is a little tricky because this allows for small batches. Not sure what best practice is for saying "those 10 are good and these 2 are bad".

src/modelgauge/suts/indirect_sut.py
superdosh: Ah yah, that's nice.

bkorycki: That makes sense. I added the SUT options and changed the request structure to something OpenAI-compatible (I think).

src/modelgauge/suts/indirect_sut.py
bkorycki: Is it possible that the server would silently encounter an issue on start up?

src/modelgauge/suts/indirect_sut.py
bkorycki: I added some error handling, let me know what you guys think.

src/modelgauge/suts/indirect_sut.py
```diff
@@ -28,13 +30,19 @@ def next(self):
             return self._last_id
 
 
+class ChatMessage(BaseModel):
+    content: str
+    role: str
+
+
 class IndirectSUTRequest(BaseModel):
```
superdosh: Maybe better:
```
from modelgauge.suts.openai_client import OpenAIChatRequest 

class IndirectSUTRequest(OpenAIChatRequest):
    request_id: int
```
Or maybe there's a better more canonical alternative we use to `OpenAIChatRequest` and we use that.

src/modelgauge/suts/indirect_sut.py
```diff
@@ -45,8 +53,9 @@ class IndirectSUTResponse(BaseModel):
 @modelgauge_sut(capabilities=[AcceptsTextPrompt])
 class IndirectSUT(PromptResponseSUT):
 
-    def __init__(self, uid: str):
+    def __init__(self, uid: str, model_name: str):
```
superdosh, line 44: Why do we need `model_name`? Is it different than `uid` for our case?

src/modelgauge/suts/indirect_sut.py
```diff
@@ -94,9 +110,26 @@ def get_requests() -> list[IndirectSUTRequest]:
 
         @app.post("/responses")
         def post_responses(responses: list[IndirectSUTResponse]):
-            # TODO: What happens if there's no queue for a request ID?
+            bad_ids = []
             for response in responses:
-                self.queues[response.request_id].put(response)
+                try:
+                    queue = self.queues[response.request_id]
+                    if queue.qsize() >= 1:
```
superdosh: I think the case where they resend a request isn't handled right. For example, if they send request A, and it's fully processed before they resend it, then we wouldn't hit this as the queue would once again be empty. Though maybe that can't practically happen since there's a thread waiting on the queue and will pop it right away? Something feels off though.

BTW, not totally familiar with fastapi, but don't we need to send back a 200 or 202 on success? Or is no exceptions going to send back a 200?

src/modelgauge/suts/indirect_sut.py
```diff
@@ -94,9 +110,26 @@ def get_requests() -> list[IndirectSUTRequest]:
...
+                        # Or should we warn them?
+                        continue
+                    queue.put(response)
+                except KeyError:
+                    if response.request_id not in self.completed_requests:
+                        bad_ids.append(response.request_id)
+                    # else: Already completed; ignore.
+                    # Or should we warn them that this item is not going to be used?
+                    continue
+            if len(bad_ids):
```
superdosh, line 108: It's a bit strange to send a 404 here; I think 202 with the same detail is better here.

src/modelgauge/suts/indirect_sut.py
bkorycki: To be compatible with the OpenAI request structure, which expects model names. Here, model name would be something like `nvidia/model`, where as the UID would be `nvidia/model:indirect`.

src/modelgauge/suts/indirect_sut.py
bkorycki: I thought about using that! But then the request includes a ton of null fields:
```
[{"request_id":1,"model":"barbara/wee","messages":[{"content":"hello","role":"user"}],"max_completion_tokens":100,"temperature":null,"frequency_penalty":null,"logit_bias":null,"logprobs":null,"top_logprobs":null,"presence_penalty":null,"response_format":null,"seed":null,"stop":null,"stream":null,"top_p":null,"tools":null,"tool_choice":null,"user":null}]
```
Maybe there's a way to change how FastAPI serializes API responses to exclude None?

src/modelgauge/suts/indirect_sut.py
bkorycki: This if statement is just handling the case where they send duplicate requests (back to back for example) and the first request wasn't yet processed so it's still in the queue. Maybe it's unlikely, but not impossible?

If a request has already been processed, then the duplicate is ignored (line 127). 
Let me know if I'm misunderstanding your point.

bkorycki: And yes, it looks like FastAPI sends back 200 by default.

src/modelgauge/suts/indirect_sut.py
superdosh: Ah yah, weird. Not sure whether it's FastAPI or maybe pydantic that's doing the serialization. Either way, that's annoying.

The only thing that still bothers me here is that it feels weird for this to be defined here. In a sense, this request definition is important to our benchmark definition, not just indirect_suts. Maybe there's a higher level thing to be defined which both the `OpenAIChatRequest` and `IndirectSUTRequest` subclass.

(Not important to do now; just a thought.)

src/modelgauge/suts/indirect_sut.py
superdosh: Ah that makes sense!

src/modelgauge/suts/indirect_sut.py
superdosh: @bkorycki but would we get a KeyError in that case? Ah, looked more closer -- I see we delete the key here: https://github.com/mlcommons/modelbench/pull/1364/files#diff-53ad1398f00b94d9cad503d3b9fdf26f3e030964aa86f6d44ebbe6398f7d94d6R143 -- and so indeed we would get a KeyError for that case.

There's weird potential for races though.

wpietri: I agree that the `qsize` bit probably isn't helping, as the other end will process it immediately. I think there are three cases:
* no previous reply received, in which case enqueue
* reply received basically at the same time, either because of a duplicate in this batch or a simultaneous call, in which case we should enqueue and just let the other end drop the second one
* reply previously received and processed, in which case there should be no queue

So I think the right code is that if there's a queue, put it in; if not, discard either silently or with a complaint.

tests/modelgauge_tests/sut_tests/test_indirect_sut.py
```diff
@@ -94,11 +94,48 @@ def test_requests(self, client, server, sut_request):
...
+        ]  # First one bad, second one good.
+        server.queues[1] = Queue()
+        client_response = client.post("/responses", json=response_data)
+        assert client_response.status_code == 404
+        assert server.queues.get(42) is None
+        assert server.queues[1].qsize() == 1
+        assert server.queues[1].get() == IndirectSUTResponse(request_id=1, response="ok")
+
+    def test_responses_completed_request_id_ignores(self, client, server, sut_request):
+        response_data = [{"request_id": 1, "response": "ok"}]
```
wpietri, line 121: this series of tests knows a fair bit about the internals, which creates more coupling than I'd like. Is it possible to redo them just using the normal interface of the object?

src/modelgauge/suts/indirect_sut.py
wpietri: Yeah, or maybe a 207, multi-status.

src/modelgauge/suts/indirect_sut.py
```diff
@@ -109,9 +110,26 @@ def get_requests() -> list[IndirectSUTRequest]:
...
+                        # Already have a response from this batch; ignore duplicates
+                        # Or should we warn them?
+                        continue
+                    queue.put(response)
+                except KeyError:
+                    if response.request_id not in self.completed_requests:
+                        bad_ids.append(response.request_id)
+                    # else: Already completed; ignore.
+                    # Or should we warn them that this item is not going to be used?
+                    continue
```
wpietri, line 107: Does the `continue` do anything different here?

src/modelgauge/suts/indirect_sut.py
```diff
@@ -109,9 +110,26 @@ def get_requests() -> list[IndirectSUTRequest]:
...
+                    if queue.qsize() >= 1:
+                        # Already have a response from this batch; ignore duplicates
+                        # Or should we warn them?
+                        continue
+                    queue.put(response)
+                except KeyError:
+                    if response.request_id not in self.completed_requests:
+                        bad_ids.append(response.request_id)
+                    # else: Already completed; ignore.
+                    # Or should we warn them that this item is not going to be used?
```
wpietri: I think ignoring it is the right choice. Generally in HTTP APIs you want them to be idempotent to deal with the usual networking/distributed chaos: https://developer.mozilla.org/en-US/docs/Glossary/Idempotent

src/modelgauge/sut_factory.py
rogthefrog: +1 for ListeningSut

src/modelgauge/suts/indirect_sut.py
rogthefrog: The zero-trust / onion model would recommend adding auth.

src/modelgauge/suts/indirect_sut.py
rogthefrog: I think it's fine to do this internal API the way the others are done. Connexion's advantages are significant for external APIs. There's no difference with FastAPI under the hood.

src/modelgauge/suts/indirect_sut.py
bkorycki: No, I just put it there to be explicit

src/modelgauge/suts/indirect_sut.py
bkorycki: But what if the client sends two different responses for the same request ID? Isn't that technically a different request, meaning that idempotency doesn't apply?

tests/modelgauge_tests/sut_tests/test_indirect_sut.py
bkorycki: Shouldn't we also test the endpoints since that's what the baas API server will be interacting with?

src/modelgauge/suts/indirect_sut.py
bkorycki: Okay I found a way to inherit from `OpenAIChatRequest` and set `exclude_none` to True for FastAPI serialization!

Yeah, I think `TestItem` was intended to encapsulate the request definition you are describing. It has both the prompt and the `SUTOptions`, which are meant to be a SUT-agnostic description of the request.
The reason we have each SUT class also define its own request definition is to represent the request in the actual shape that is sent to that SUT's model/API/whatever. That's why we have the `translate_text_prompt` method, which transforms a request from a SUT-agnostic form to the SUT's native representation. But maybe widely-used SUT request types (e.g the OpenAI request) should be defined somewhere in a more central location in the code.

src/modelgauge/suts/indirect_sut.py
bkorycki: I'm going to leave auth out for now. We can add it later if we decide thats important.

src/modelgauge/suts/indirect_sut.py
wpietri: I can't think of an issue that would result in it opening a port but still not working.

src/modelgauge/suts/indirect_sut.py
bkorycki: updated

src/modelgauge/suts/indirect_sut.py
wpietri: Yes, that's true. In that case I'd fall back on [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle). If they've already given us an answer, let's assume the second one was a mistake, not worrying too much about the content of it.

If this does come up in practice and it matters to somebody, we can add more error detection then.

tests/modelgauge_tests/sut_tests/test_indirect_sut.py
wpietri: I think that's another good layer of tests. Unit tests serve as documentation of what the object's external behavior is supposed to be like. Integration tests and end-to-end tests make sure that larger systems keep working as we intend.

src/modelgauge/suts/indirect_sut.py
bkorycki: ok thanks!

PR #1369, opened 2025-10-28
ChevalAnnotator implementation

superdosh:
* Implements the ChevalAnnotator and registers it directly in modelbench. (If it can be instantiated, it means we can connect to the cheval at the config given.)
* Related cheval PR: https://github.com/mlcommons/sugar/pull/215

### DIFF
--- a/src/modelbench/cli.py
+++ b/src/modelbench/cli.py
@@ -15,7 +15,7 @@ from click import echo
 from rich.console import Console
 from rich.table import Table
 
-import modelgauge.annotators.cheval.registration # noqa: F401
+import modelgauge.annotators.cheval.registration  # noqa: F401
 from modelgauge.config import load_secrets_from_config, write_default_config
 from modelgauge.load_namespaces import load_namespaces
 from modelgauge.locales import DEFAULT_LOCALE, LOCALES


src/modelgauge/annotators/cheval/ids.py
superdosh, line 1: Note: I thought it'd be good to align these ids with the keys in the secrets.toml file, so I had to change the ids to get rid of the ".". Alternatively, we could keep `"safety_v1.1"` but then the toml file would need to look like:
```
["safety_v1.1"]
...
```
instead of 
```
[safety_v1_1]
...
```
I don't have a strong opinion.

src/modelgauge/annotators/cheval/annotator.py
```diff
@@ -0,0 +1,114 @@
...
+from modelgauge.annotation import EnsembleSafetyAnnotation, SafetyAnnotation
+from modelgauge.annotator import Annotator
+from modelgauge.annotators.cheval.request import AnnotationRequest
+from modelgauge.prompt import ChatPrompt, TextPrompt
+from modelgauge.retry_decorator import retry
+from modelgauge.secret_values import RequiredSecret, SecretDescription
+from modelgauge.sut import SUTResponse
+
+
+class ChevalAPIKey(RequiredSecret, ABC):
```
bkorycki: I don't think this is an abstract base class, right? Same for `ChevalEndpointUrl`

src/modelgauge/annotators/cheval/ids.py
bkorycki: I kind of prefer safety_v1.1. But it's not a strong opinion either.

src/modelgauge/cli_lazy.py
```diff
@@ -42,11 +43,13 @@ def resolve_command(self, ctx, args):
...
 
         annotator_uids = cmd_ctx.params.get("annotator_uids")
+        if not annotator_uids and cmd_ctx.params.get("annotator"):
+            annotator_uids = [cmd_ctx.params.get("annotator")]
         if annotator_uids:
             # try loading the private annotators
             try:
-                import modelgauge.private_ensemble_annotator_uids
-            except NotImplementedError:
+                import modelgauge.annotators.cheval.registration
```
bkorycki: How would the registration ever result in an exception? Same in modelbench/cli.py
My understanding is that now that we are moving to cheval, we don't need to do these try-excepts. The code doesn't need to do anything special for these cheval annotators. "Access" to these annotators will be determined when the code looks for any missing secrets. S

src/modelgauge/annotators/cheval/annotator.py
```diff
@@ -0,0 +1,114 @@
...
+
+    @classmethod
+    def for_provider(cls, provider):
+        cls.provider = provider
+        return cls
+
+    @classmethod
+    def description(cls) -> SecretDescription:
+        return SecretDescription(
+            scope=cls.provider,
```
bkorycki: All annotators use the same api key and endpoint, right? Would it be possible to just make the scope `cheval`?

src/modelgauge/tests/safe_v1.py
```diff
@@ -249,7 +249,7 @@ def get_annotators(cls) -> List[str]:
 register_tests(SafeTestVersion1)
 # Try to register test with private annotator.
 try:
-    from modelgauge.private_ensemble_annotator_uids import SAFETY_ANNOTATOR_V1_1_UID
+    from modelgauge.annotators.cheval.registration import SAFETY_ANNOTATOR_V1_1_UID
```
bkorycki: This doesn't need to be a try-except either :) I think there can be some major code simplification by getting rid of it. Same in the security test.

src/modelgauge/annotators/cheval/ids.py
wpietri: For snake case, I prefer underscores only. For kebab case, I'm ok with dots. Maybe for reasons that are obsolete, so I'm flexible. But it sounds like TOML has the same bias.

wpietri (approved):

src/modelgauge/annotators/cheval/annotator.py
superdosh: Definitely not; will remove!

src/modelgauge/annotators/cheval/ids.py
superdosh: NOTE: I'm going to change the cheval scope to "cheval" per Barbara's suggestion so it doesn't have to match these annotator ids at all, so the "." semantics of toml don't matter.

Options:
* safety_v1.1 (only thing it has going for it is its currently what cheval responds to, but I don't think this is a particularly strong reason as we don't use it anywhere yet)
* safety-v1.1 (consistent with William's comment on kebab-case) - my slight preference and currently in this PR
* safety_v1_1 (consistent with William's comment on snake_case)

I don't have a strong opinion, but I figure now's the easiest time to change it if we want!

@bkorycki @wpietri

src/modelgauge/cli_lazy.py
superdosh: This makes sense! What do you think about having this always registered? e.g. we could put the registration in modelgauge.annotators.__init__ or something? Or I could leave the import here?

src/modelgauge/tests/safe_v1.py
superdosh: Updated, let me know if that's what you had in mind! See here for just the commit: https://github.com/mlcommons/modelbench/pull/1369/commits/4bdc6cd5aa2be4dde37e76160464d1f08c5eb8df

src/modelgauge/cli_lazy.py
superdosh: See here for my take: https://github.com/mlcommons/modelbench/pull/1369/commits/4bdc6cd5aa2be4dde37e76160464d1f08c5eb8df

src/modelgauge/annotators/cheval/annotator.py
superdosh: Updated!

src/modelgauge/annotators/cheval/ids.py
bkorycki: I'm fine with the first or second option. I strongly dislike `safety_v1_1` lol

src/modelgauge/cli_lazy.py
bkorycki: I don't think we need `ensure_private_annotators_loaded()` at all!

src/modelgauge/cli_lazy.py
```diff
@@ -46,11 +46,8 @@ def resolve_command(self, ctx, args):
         if not annotator_uids and cmd_ctx.params.get("annotator"):
             annotator_uids = [cmd_ctx.params.get("annotator")]
         if annotator_uids:
-            # try loading the private annotators
-            try:
-                import modelgauge.annotators.cheval.registration
-            except ChevalAnnotatorError:
-                pass
+            ensure_private_annotators_loaded()
```
bkorycki: I don't think we need this-- it's impossible for the private annotators to not be "loaded"

src/modelgauge/cli_lazy.py
superdosh: But if that file isn't imported, we won't have the annotators registered, right?

src/modelgauge/cli_lazy.py
superdosh: If the import hasn't happened, the cheval annotators aren't registered, I think. Maybe there's a better place to do the registration so they're always registered?

bkorycki: Ah I forgot that we made changes to the registry/imports. Can we just handle the private annotators the same way as any other annotator? I don't remember where all those imports happen, but I don't think we should be doing anything special/different for the private annotators.

superdosh: Ah good idea; let me check that.

src/modelbench/cli.py
```diff
@@ -15,6 +15,7 @@
 from rich.console import Console
 from rich.table import Table
 
+import modelgauge.annotators.cheval.registration # noqa: F401
```
superdosh: @bkorycki I don't think I can get rid of this import because we only did the lazy loading / annotator registry thing in modelgauge, not modelbench, but I changed it so we always load it since it doesn't cost much to load!

src/modelgauge/annotators/cheval/ids.py
superdosh: Going with `safety-v1.1` :)

src/modelbench/cli.py
bkorycki: Ah bummer. Sounds good. Can we get rid of `ensure_private_annotators_loaded` if possible?

superdosh: I think I did?

PR #1372, opened 2025-10-30
Huge cache

bkorycki:


src/modelbench/cache.py
```diff
@@ -10,7 +10,7 @@
 CACHE_HITS = PROMETHEUS.counter("mm_cache_hits", "Cache hits", ["name"])
 CACHE_SIZE = PROMETHEUS.gauge("mm_cache_size", "Cache size", ["name"])
 
-MAX_CACHE_SIZE = 3 * 2**30  # 3 GB
```
wpietri, line 13: Here's a good example of the danger of duplicative comments. I'd say just remove the comment.

superdosh: Potentially a still useful comment would be something like `2**30 = 1 GB` so someone could read it quickly? Unless that's obvious to folks.

bkorycki: Ah good catch

PR #1381, opened 2025-11-11
OpenAI Base Url as secret

superdosh:
Allows passing the base url at runtime in cheval.


superdosh:
> Is this actually a secret? Or just a way to easily make something configurable? I'm fine either way, but if the latter we should figure out a better config story.

Unlikely this is a real secret (though maybe we do want the vllm ip to be obscured?). 

Agree on the better config story! Ideally one that unifies how we deal with actual secrets and these without having a secrets.toml (hacked or otherwise).

Issue #1406, opened 2025-12-03
Add benchmark version to results.json


rogthefrog:
The benchmark version only exists as part of UIDs in results.json. We should add it as its own key.

PR #1407, opened 2025-12-05
Add benchmark version to result json.

superdosh:
For #1406

### DIFF
--- a/src/modelbench/record.py
+++ b/src/modelbench/record.py
@@ -103,8 +103,9 @@ class BenchmarkScoreEncoder(json.JSONEncoder):
                 del result["_scoring_log"]
             return result
         elif isinstance(o, BenchmarkDefinition):
-            benchmark_version = o.get_uid_part("version")
-            return {"uid": o.uid, "hazards": o.hazards(), "version": benchmark_version}
+            def_dict = {"uid": o.uid, "hazards": o.hazards()}
+            def_dict.update(o.uid_dict)
+            return def_dict
         elif isinstance(o, HazardDefinition):
             result = {"uid": o.uid, "reference_standard": o.reference_standard()}
             if o._tests:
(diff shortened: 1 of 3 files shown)


src/modelbench/record.py
```diff
@@ -103,7 +103,8 @@ def default(self, o):
                 del result["_scoring_log"]
             return result
         elif isinstance(o, BenchmarkDefinition):
-            return {"uid": o.uid, "hazards": o.hazards()}
+            benchmark_version = o.get_uid_part("version")
```
superdosh: Do I need to handle the case where this raises a KeyError? Current version will raise that error, which seems fine to me, but want to confirm.

src/modelbench/record.py
```diff
@@ -103,7 +103,9 @@ def default(self, o):
                 del result["_scoring_log"]
             return result
         elif isinstance(o, BenchmarkDefinition):
-            return {"uid": o.uid, "hazards": o.hazards()}
+            def_dict = {"uid": o.uid, "hazards": o.hazards()}
+            def_dict.update(o.uid_dict)
```
bkorycki, line 107: Why do we need both the uid and the uid dictionary? Do they contain different information?

wpietri: The dictionary contains all the elements of the uid, but it doesn't explain how to turn the elements into the full string. So if anybody needs to compare to a string uid, they'll need either import modelbench or duplicate the logic. Plus with file format it's generally better to be backwards compatible, so that we don't have to hunt down and rewrite anything that parses the results JSON.

superdosh: As an example of what uid_dict looks like:
```
{'class': 'general_purpose_ai_chat_benchmark', 'version': '1.1', 'locale': 'en_us', 'prompt_set': 'practice', 'evaluator': 'default'}
```

PR #1409, opened 2025-12-08
Image build

bkorycki:
maybe?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Base Stage
-FROM python:3.10-slim AS base
+FROM python:3.12-slim AS base
 
 ENV PYTHONFAULTHANDLER=1 \
     PYTHONHASHSEED=random \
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -17,73 +22,75 @@ jobs:
...
+          TAGS_RESPONSE=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          -H "Accept: application/vnd.github.v3+json" \
+          "https://api.github.com/orgs/${{ github.repository_owner }}/packages/container/modelbench-private/versions" 2>/dev/null || echo "[]")
+          
+          if [ $? -ne 0 ] || [ -z "$TAGS_RESPONSE" ]; then
+          echo "Error: Failed to fetch package versions from GitHub API"
+          exit 1
+          fi
+          
+          if echo "$TAGS_RESPONSE" | jq -e '.message' >/dev/null 2>&1; then
```
rogthefrog, line 46: Wouldn't it be simpler to do the version string stuff in a small Python program?

bkorycki: I don't know, I copied this chunk directly from modelbench-private. I'm not eager to mess with it because I know it works.

PR #1413, opened 2025-12-10
first pass at outputting annotations

rogthefrog:
This generates the list of annotations for a given job, for consumption by modelrunner.

I still need to double check that the job ID is what we expect.

Questions:

* is this the right place to pull the annotations?
* is the source_id sufficient to reconcile these with ther matching prompts or responses?
* is there anything else that would be useful to output? 

Related to https://github.com/mlcommons/sugar/pull/303

### DIFF
--- a/tests/modelbench_tests/test_run.py
+++ b/tests/modelbench_tests/test_run.py
@@ -7,30 +7,38 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 from click.testing import CliRunner
+from modelgauge_tests.fake_sut import FakeSUT
 
 import modelbench
 from modelbench.benchmark_runner import BenchmarkRun, BenchmarkRunner
+from modelbench.benchmark_runner_items import ModelgaugeTestWrapper, TestRunItem
 from modelbench.benchmarks import (
     BenchmarkDefinition,
     BenchmarkScore,
     GeneralPurposeAiChatBenchmarkV1,
     NaiveBenchmark,
-    SecurityScore,
     SecurityBenchmark,
+    SecurityScore,
 )
 from modelbench.cli import cli
 from modelbench.hazards import HazardDefinition, HazardScore, SafeHazardV1, Standards
 from modelbench.scoring import ValueEstimate
 from modelbench.standards import NoStandardsFileError, OverwriteStandardsFileError
+from modelgauge.annotation import SafetyAnnotation
 from modelgauge.base_test import PromptResponseTest
-from modelgauge.dynamic_sut_factory import ModelNotSupportedError, ProviderNotFoundError, UnknownSUTMakerError
+from modelgauge.dynamic_sut_factory import (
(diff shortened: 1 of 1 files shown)


superdosh (commented):
I remember we had a conversation about how these annotations would be filtered to only those specifically marked as "exportable" in some way. Is that still correct?

wpietri (approved):
Looks good, and I like the choice of only doing demo for now. Should have some tests eventually, though, at least for the extractor.

src/modelbench/cli.py
```diff
@@ -185,21 +186,27 @@ def security_benchmark(
...
     output_dir.mkdir(exist_ok=True, parents=True)
     print_summary(benchmark, benchmark_scores)
     json_path = output_dir / f"benchmark_record-{benchmark.uid}.json"
     scores = [score for score in benchmark_scores if score.benchmark_definition == benchmark]
     dump_json(json_path, start_time, benchmark, scores, run_uid)
     print(f"Wrote record for {benchmark.uid} to {json_path}.")
+
+    # export the annotations separately
+    annotations = {"job_id": run.run_id, "annotations": run.compile_annotations()}
+    annotation_path = output_dir / f"annotations-{benchmark.uid}.json"
```
bkorycki, line 205: is it possible to include a timestamp in the file name? So that it doesn't get overwritten with every run.

rogthefrog: The results file doesn't have a timestamp. What is the reasoning for having one in annotations and not in results? I'm happy to add it, I'd just like to understand this better.

bkorycki: I think the results file should also have it haha its annoying that it gets overwritten.

rogthefrog: I see! I'm hesitant to just add it here, because there may be ramifications in other parts of the code. E.g. modelrunner finds the results file like this:

```python
    def results(self):
        try:
            results_files = self.records_path().glob("benchmark_record-*.json")
            file = next(results_files)
            with open(file, "rb") as f:
                data = json.load(f)
            return data
        except StopIteration:
            return None
```

i.e. it grabs whichever one is first in the `glob`. When there's only one file, it's always correct, but if there's more than one, we'll have to make sure they are sorted correctly, etc and that's not something I think we should deal with today :) I'll write a card to add it later.

rogthefrog: https://github.com/mlcommons/modelbench/issues/1415

bkorycki: okay that makes sense! thanks!

bkorycki (approved):

tests/modelbench_tests/test_run.py
wpietri: Seems fine to me, but I might rename the test to `test_benchmark_basic_run_produces_output` or something like that and then you can drop the todo.

PR #1422, opened 2025-12-17
Update dependencies

wpietri:
This updates a few libraries that required code changes on our end. The Mistral and Google updates changed how they did moderation-ish things, so I have removed the SUTs that depended on the old code. Per Kurt, next we run benchmarks those models will generally be too out of date to bother with.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1000,13 +1000,13 @@ tests = ["pytest"]
 
 [[package]]
 name = "fastapi"
-version = "0.124.4"
+version = "0.125.0"
 description = "FastAPI framework, high performance, easy to learn, fast to code, ready for production"
 optional = false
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -78,16 +81,15 @@ together = "^1.2.3"
 anthropic = "*"
 azure-ai-ml = "^1.22"
 google-api-python-client = ">=2.179.0"
-google-generativeai = "^0.8.0"
 google-genai = "^1.17.0"
 google-auth = "^2.36.0"
-mistralai = "1.6.0"
+mistralai = "1.9.11"
```
superdosh: We can't let this one float?

wpietri: Good point.

PR #1425, opened 2025-12-19
Make test sut options consistent and easier to access.

superdosh:


src/modelgauge/tests/safe_v1.py
```diff
@@ -99,6 +99,11 @@ class BaseSafeTestVersion1(PromptResponseTest, ABC):
     """
 
     hazards = Hazards()
+    _sut_options = SUTOptions(
```
rogthefrog, line 102: Is this something that could be in a superclass of the safe test and the security test, or is that too much of a twee "optimization"?

superdosh: It is, actually: https://github.com/mlcommons/modelbench/blob/main/src/modelgauge/base_test.py#L30

I wasn't sure we'd want the same default in the superclass though? I'm not sure; could be convinced.

wpietri (commented):
Nice! A good step forward.

superdosh:
Discussed live with @bkorycki and we're on the same page!

Issue #1385, opened 2025-11-14
A run's results needs to include who started the run


rogthefrog:
Add the name or identifier of the person who ran a job to the job's results json. It should already be in the journal and the modelrunner database.

PR #1438, opened 2026-01-07
Pass optional user to modelbench CLI

bkorycki:
First step of #1385

### DIFF
--- a/src/modelbench/cli.py
+++ b/src/modelbench/cli.py
@@ -172,7 +172,7 @@ def general_benchmark(
     sut_uid: str,
     locale: str,
     run_uid: str,
-    user: str,
+    user: str | None,
     prompt_set="demo",
     evaluator="default",
 ) -> None:
@@ -192,7 +192,7 @@ def security_benchmark(
     sut_uid: str,
     locale: str,
     run_uid: str,
-    user: str,
+    user: str | None,
     prompt_set="official",
     evaluator="default",
 ) -> None:


src/modelbench/cli.py
```diff
@@ -104,6 +104,12 @@ def decorator(func):
             required=False,
             help="The run_uid for the run if provided, otherwise one will be generated",
         )
+        @click.option(
```
superdosh, line 107: Can you remind me -- who's the caller that would set this? modelrunner?

src/modelbench/cli.py
```diff
@@ -104,6 +104,12 @@ def decorator(func):
...
+            "--user",
+            type=str,
+            required=False,
+            help="The user who ran this benchmark (metadata for record keeping).",
```
wpietri, line 111: Ooh, I love the help line. Great idea.

src/modelbench/record.py
```diff
@@ -47,11 +47,13 @@ def benchmark_library_info():
         return {"error": "pip not found"}
 
 
-def benchmark_metadata():
+def benchmark_metadata(user: str | None = None):
+    if user is None:
+        user = os.environ.get("USER", os.environ.get("USERNAME"))
```
wpietri, line 52: This surprised me a bit at first, as I was thinking of the user as only the externally injected name. But this makes sense for other runs. Good choice.

tests/modelbench_tests/test_record.py
```diff
@@ -319,3 +318,35 @@ def test_dump_json(benchmark_score, tmp_path, run_uid):
...
+    saved_user = os.environ.get("USER")
+    os.environ["USER"] = "me"
+    j = dump_and_read_record()
+    assert j["_metadata"]["run"]["user"] == "me"
+
+    # Dump correctly writes user from argument
+    j = dump_and_read_record(user="custom_user")
+    assert j["_metadata"]["run"]["user"] == "custom_user"
+
+    if saved_user is None:
```
wpietri: If this is to put the environment back the way it was found, it won't work if the test fails. So you'll probably want a try/finally block here.

src/modelbench/cli.py
bkorycki: Yes!

src/modelbench/record.py
superdosh, line 52: Could modelrunner (or other users) set `USER` or `USERNAME` to avoid having to add user as a command-line argument? I think I can see an argument that it's way less explicit and magical using env instead.

If the user doesn't provide `user` -- is there some `USER` env that's given by modelrunner's default? Is it the user running the modelrunner process?

src/modelbench/cli.py
```diff
@@ -166,13 +172,14 @@ def general_benchmark(
     sut_uid: str,
     locale: str,
     run_uid: str,
+    user: str,
```
superdosh: Might be nice to mark this as Optional in the type hinting. (Also below.)

tests/modelbench_tests/test_record.py
bkorycki: Ah good point!

src/modelbench/record.py
bkorycki: Hm, we could do it that way. But if multiple jobs are submitted at the same time from different users, could that introduce race conditions in the environment variable?

I wasn't planning on explicitly requiring the user to specify a `user` value in modelrunner.  It would just use the baas user in the baas context and the user running the job otherwise.

superdosh: I don't really have an opinion. I was just curious what we were doing before (meaning we were grabbing this env previously, was it at all meaningful?)

bkorycki: Yes it is meaningful! That envvar is set by unix systems, so when you run a benchmark locally the metadata stores your username:) Which is useful record-keeping if you ever put your local runs in DVC

superdosh: Oh right! That makes sense!

PR #1442, opened 2026-01-08
Dedicated hf dynamic SUT fixes

bkorycki:
4 bugs, 1 stone

### DIFF
--- a/src/modelgauge/suts/huggingface_sut_factory.py
+++ b/src/modelgauge/suts/huggingface_sut_factory.py
@@ -104,7 +104,7 @@ class HuggingFaceChatCompletionDedicatedSUTFactory(DynamicSUTFactory):
         try:
             endpoints = hfh.list_inference_endpoints()
             for e in endpoints:
-                if e.repository.lower() == model_name:
+                if e.repository.lower() == model_name.lower():
                     if e.status != "running":
                         try:
                             e.resume()


bollacker (approved):
LGTM

src/modelgauge/suts/huggingface_chat_completion.py
```diff
@@ -124,9 +124,10 @@ def translate_response(
 class HuggingFaceChatCompletionDedicatedSUT(BaseHuggingFaceChatCompletionSUT):
     """A Hugging Face SUT that is hosted on a dedicated inference endpoint and uses the chat_completion API."""
 
-    def __init__(self, uid: str, inference_endpoint: str, token: HuggingFaceInferenceToken):
+    def __init__(self, uid: str, inference_endpoint: str, model: str, token: HuggingFaceInferenceToken):
         super().__init__(uid, token)
         self.inference_endpoint = inference_endpoint
+        self.model = model
```
superdosh, line 130: Funny, we already include this for the serverless, where I suppose it's always been required?

src/modelgauge/suts/huggingface_sut_factory.py
```diff
@@ -104,23 +104,24 @@ def _find(sut_definition: SUTDefinition) -> str | None:
...
             for e in endpoints:
-                if e.repository == model_name and e.status != "running":
-                    try:
-                        e.resume()
-                    except Exception as ie:
-                        logger.error(
-                            f"Found endpoint for {model_name} but unable to start it. Check your token's permissions. {ie}"
-                        )
-                    return e.name
+                if e.repository.lower() == model_name:
```
superdosh: Do we need a `model_name.lower()` too?

bkorycki: It's already all lower but I can make it explicit here in case that ever changes in the future

src/modelgauge/suts/huggingface_chat_completion.py
bkorycki: I don't think it's always been required! I think some models require them and some don't. And we likely encountered this same issue with serverless in the past (since we use serverless more than dedicated), so then we added it to serverless sut requests. but not dedicated.

PR #1448, opened 2026-01-13
Incorporate modellogger

superdosh:
https://github.com/mlcommons/sugar/issues/307

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -191,13 +191,13 @@ files = [
 
 [[package]]
 name = "anthropic"
-version = "0.75.0"
+version = "0.76.0"
 description = "The official Python library for the anthropic API"
 optional = false
 python-versions = ">=3.9"
 files = [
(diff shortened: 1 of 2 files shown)


tests/modelgauge_tests/conftest.py
superdosh, line 1: @wpietri did you have to do something like this in cheval too? I wonder if maybe it's better handled in modellogger (e.g. we clear all but the caplog handler, if present?).

bkorycki: I personally think it makes more sense to handle it in modellogger (if possible) so that we don't have to repeat this code in our various projects.

bkorycki (approved):
👍

tests/modelgauge_tests/conftest.py
superdosh: Took a stab at it here: https://github.com/mlcommons/modellogger/pull/4

PR #1449, opened 2026-01-13
Prep for annotator logprobs

bkorycki:
Currently there is no ability to specifically request logprobs from annotators. You can request logprobs from SUTs by using the `top_logprobs` SUTOption. My plan is to use the same mechanism for annotators. This pull request doesn't actually implement annotator logprobs, just does some prep work.

The commit history is a good summary of everything that was done in this PR:)

### DIFF
--- a/src/modelgauge/model_options.py
+++ b/src/modelgauge/model_options.py
@@ -1,4 +1,4 @@
-from typing import Optional, List
+from typing import Optional, List, Sequence
 
 from pydantic import BaseModel, model_validator
 
@@ -70,3 +70,16 @@ class ModelOptions(BaseModel):
             options.top_logprobs = top_logprobs
 
         return options
+
+
+class TokenProbability(BaseModel):
+    """Probability assigned to a given token."""
+
+    token: str
+    logprob: float
+
+
+class TopTokens(BaseModel):
+    """List of most likely tokens and their probabilities."""
+
+    top_tokens: Sequence[TokenProbability]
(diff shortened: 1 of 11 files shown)


src/modelgauge/model_options.py
```diff
@@ -0,0 +1,85 @@
+from typing import Optional, List, Sequence
+
+from pydantic import BaseModel, model_validator
+
+
+class ModelOptions(BaseModel):
```
bkorycki, line 6: Everything in this file was moved from `sut.py`

src/modelgauge/model_options.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    An exhaustive set of options that could potentially be desired by a model.
+
+    Not all SUTs and annotators respect all options.
+    """
+
+    max_tokens: Optional[int] = None
```
bkorycki, line 13: This no longer defaults to 100. The CLI sets the default instead.

src/modelgauge/pipeline_runner.py
```diff
@@ -120,7 +118,9 @@ def _write_metadata(self):
 
 
 class PromptRunner(PipelineRunner):
-    def __init__(self, suts, **kwargs):
+    def __init__(self, suts, sut_options=ModelOptions(), **kwargs):
+        self.sut_options = sut_options
```
bkorycki, line 122: I moved this out of the `PipelineRunner` base class because it doesn't make sense for the annotator-only runner to have `sut_options`.

wpietri (approved):
Looks good to me. Thanks for breaking the commits down by purpose; that made this easier to follow.

PR #1453, opened 2026-01-15
Thinking SUTs

bkorycki:
The thinking logic used by the old together thinking sut (e.g. deepseek) is almost identical to that of nemotron 3 nano. So I was able to extract that logic to a mixin. Note that you can only access these thinking suts from the registry.

### DIFF
--- a/src/modelgauge/reasoning_handlers.py
+++ b/src/modelgauge/reasoning_handlers.py
@@ -20,9 +20,8 @@ class ThinkingMixin(PromptResponseSUT):
     """
     A mixin for SUTs that parses out thinking text from the output.
 
-    Reasoning is expected to be seperated from  content by a </think> tags.
-    If max_total_output_tokens is set in ModelOptions, that value will be used in the model call
-    and the content text will be truncated to max_tokens.
+    The output is expected to be in the form: {reasoning text}</think>{content text}.
+    If max_total_output_tokens is set in ModelOptions, that value will be used in the model call and the content text will be truncated to max_tokens.
     Otherwise, max_tokens is used in the model call and everything after </think> is returned as content.
     """
 


src/modelgauge/reasoning_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+            max_content_tokens=max_content_tokens,
+        )
+
+    def evaluate(self, request: ReasoningRequest) -> Any:
+        return super().evaluate(request.request)  # type: ignore
+
+    def translate_response(self, request: ReasoningRequest, response: Any) -> SUTResponse:
+        text = super().translate_response(request.request, response).text  # type: ignore
+
+        think_close = text.find("</think>")
```
wpietri: Could we make "</think>" a class or instance variable? It's duplicated quite a bit.

src/modelgauge/reasoning_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+        reasoning = text[: think_close + len("</think>")].strip()
+        content = text[think_close + len("</think>") :].strip()
+
+        reasoning_tokens = self.tokenizer.count_tokens(reasoning)
+        content_tokens = self.tokenizer.count_tokens(content)
+        reasoning_budget = request.request.max_tokens - request.max_content_tokens
+
+        if reasoning_tokens >= reasoning_budget and content_tokens >= request.max_content_tokens:
+            logger.warning(
+                f"SUT {self.uid} reasoning likely ate into the token budget of the actual output. Consider increasing max_total_output_tokens."
```
wpietri, line 77: shouldn't we also warn in the case where there's nothing but thinking text?

src/modelgauge/reasoning_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+        return ReasoningRequest(
+            request=request,
+            max_content_tokens=max_content_tokens,
+        )
+
+    def evaluate(self, request: ReasoningRequest) -> Any:
+        return super().evaluate(request.request)  # type: ignore
+
+    def translate_response(self, request: ReasoningRequest, response: Any) -> SUTResponse:
+        text = super().translate_response(request.request, response).text  # type: ignore
```
wpietri, line 52: This throws away anything else in the SUTResponse, which seems iffy to me. Currently that's only the logprobs, which in itself doesn't seem good, but it might not always be. Would it be better to just replace the text in the original SUTResponse?

Long term, I think we may also want to have something likes a ReasoningSutResponse, which keeps `text` as the text and adds a `reasoning` but no need to do that here.

src/modelgauge/reasoning_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+
+        think_close = text.find("</think>")
+        if think_close == -1:
+            # no closing tag: everything is thinking text
+            return SUTResponse(text="")
+
+        # Warn if reasoning was so large that remaining content had less than max_content_tokens available.
```
wpietri: In a longer method, a comment is often a sign of a smaller method waiting to be extracted.

tests/modelgauge_tests/test_reasoning_handlers.py
```diff
@@ -0,0 +1,96 @@
...
+        ],
+    )
+    def test_truncation(self, full_text, content_text, sut):
+        request = ReasoningRequest(request=FakeSUTRequest(text="", max_tokens=100), max_content_tokens=2)
+        response = FakeSUTResponse(text=full_text)
+
+        result = sut.translate_response(request, response)
+        assert result.text == content_text
+
+    def test_translate_response_warns_reasoning_over_budget(self, sut, caplog):
```
wpietri, line 106: Glad to see a test for the warning.

src/modelgauge/reasoning_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+        reasoning = text[: think_close + len("</think>")].strip()
+        content = text[think_close + len("</think>") :].strip()
+
+        reasoning_tokens = self.tokenizer.count_tokens(reasoning)
+        content_tokens = self.tokenizer.count_tokens(content)
+        reasoning_budget = request.request.max_tokens - request.max_content_tokens
+
+        if reasoning_tokens >= reasoning_budget and content_tokens >= request.max_content_tokens:
+            logger.warning(
```
wpietri, line 76: We don't do anything with warnings in benchmark runs right now. Should we?

wpietri (approved):
A solid start, but I think there are a few small things.

src/modelgauge/reasoning_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+
+from modelgauge.model_options import ModelOptions
+from modelgauge.prompt import TextPrompt
+from modelgauge.sut import SUTResponse, PromptResponseSUT
+from modelgauge.tokenizer import GeneralTokenizer
+
+logger = get_logger(__name__)
+
+
+class ReasoningRequest(BaseModel):
```
superdosh, line 13: Maybe dataclass instead of pydantic?

src/modelgauge/reasoning_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+    request: Any  # Request that is actually sent to the model.
+    max_content_tokens: int | None = None  # Number of tokens allowed for content (excluding thinking text).
+
+
+class ThinkingMixin(PromptResponseSUT):
+    """
+    A mixin for SUTs that parses out thinking text from the output.
+
+    Reasoning is expected to be seperated from  content by a </think> tags.
```
superdosh: Maybe also worth mentioning that <think></think> tags are expected to be before any content?

src/modelgauge/suts/huggingface_chat_completion.py
```diff
@@ -185,6 +186,14 @@ def translate_chat_prompt(self, prompt: ChatPrompt, options: ModelOptions) -> Hu
         )
 
 
+@modelgauge_sut(capabilities=[AcceptsTextPrompt, AcceptsChatPrompt])
+class HuggingFaceChatCompletionDedicatedThinkingSUT(ThinkingMixin, HuggingFaceChatCompletionDedicatedSUT):
```
superdosh, line 190: Why not mix this in to the serverless as well? The ThinkingMixin seems to work fine even if there is no thinking present. I can see the appeal of being explicit where we want to truncate thinking, but if our current policy is to always do it, then having it be general (and so not requiring registration) would be nice. Maybe if we did that, we'd rename it to `ThinkingTruncatorMixin` or similar?

src/modelgauge/reasoning_handlers.py
bkorycki: Throwing away logprobs was intentional. I could keep them, but I would want to discard the thinking portion of the logprobs, which is additional logic that I left out in favor of getting this done faster.

I think it's fine as is; SUTs that use the ThinkingMixin do not claim the `ProducesPerTokenLogProbabilities` capability. And we have some SUT capability verification so that if you run something that does require logprobs on a SUT that doesn't claim that capability, it will fail upfront. And our benchmarks don't request logprobs anyways.

The ReasoningSutResponse is a good idea!

src/modelgauge/reasoning_handlers.py
bkorycki: I don't think so. Kurt wanted this logged so that we can do some manual analysis later on.

src/modelgauge/reasoning_handlers.py
bkorycki: sure!

src/modelgauge/reasoning_handlers.py
bkorycki: Yes definitely. updated.

src/modelgauge/reasoning_handlers.py
bkorycki: All of our requests/responses are pydantic models so I wanted to be consistent. Also, I'm not sure if some of our serialization functionality depends on the objects being pydantic.

superdosh: Ah, makes sense!

src/modelgauge/reasoning_handlers.py
bkorycki: good idea!

src/modelgauge/suts/huggingface_chat_completion.py
bkorycki: Because different models handle reasoning in different ways! Not all reasoning models use the `</think>` formatting. Also some reasoning models accept thinking token budget options in the request. So as we encounter more reasoning SUTs, we will likely need to add more reasoning mix-ins. Thus, we can't use this `ThinkingMixin` as a default.

superdosh: Ah, I see. Because it's not a HF-level thing. Makes sense!

PR #1454, opened 2026-01-15
Fix dependencies causing intermittent errors

superdosh:
Intermittent test errors caused by:

* together requires typer<0.20, which resolves to 0.19.2
* huggingface_hub requires typer-slim (unpinned), which resolves to latest which is currently 0.21.1
* both typer and typer-slim install to a package called typer
* when poetry installs, it does so in parallel, and one clobbers the other, or leaves certain files in syntactically bad states (occasionally)

So here, we explicitly pin both typer and typer-slim in modelbench to 0.19.2 (latest <0.20). This kind of sucks, but I can't think of anything else that would work.

Also, unpin modellogger.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -547,17 +547,17 @@ uvloop = ["uvloop (>=0.15.2)"]
 
 [[package]]
 name = "boto3"
-version = "1.42.27"
+version = "1.42.29"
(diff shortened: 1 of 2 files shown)


wpietri (approved):
I propose another solution in chat, but I'm fine with this one if you think it better.

PR #1458, opened 2026-01-16
uv migration

superdosh:


tests/modelgauge_tests/data/install_pyproject.toml
superdosh, line 1: This didn't seem to be used anywhere.

publish_all.py
superdosh, line 1: I couldn't figure out whether we used this for anything?

README.md
```diff
@@ -43,46 +43,40 @@ a benchmark with fewer items via the `--max-instances` or `-m` flag.
...
+### Install `modelbench` with [`uv`](https://docs.astral.sh/uv/)
 
-1. Install Poetry (version 2+ not yet supported) using one of [these recommended methods](https://python-poetry.org/docs/#installation).  All of these methods require a recent version of python3 to be installed and available on your shell path.  For example:
-```shell
-pipx install poetry
-```
+1. Install `uv` using one of [these recommended methods](https://docs.astral.sh/uv/getting-started/installation/).
 
-2. Clone this repository.
+1. Clone this repository.
```
bkorycki, line 53: It looks like this list enumeration got messed up!

publish_all.py
bkorycki: I don't think we use this anymore.

README.md
superdosh: I changed this on purpose. Markdown renders it the same either way, and if you use all 1s, then if we insert steps in between, we don't have to renumber anything! See here: https://www.markdownguide.org/basic-syntax/#ordered-lists

Definitely a personal preference thing though, haha, so if we'd prefer the other way, happy to change it back.

.github/workflows/python-app.yml
```diff
@@ -15,26 +15,21 @@ jobs:
...
-    - name: Check poetry.lock file
-      run: poetry check
-
-    - name: Install dependencies
+    - name: Install uv
       run: |
-        set -e
-        poetry cache clear --no-interaction --all .
-        poetry install --no-interaction --with dev
+        curl -LsSf https://astral.sh/uv/install.sh | sh
```
dhosterman: I'm sure this is probably fine here, but I think we should prefer installation through something like pypi or a package manager over this. We might also want to pin a version, which we can do either way.

.github/workflows/python-app.yml
```diff
@@ -15,26 +15,21 @@ jobs:
...
+        echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+
+    - name: Check lock file and sync dependencies
+      run: uv sync --group dev --frozen
```
dhosterman: Do we want to add `--refresh` here to clear the cache like we were doing with poetry?

.gitignore
```diff
@@ -1,3 +1,5 @@
+modelbench.egg-info/
```
dhosterman, line 1: Might want to add this to .dockerignore also?

Dockerfile
```diff
@@ -6,36 +6,28 @@ ENV PYTHONFAULTHANDLER=1 \
...
 WORKDIR /app
 
 # Build Stage
 FROM base AS builder
 
-ENV PIP_DEFAULT_TIMEOUT=100 \
-    PIP_DISABLE_PIP_VERSION_CHECK=1 \
-    PIP_NO_CACHE_DIR=1 \
-    POETRY_VERSION=1.8.4
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh
```
dhosterman: See above re: installation method and pinning version.

I'm less excited about the "execute remote shell script" install method here than I am in CI/CD.

dhosterman (changes requested):
I think this looks generally fine! I've left a couple of comments and am specifically interested in ensuring the UV version is pinned.

I haven't tried it, and I'm not particularly confident with UV, but I see that you have and that it works!

.github/workflows/python-app.yml
superdosh: Did `--no-cache` instead, I think that's probably better -- https://docs.astral.sh/uv/reference/cli/#uv-sync--no-cache

.github/workflows/python-app.yml
superdosh: I'd like to leave it unpinned until it becomes an issue? Is that too dangerous? Don't want to end up in the situation we did in poetry where we ended up pinned to an old version forever.

dhosterman (approved):

PR #1468, opened 2026-01-26
update security prompt set name

rogthefrog:


wpietri (changes requested):
Does this change the definition of a benchmark? If so, the benchmark version number should change at the same time.

PR #1473, opened 2026-01-28
Fix docker image build

bkorycki:
Modelrunner expects modelbench to live in `venv` not `.venv`


bkorycki:
@superdosh no worries!

PR #1489, opened 2026-02-17
Upgrade Together and Uvicorn packages

dhosterman:
Upgrade Together and Uvicorn packages to satisfy dependabot.

The `--expensive-tests` don't all pass, but I don't think they have in quite a while.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,11 +109,6 @@ dev = [
 
 [tool.uv]
 package = true
-# together depends on typer and hfh depends on typer-slim. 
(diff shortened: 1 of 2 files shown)


superdosh:
Worth checking -- the typer-slim thing is supposedly fixed now, so in theory, removing the `exclude-dependency` for typer-slim should work.

The check would be whether after removing it, and re-locking, whether typer and typer-slim are both in uv.lock with different versions.

https://github.com/fastapi/typer/issues/1503#issuecomment-3884820285

(I'm happy to do it, if you're okay with me adding a commit here!)

PR #1521, opened 2026-05-01
Fix items with empty responses.

superdosh:
Prior fix only updated the item's measurement, not the underlying annotations. This does that and also journals it as needed.

The other change is that the update is applied at the end of `collect_annotations`, NOT in `handle_item` anymore, so the measurement is done on the corrected item.

### DIFF
--- a/src/modelbench/benchmark_runner.py
+++ b/src/modelbench/benchmark_runner.py
@@ -401,28 +401,17 @@ class TestRunAnnotationWorker(IntermediateCachingPipe):
 
     def fix_items_with_empty_responses(self, item):
         if item.sut_response.text is not None and item.sut_response.text.strip() == "":
-            # log if overall measurement is changed by this
-            if "is_safe" in item.measurements and item.measurements["is_safe"] == 0.0:
-                old_measurements = item.measurements.copy()
-                item.measurements["is_safe"] = 1.0
-                self.test_run.journal.item_entry(
-                    "overrode item quality",
-                    item,
-                    old_measurements=old_measurements,
-                    new_measurements=item.measurements,
-                )
-
-            overridden_annotations: list[str] = []
+            overridden_annotators: list[str] = []
             for annotator_uid, annotation in item.annotations.items():
                 if not annotation.is_safe:
                     annotation.is_safe = True
-                    overridden_annotations.append(annotator_uid)
+                    overridden_annotators.append(annotator_uid)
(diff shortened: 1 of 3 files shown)


bkorycki (approved):
looks good for an immediate fix.

src/modelbench/benchmark_runner.py
```diff
@@ -400,21 +399,20 @@ def handle_item(self, item: TestRunItem) -> TestRunItem:
...
+        if item.sut_response.text is not None and item.sut_response.text.strip() == "":
+            overridden_annotators: list[str] = []
+            for annotator_uid, annotation in item.annotations.items():
+                if not annotation.is_safe:
+                    annotation.is_safe = True
+                    overridden_annotators.append(annotator_uid)
+            # log if any annotations were overridden
+            if overridden_annotators:
+                self.test_run.journal.item_entry(
+                    "overrode item annotation",
```
rogthefrog, line 412: Does the consistency checker reference journal entries by label? If so, it may be good to check if changing "overrode item quality" to "overrode item annotation" will break anything there.

tests/modelbench_tests/test_benchmark_runner.py
```diff
@@ -653,31 +659,40 @@ def test_benchmark_annotation_worker_throws_exception(
...
-        assert measurement_entry["new_measurements"]["is_safe"] == 1.0
+        pipeline_item = TestRunItem(a_wrapped_test_with_goodyannotator, item_from_test, a_sut, sut_response)
+
+        # just collect annotations for overwriting, but don't measure yet
+        baw.collect_annotations(pipeline_item)
+        assert len(pipeline_item.measurements) == 0
+        # overwrite the annotations as unsafe
+        for annotator_uid in pipeline_item.annotations:
+            pipeline_item.annotations[annotator_uid].is_safe = False
+        # re-run the handle, which should populate the measurement (and fix)
```
rogthefrog, line 688: 👍🏻

rogthefrog (approved):

src/modelbench/benchmark_runner.py
superdosh: The only place I see "overrode" in the code base is here and in the test also in this PR. So I don't think anything else references it. That seems safe right?

superdosh:
Follow up here: https://github.com/mlcommons/sugar/issues/477

PR #1522, opened 2026-05-05
Simplified Prompt Engineered Annotator

superdosh:
Requested for use in modelplane https://github.com/mlcommons/modelplane/issues/131

The idea is for there to be a simple way to use our SUT machinery to set up new prompt engineered annotators. The existing prompt engineered annotator is more complex  (parameterized by labels, taxonomy, etc).


superdosh:
@bkorycki I still need to test this (and add unit tests), but wanted to get your feedback first on the design.

src/modelgauge/annotators/prompt_engineered.py
```diff
@@ -0,0 +1,79 @@
...
+        model_options=None,
+        sut_secrets: Optional[RawSecrets] = None,
+        **sut_kwargs,
+    ):
+        super().__init__(uid)
+
+        subs = prompt_template.get_identifiers()
+        if set(subs) != {"prompt", "response"}:
+            raise ValueError(
+                f"Prompt template must have 'prompt' and 'response' " f"placeholders (and no others), but got {subs}."
```
bkorycki: Does our current prompt template use the prompt? If not, we should make the prompt optional

bkorycki (commented):
I think I'm missing some context here. Are we moving the prompt engineered annotator from cheval back to modelbench?

The design looks good though!

src/modelgauge/annotators/prompt_engineered.py
superdosh: Addressed!

bkorycki (approved):
looks great!

Issue #1532, opened 2026-06-12
Composer bug fixes


superdosh:
Need to:
 - [x] Relax condition on parents to update metadata as long as there are no conflicts - https://github.com/mlcommons/modelbench/pull/1534
 - [x] Add helper to convert subset of columns to metadata column - https://github.com/mlcommons/modelbench/pull/1535
 - [x] Migrate frozenset bugfix to modelbench (cc @bkorycki ) - https://github.com/mlcommons/modelbench/pull/1533

PR #1533, opened 2026-06-12
Migrates cache key bug fix from modelplane.

superdosh:
Addresses #1532

Copies https://github.com/mlcommons/modelplane/pull/155 since code lives here now.


bollacker (approved):
LGTM

PR #1539, opened 2026-06-17
Security v1.0.1 definition

superdosh:
Also make future version updates easier.


src/modelbench/standards/security_naive_benchmark-1.0.1-en_us-official-private.json
superdosh, line 1: This will need to be updated when we finalize the actual evaluator. But this is just renamed for now to allow the tests to pass.

src/modelgauge/annotators/cheval/ids.py
```diff
@@ -1,2 +1,3 @@
 SAFETY_ANNOTATOR_V1_1_UID = "safety-v1.1"
 SECURITY_ANNOTATOR_V0_5_UID = "security-v0.5"
+SECURITY_ANNOTATOR_V1_0_UID = "security-v1.0"
```
superdosh: @rogthefrog this is the only thing we need to make sure we're in sync on.

src/modelbench/cli.py
```diff
@@ -200,10 +201,19 @@ def general_benchmark(
 
 
 @benchmark.command("security", help="run a security benchmark")
+@click.option(
```
bkorycki, line 204: Is this necessary if we are only supporting one version?

src/modelgauge/annotators/cheval/ids.py
bkorycki: maybe it would be simpler if the annotator version was aligned with the benchmark version?

src/modelbench/standards/security_naive_benchmark-1.0.1-en_us-official-private.json
bkorycki: Could we keep the old 1.0 file for comparison purposes? Also, it might be a good idea to put dummy reference scores in the 1.0.1 file (like 0.0 or 0.5) so that it's obvious that the numbers are not real.

bkorycki (commented):
The version constants are nice! that will make it much easier to make version updates in the future.

src/modelbench/cli.py
superdosh: Two things:
1. I believe it's needed by modelrunner, so we have to support the option (or modify modelrunner).
2. We want to be able to support multiple versions in the future per discussion in discord, so modifying modelrunner doesn't seem worth it.

src/modelbench/standards/security_naive_benchmark-1.0.1-en_us-official-private.json
superdosh: Oh good point, yah, I can do that.

src/modelgauge/annotators/cheval/ids.py
superdosh: Yah, I'd be up for this, but do we really want `security_v1.0.1` ? I'm okay with it either way!

src/modelbench/cli.py
bkorycki: Doesn't modelrunner already support the security benchmark? But either way, if need to support multiple versions in the future then nbd, might as well add it now.

src/modelgauge/annotators/cheval/ids.py
bkorycki: lol that's true, maybe that's a bit much. I'm also okay either way.

src/modelbench/cli.py
superdosh: Oh you're probably right on modelrunner support. Will leave it in though for future proofing!

src/modelgauge/annotators/cheval/ids.py
superdosh: @rogthefrog will leave it up to you :)

wpietri (approved):

src/modelgauge/annotators/cheval/ids.py
rogthefrog: If we're using 1.0.1 somewhere, I think we should use it everywhere, otherwise we'll wonder why they don't match. So my vote is for v1.0.1 (I've already named the cheval annotator `SecurityDAGV1_0_1Annotator` to match). We're not at Java levels of naming, so I'm ok with it. :D

superdosh: Updated!

src/modelgauge/annotators/cheval/ids.py
```diff
@@ -1,2 +1,5 @@
-SAFETY_ANNOTATOR_V1_1_UID = "safety-v1.1"
+from modelgauge.versions import CURRENT_GENERAL_VERSION, CURRENT_SECURITY_VERSION
+
+SAFETY_ANNOTATOR_V1_1_UID = f"safety-v{CURRENT_GENERAL_VERSION}"
```
rogthefrog, line 3: I like this!

src/modelgauge/annotators/cheval/ids.py
```diff
@@ -1,2 +1,5 @@
...
 SECURITY_ANNOTATOR_V0_5_UID = "security-v0.5"
+SECURITY_ANNOTATOR_V1_0_1_UID = f"security-v{CURRENT_SECURITY_VERSION}"
```
rogthefrog, line 5: And this!

bkorycki (approved):
looks great! thanks

src/modelbench/hazards.py
```diff
@@ -5,6 +5,12 @@
 import casefy
 from pydantic import BaseModel, ConfigDict
 
+from modelgauge.versions import (
+    CURRENT_GENERAL_V_UNDER,
```
rogthefrog, line 9: Nice.

rogthefrog (approved):
👏🏻

PR #1544, opened 2026-06-18
Prompt node bug

bkorycki:
AFAICT, it does not make sense to replace the original sut response context with the response from this node. This node's "sut" response is already stored in the output value.


superdosh (approved):
This makes sense. Original was built without realizing how we'd flow changes around.

superdosh:
I feel like maybe the only thing updateable should be metadata?

Issue #1353, opened 2025-10-15
Accurately report # errors in consistency check repoort


bkorycki:
If you run a consistency check on a run with some exceptions, it will show 0 errors in the table.

WatchTree-19:
root cause is in `_make_test_record` (benchmark_runner.py): `test_item_exceptions` is hardcoded to `[]`, so the exceptions collected on failed items never reach the `TestRecord`. `HazardScore.score()` then sums `len(test_record.test_item_exceptions)`, which is always 0, so the `# errors` column reads 0 even when items raised. happy to take this - fix populates `test_item_exceptions` from `run.failed_items_for(sut, test)` (same shape `simple_test_runner` already uses), with a regression test. will open a small PR.

superdosh:
@bkorycki can you confirm -- should # errors count all failed items or only recorded exceptions? It looks like we can have potentially have failures without recording the exception. For example, the `TestRunSutWorker` when it fails out records a failure on the item but doesn't append the exception. Is that considered a bug on the TestRunSutWorker or expected behavior?

@WatchTree-19 reviewed your PR, but my comments depend on the answer to the above question.

bkorycki:
@superdosh In my opinion, the # errors should represent the number of failed items. Maybe we should rename that column for clarity.

PR #1552, opened 2026-06-24
fix(benchmark_runner): record failed-item exceptions on TestRecord so…

WatchTree-19:
… # errors is accurate (#1353)

_make_test_record hardcoded test_item_exceptions=[], so exceptions captured on failed items never reached the TestRecord. HazardScore.score() sums len(test_record.test_item_exceptions) into HazardScore.exceptions, which feeds the '# errors' column of the results table, so it always read 0 even when items raised.

Populate test_item_exceptions from run.failed_items_for(sut, test), one TestItemExceptionRecord per failed item carrying exceptions (mirrors simple_test_runner). Adds a regression test plus a clean-run test.

### DIFF
--- a/src/modelbench/benchmark_runner.py
+++ b/src/modelbench/benchmark_runner.py
@@ -23,7 +23,7 @@ from modelgauge.config import raise_if_missing_from_config
 from modelgauge.monitoring import PROMETHEUS
 from modelgauge.pipeline import NullCache, Pipe, Pipeline, Sink, Source
 from modelgauge.pipeline_runner import PipelineRunner
-from modelgauge.records import TestRecord
+from modelgauge.records import TestItemExceptionRecord, TestRecord
 from modelgauge.single_turn_prompt_response import TestItem
 from modelgauge.sut import PromptResponseSUT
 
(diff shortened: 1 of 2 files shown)


WatchTree-19:
recheck

src/modelbench/benchmark_runner.py
```diff
@@ -540,6 +540,19 @@ def _calculate_test_results(self, test_run):
...
 
     def _make_test_record(self, run, sut, test, test_result):
+        # Collect the exceptions captured on failed items for this (sut, test) so they
+        # are recorded on the TestRecord. Previously this was hardcoded to an empty list,
+        # which meant downstream consumers (e.g. HazardScore.exceptions and the "# errors"
+        # column of the results table) always reported zero even when items errored.
+        test_item_exceptions = [
+            TestItemExceptionRecord(
+                test_item=item.test_item,
+                error_message=str(item.exceptions[-1]),
```
superdosh: Per discussion in #1353 , let's simplify this to remove `if exceptions` in the loop below, but for error_message and cause below, add something like `str(item.exceptions[-1]) if item.exceptions else ""`. 

This way we'll count all failed items and not exceptions-only.

superdosh (changes requested):
Thanks for this contribution! We've added some better clarity on the desired outcome in the card, and I've suggested a way to address below.

WatchTree-19:
thanks @superdosh - done. dropped the `if item.exceptions` filter so there's one record per failed item, and defaulted `error_message`/`cause` to `""` when an item has no captured exception. so `HazardScore.exceptions` and the `# errors` column now count all failed items, not exceptions-only. added a test for the no-exception case.

superdosh (approved):

PR #1553, opened 2026-06-24
Support passing metadata to cheval annotators.

superdosh:
See related PR: https://github.com/mlcommons/sugar/pull/568


bkorycki (approved):
Looks good! But why do we need to update both the ChevalAnnotator and DagAnnotator? Are they both used by security?

rogthefrog (approved):
👍🏻

src/modelgauge/annotators/cheval/request.py
```diff
@@ -1,7 +1,17 @@
...
 class AnnotationRequest(BaseModel):
     annotator: str
     prompt: str = ""
     response: str
+    metadata: dict = {}  # optional metadata to pass along with the request; used for benchmarking
+
+    def to_prompt(self) -> TextPromptWithMetadata:
+        return TextPromptWithMetadata(text=self.prompt, metadata=self.metadata)
+
+    def to_response(self) -> SUTResponse:
```
wpietri: A `to_x` method to me generally implies that it should contain the whole thing in form `x`. And `request.to_response` is confusing. Maybe `get_sut_response` if we want to put the construction here?

superdosh:
> Looks good! But why do we need to update both the ChevalAnnotator and DagAnnotator? Are they both used by security?

`ChevalAnnotator` makes sure the request sent by modelbench to cheval has the metadata. Then the change to `DAGAnnotator` is used within cheval to make sure the metadata is read into the `EvalContext`.

@bkorycki

src/modelgauge/annotators/cheval/annotator.py
```diff
@@ -92,6 +92,7 @@ def translate_prompt(self, prompt: TextPrompt | ChatPrompt, response: SUTRespons
             annotator=self.uid,
             prompt=prompt.text,
             response=response.text,
+            metadata=prompt.metadata if isinstance(prompt, TextPromptWithMetadata) else {},
```
wpietri: This is concerning to me. It's a sign that it's warping the existing object model, possibly because it's a convenient place to tuck things.

bkorycki:
@superdosh That makes sense!

src/modelgauge/annotators/cheval/request.py
```diff
@@ -1,7 +1,17 @@
...
 
+from modelgauge.prompt import TextPromptWithMetadata
+from modelgauge.sut import SUTResponse
+
 
 class AnnotationRequest(BaseModel):
     annotator: str
     prompt: str = ""
     response: str
+    metadata: dict = {}  # optional metadata to pass along with the request; used for benchmarking
```
wpietri: The comment here is suggests this is not really part of the object in a way that takes us away from a good domain model. The term metadata is also suspicious. We should at least figure out its true name, but probably also need a better structure.

src/modelgauge/prompt.py
```diff
@@ -22,3 +23,11 @@ class TextPrompt(BaseModel, frozen=True):
     """What actually goes to the SUT."""
 
     text: str
+
+
+class TextPromptWithMetadata(TextPrompt, frozen=True):
+    """TextPrompt with additional metadata. The
```
wpietri: As before, needing to explain like this is a sign we're going the wrong direction.

wpietri (commented):
I think this is going the wrong direction in a way that makes the domain model muddier and harder to maintain. If it's something we need to do for expediency, go for it, but please add a card for doing it right that we tackle as soon as the emergency is past.

superdosh:
> I think this is going the wrong direction in a way that makes the domain model muddier and harder to maintain. If it's something we need to do for expediency, go for it, but please add a card for doing it right that we tackle as soon as the emergency is past.

@wpietri I think there's a few changes I can make quickly that should alleviate many of the concerns. Will do a quick update / test and then ping you!

wpietri (approved):
This makes the problem clearer, which I think is a good step forward, but let's definitely come back to clean this up when we're past the rush.

superdosh:
> This makes the problem clearer, which I think is a good step forward, but let's definitely come back to clean this up when we're past the rush.

Added https://github.com/mlcommons/modelbench/issues/1555 with details as follow up!

PR #1562, opened 2026-06-25
Context for TestItems

superdosh:
#1560

### DIFF
--- a/src/modelgauge/tests/security.py
+++ b/src/modelgauge/tests/security.py
@@ -1,7 +1,9 @@
 import csv
+import json
 from abc import ABC, abstractmethod
 from typing import Dict, List, Mapping, Optional
 
+from airrlogger.log_config import get_logger
 from pydantic import BaseModel
 
 from modelgauge.aggregations import get_measurement_stats
@@ -34,6 +36,8 @@ from modelgauge.tests.safe_v1 import Hazards
 from modelgauge.test_decorator import modelgauge_test
 from modelgauge.test_registry import TESTS
 from modelgauge.versions import CURRENT_SECURITY_VERSION
+
+logger = get_logger(__name__)
 
 
 class SecurityTestResult(BaseModel):
@@ -123,11 +127,23 @@ class BaseSecurityJailbreakTest(BaseSecurityTest, ABC):
 
     def test_item_from_row(self, row) -> TestItem:
         hazard = self.hazards.get_hazard_from_row(row)
+        attack_type_id = row["attack_type_id"]
+        bijection_mapping = row["bijection_mapping"]
+        if bijection_mapping:
+            try:
+                bijection_mapping = json.loads(row["bijection_mapping"])
+            except:
+                logger.warning(f"Failed to load bijection_mapping from prompt row: {row["prompt_uid"]}.")
+        ctx = {
+            "hazard": hazard,
+            "attack_type_id": attack_type_id,
+            "bijection_mapping": bijection_mapping,
+        }
         return TestItem(
             prompt=TextPrompt(text=row["prompt_text"]),
             evaluated_prompt=TextPrompt(text=row["seed_prompt_text"]),
             source_id=row["prompt_uid"],
-            context={"hazard": hazard},
+            context=ctx,
         )
 
 
--- a/src/modelgauge/tests/security.py
+++ b/src/modelgauge/tests/security.py
@@ -127,23 +127,18 @@ class BaseSecurityJailbreakTest(BaseSecurityTest, ABC):
 
     def test_item_from_row(self, row) -> TestItem:
         hazard = self.hazards.get_hazard_from_row(row)
-        attack_type_id = row["attack_type_id"]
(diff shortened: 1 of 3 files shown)


superdosh:
Confirmed locally when running a benchmark that context is passed correctly from test item to annotator to nodes (and eventually to journal too).

Current keys are tied to column names in the official prompt set, which @bkorycki is also aligning in the DAG.

src/modelgauge/tests/security.py
```diff
@@ -123,11 +127,23 @@ class BaseSecurityJailbreakTest(BaseSecurityTest, ABC):
 
     def test_item_from_row(self, row) -> TestItem:
         hazard = self.hazards.get_hazard_from_row(row)
+        attack_type_id = row["attack_type_id"]
+        bijection_mapping = row["bijection_mapping"]
+        if len(bijection_mapping) > 0:
+            try:
+                bijection_mapping = json.loads(row["bijection_mapping"])
+            except:
+                logger.warning(f"Failed to load bijection_mapping from prompt row: {row["prompt_uid"]}.")
```
dhosterman: TIL, you can nest double quotes in f-strings.

tests/modelgauge_tests/test_security.py
```diff
@@ -22,9 +22,9 @@ def _make_prompts_file(content, tmp_path):
 @pytest.fixture(scope="module")
 def prompts_file_jailbreak(tmp_path_factory):
     content = (
-        f"prompt_uid,prompt_text,hazard,locale,seed_prompt_text\n"
-        "001,prompt 1,cse,en_US,seed 1\n"
-        "002,prompt 2,cse,en_US,seed 2\n"
+        f"prompt_uid,prompt_text,hazard,locale,seed_prompt_text,attack_type_id,bijection_mapping\n"
+        '001,prompt 1,cse,en_US,seed 1,bijection,{"a":"b"}\n'
```
dhosterman: Would this still work if the JSON were had multiple kv pairs? Would the CSV parser think it's another row?

dhosterman (approved):

tests/modelgauge_tests/test_security.py
superdosh: Ohh good question, I believe so, but will update the test to be sure!

src/modelgauge/tests/security.py
superdosh: Ha, didn't even notice I did that because the linter didn't see an issue.

tests/modelgauge_tests/test_security.py
superdosh: As the test was written, it failed. But I updated the test to match the format we actually have in our prompt file, and now it works!

src/modelgauge/tests/security.py
```diff
@@ -123,11 +127,23 @@ class BaseSecurityJailbreakTest(BaseSecurityTest, ABC):
 
     def test_item_from_row(self, row) -> TestItem:
         hazard = self.hazards.get_hazard_from_row(row)
+        attack_type_id = row["attack_type_id"]
+        bijection_mapping = row["bijection_mapping"]
```
rogthefrog, line 131: Can you do all that in one shot?

```python
try:
    bijection_mapping = json.loads(row["bijection_mapping"])
except:
    bijection_mapping = None
    logger.warning(...)
```

This is the third time I've wished for a `maybe` construct in Python this week :sob:

rogthefrog (approved):

bkorycki (changes requested):
This works for non-bijection jailbreaks, but the jailbreak decoder is a little weird and expects a different format:

"attack_type_id": {"name": "bijection", "mapping": ...}

src/modelgauge/tests/security.py
```diff
@@ -123,11 +127,23 @@ class BaseSecurityJailbreakTest(BaseSecurityTest, ABC):
...
+        if len(bijection_mapping) > 0:
+            try:
+                bijection_mapping = json.loads(row["bijection_mapping"])
```
bkorycki: Also the decoder expects the mapping to be a json string.

src/modelgauge/tests/security.py
superdosh: I thought so, but it's clearer your way regardless!

src/modelgauge/tests/security.py
superdosh: Ah that makes it easier!

superdosh:
> This works for non-bijection jailbreaks, but the jailbreak decoder is a little weird and expects a different format:
> 
> "attack_type_id": {"name": "bijection", "mapping": ...}

I think I got it now, or we can hold off on merging this if it's helpful for your testing! @bkorycki

bkorycki (approved):
yay! thanks! feel free to merge.

src/modelgauge/tests/security.py
superdosh: Turns out we didn't need this at all anyways, haha, so the code is gone now.

PR #1566, opened 2026-06-30
Dynamic dedicated together SUTs

bkorycki:


src/modelgauge/suts/together_sut_factory.py
```diff
@@ -1,16 +1,47 @@
...
+from modelgauge.dynamic_sut_factory import DynamicDriverSUTFactory, DynamicSUTFactory, ModelNotSupportedError
 from modelgauge.dynamic_sut_metadata import DynamicSUTMetadata
 from modelgauge.secret_values import InjectSecret, RawSecrets
+from modelgauge.sut import PromptResponseSUT
 from modelgauge.sut_definition import SUTDefinition
-from modelgauge.suts.together_client import TogetherChatSUT
+from modelgauge.suts.together_client import TogetherChatSUT, TogetherDedicatedChatSUT
+
+logger = get_logger(__name__)
+logging.getLogger("together_factory").setLevel(logging.ERROR)
```
superdosh: Shouldn't this be `together_sut_factory` ?

src/modelgauge/suts/together_sut_factory.py
superdosh, line 1: I think this can be simplified a good deal. We should move the common functions (which I believe is everything besides _find, and then we could have make_sut be a bit simpler:
```
try:
    name = self._serverless_find(...)
    return TogetherChatSUT(...)
except:
    try:
        name = self._dedicated_find(...)
        return TogetherDedicatedChatSUT(...)
    except:
        raise ...
```

We wouldn't need the other factory classes at all.

superdosh (approved):
What you have makes sense, but not very DRY, so some suggestions.

src/modelgauge/suts/together_sut_factory.py
bkorycki: Yeah, I think you're right. I was just copying the structure of the huggingface dynamic sut factory, but I like your suggestion better.

src/modelgauge/suts/together_sut_factory.py
```diff
@@ -22,37 +30,53 @@ def client(self) -> Together:
...
                 messages=[
                     {"role": "user", "content": "Anybody home?"},
                 ],
             )
+            return model
         except Exception as e:
-            raise ModelNotSupportedError(
-                f"Model {sut_metadata.external_model_name()} not found or not available on together: {e}"
-            )
+            logger.error(f"Error looking up serverless model {model} on together: {e}")
```
superdosh: Maybe warning or even info instead of error? (Both here and below for the dedicated message.)

superdosh (approved):
Looks great!

src/modelgauge/suts/together_sut_factory.py
```diff
@@ -1,19 +1,27 @@
...
 from modelgauge.auth.together_key import TogetherApiKey
 from modelgauge.dynamic_sut_factory import DynamicDriverSUTFactory, ModelNotSupportedError
-from modelgauge.dynamic_sut_metadata import DynamicSUTMetadata
 from modelgauge.secret_values import InjectSecret, RawSecrets
+from modelgauge.sut import PromptResponseSUT
 from modelgauge.sut_definition import SUTDefinition
-from modelgauge.suts.together_client import TogetherChatSUT
+from modelgauge.suts.together_client import TogetherChatSUT, TogetherDedicatedChatSUT
+
+logger = get_logger(__name__)
```
rogthefrog, line 14: Can we use the airr logger package? Or are we trying to keep that out of modelbench?

bkorycki: `get_logger` is being imported from the airlogger package

src/modelgauge/suts/together_sut_factory.py
bkorycki: I changed the serverless to info but I'm keeping the dedicated one as error because it would only be raised if an actual error occurred

superdosh: That makes sense!

PR #1567, opened 2026-07-02
Benchstalling

superdosh:
Addresses: #1565 

Issue was half open connections silently black-holed by GFE (via the Domain Mapping we have in front of Cloud Run for cheval).

TCP Keep Alive resolves the stalling.


dhosterman (approved):
Good find! I'm excited to see how much this helps with the stalling.

wpietri (approved):

PR #1575, opened 2026-07-15
Run current versions of requested benchmarks.

superdosh:
Follow up:

* If we want to support multiple versions in modelbench, we'd change the CLI to instantiate different benchmark definitions depending on the input version. This should be easy to add when we actually need to add it.
* Candela change -- note this change (with this PR) would ONLY be used for the naming of the benchmark job / run uid. modelbench would run the same thing regardless. https://github.com/mlcommons/sugar/pull/627 **NOTE:** it would be good to have a better strategy for sharing these versions so updating versions doesn't result in pain. Adding modelbench as a dependency is one option, but kind of painful.
* In modelrunner, BenchmarkJob.version defaults to 1.1 always. I think we should remove the default (or make the default configured per benchmark_type).

### DIFF
--- a/src/modelbench/cli.py
+++ b/src/modelbench/cli.py
@@ -16,13 +16,18 @@ from functools import wraps
 
 import click
 import termcolor
-from airrlogger.log_config import configure_logging
+from airrlogger.log_config import configure_logging, get_logger
 from click import echo
 from rich.console import Console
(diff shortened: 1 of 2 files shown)


superdosh:
Moving this back to draft as there's a good chance we don't want to do this.

PR #1581, opened 2026-07-21
Rev jailbreak --> 1.0.2

bkorycki:
This new benchmark version is the same except it uses a new annotator.


src/modelbench/standards/security_naive_benchmark-1.0.2-en_us-official-private.json
superdosh, line 1: When I created the file, you'd asked me to keep the copy of 1.0 for records. Do we want to keep 1.0.1 here also?

superdosh (approved):

src/modelbench/standards/security_naive_benchmark-1.0.2-en_us-official-private.json
bkorycki: Yeah, maybe that's a good idea!

dhosterman (approved):
👍

Issue #1579, opened 2026-07-20
Preference for dedicated endpoints in dynamic HF and Together factories


bkorycki:
The way our together and HF dynamically factories currently work is:

1) check if the model is available serverless. if so: use that model for the SUT
2) if not, check if we have a dedicated endpoint for the model.

There is no way to dynamically request a dedicated endpoint over a serverless one. 
This has been making updating our annotators more difficult because we *have* to use dedicated endpoints due to legal reasons.

PR #1582, opened 2026-07-21
HF and together prefer dedicated endpoints

rogthefrog:
https://github.com/mlcommons/modelbench/issues/1579

### DIFF
--- a/src/modelgauge/sut_definition.py
+++ b/src/modelgauge/sut_definition.py
@@ -2,9 +2,13 @@ import json
 import os
 import warnings
 from dataclasses import dataclass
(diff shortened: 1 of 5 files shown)


bkorycki (changes requested):
I don't think we *always* want to prefer dedicated or serverless at a factory level. Sometimes we prefer a together serverless sut, sometimes we prefer a together dedicated sut. And in the dedicated case, we typically don't want to fall back to serverless because it *has* to be dedicated sometimes for legal reasons.

So I think the preference needs to be set in the UID by the user, not by the factory.

PR #1583, opened 2026-07-22
Update to Together 2.0 endpoints

bkorycki:
This fixes the broken together dedicated endpoint SUTs by upgrading to the new together API version.

If you want to use together dedicated endpoints, you now must add the project ID to your secrets.toml file under "project_id" in the `[together]` scope. You can find this key in 1Password.

### DIFF
--- a/src/modelgauge/suts/together_client.py
+++ b/src/modelgauge/suts/together_client.py
@@ -1,5 +1,5 @@
 import time
-from dataclasses import dataclass
+from enum import StrEnum
 from typing import List, Optional
 
 import requests  # type: ignore
@@ -32,8 +32,7 @@ _ROLE_MAP = {
 }
 
 
-@dataclass
-class TogetherEndpointState:
+class TogetherEndpointState(StrEnum):
     STARTED: str = "DEPLOYMENT_STATE_READY"
     STOPPED: str = "DEPLOYMENT_STATE_STOPPED"
     ERROR: str = "DEPLOYMENT_STATE_FAILED"
--- a/src/modelgauge/suts/together_client.py
+++ b/src/modelgauge/suts/together_client.py
@@ -416,12 +416,7 @@ CHAT_MODELS = {
 for uid, model_name in CHAT_MODELS.items():
     SUTS.register(TogetherChatSUT, uid, model_name, InjectSecret(TogetherApiKey))
 
-DEDICATED_CHAT_MODELS = {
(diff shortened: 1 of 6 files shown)


src/modelgauge/suts/together_client.py
```diff
@@ -31,7 +32,18 @@
 }
 
 
-def _retrying_request(url, headers, json_payload, method):
+@dataclass
```
superdosh: Maybe this is more pythonic? https://docs.python.org/3/library/enum.html#enum.StrEnum

src/modelgauge/suts/together_client.py
```diff
@@ -377,13 +420,10 @@ def evaluate(self, request: TogetherChatRequest) -> TogetherChatResponse:
 for uid, model_name in CHAT_MODELS.items():
     SUTS.register(TogetherChatSUT, uid, model_name, InjectSecret(TogetherApiKey))
 
-DEDICATED_CHAT_MODELS = {
-    "mistral-8x22b-instruct-dedicated-together": "mlc_ai_safety_2/mistralai/Mixtral-8x22B-Instruct-v0.1-e1c7d251",
-    "gpt-oss-20b-dedicated-together": "mlc_ai_safety_2/openai/gpt-oss-20b-0fbd26f0",
-    "llama-3.3-70b-instruct-turbo-dedicated-together": "mlc_ai_safety_2/meta-llama/Llama-3.3-70B-Instruct-Turbo-8f91de15",
-    "gemma-4-26b-a4b-it-dedicated-together": "mlc_ai_safety_2/google/gemma-4-26B-A4B-it-c33cb205",
-}
+DEDICATED_CHAT_MODELS = {}
```
superdosh, line 422: This means we don't currently have any, right?

superdosh (approved):
LGTM!

wpietri (approved):
A good step forward, but I don't think a logger warning is sufficient to insure spin-down. What's the cost of accidentally leaving it up for a week? If it's significant, we should create some way to monitor this. If not as part of this card, then definitely as a new card.

src/modelgauge/suts/together_client.py
bkorycki: correct!

src/modelgauge/suts/together_client.py
bkorycki: updated

bkorycki:
@wpietri I added a card for monitoring

PR #1584, opened 2026-07-22
Allow the user to select dedicated or serverless on Together, Huggingface

rogthefrog:
... and remove the automated fallback.

https://github.com/mlcommons/modelbench/issues/1579


bkorycki (approved):
This looks great!

src/modelgauge/suts/together_sut_factory.py
superdosh, line 1: Will end up conflicting with @bkorycki 's PR.

superdosh (approved):
LGTM. I think this will break some notebooks maybe?

wpietri (approved):
Seems like a reasonable solution for users.

PR #1585, opened 2026-07-23
Migrate safety to dag

rogthefrog:
https://github.com/mlcommons/sugar/issues/562


src/modelgauge/annotators/cheval/ids.py
```diff
@@ -1,5 +1,6 @@
 from modelgauge.versions import CURRENT_GENERAL_VERSION, CURRENT_SECURITY_VERSION
 
 SAFETY_ANNOTATOR_V1_1_UID = f"safety-v{CURRENT_GENERAL_VERSION}"
+SAFETY_ANNOTATOR_V1_1_1_UID = "safety-dag-v1.1.1"
```
superdosh, line 4: IMO, we should update `CURRENT_GENERAL_VERSION` to 1.1.1 and leave `SAFETY_ANNOTATOR_V1_1_UID` everywhere and not introduce `SAFETY_ANNOTATOR_V1_1_1_UID`. I don't think we need to expose patch versions?

src/modelgauge/annotators/cheval/registration.py
```diff
@@ -1,16 +1,8 @@
 from modelgauge.annotator_registry import ANNOTATORS
 from modelgauge.annotators.cheval.annotator import ChevalAPIKey, ChevalAnnotator, ChevalEndpointUrl
-from modelgauge.annotators.cheval.ids import SAFETY_ANNOTATOR_V1_1_UID, SECURITY_ANNOTATOR_V1_0_2_UID
+from modelgauge.annotators.cheval.ids import SECURITY_ANNOTATOR_V1_0_2_UID
 from modelgauge.secret_values import InjectSecret
 
-ANNOTATORS.register(
```
superdosh, line 6: We import this registration file in the CLI as well (`modelbench.cli`) to make sure we can run that CLI. Any reason to separate the registrations?

src/modelgauge/annotators/safety_dag.py
```diff
@@ -0,0 +1,31 @@
...
+        api_key: ChevalAPIKey,
+        endpoint_url: ChevalEndpointUrl,
+    ) -> None:
+        cheval_annotator = ChevalAnnotator(
+            uid=SAFETY_ANNOTATOR_V1_1_UID,
+            api_key=api_key,
+            endpoint_url=endpoint_url,
+        )
+        dag = Composer(name=uid, verdict_type=Safety).add_node(
+            AnnotatorArbiter(name="cheval_safety", annotator=cheval_annotator)
```
superdosh, line 29: I'm not opposed to this, but I believe this will end up burying the information we want.

If we want to keep the output of the inner annotators, this wouldn't expose that, I think?

I was thinking we'd have a node for each inner annotator, and then a new node / arbiter that handles the ensemble logic.

superdosh:
I think this should happen in sugar, not modelbench.

src/modelgauge/annotators/cheval/ids.py
rogthefrog: I agree about patch versions. I put that in there so I could test without breaking everything else.

PR #1590, opened 2026-07-27
First stab at ensuring everything is versioned

dhosterman:
Versions are now tied to objects and dispatched from the CLI to the correct object/version. Run UID is confirmed to be respected.


src/modelbench/benchmarks.py
```diff
@@ -197,12 +205,15 @@ class GeneralPurposeAiChatBenchmarkV1(BenchmarkDefinition):
...
 
+    VERSION = "1.1"
+
     def __init__(self, locale: str, prompt_set: str, evaluator: str = "default"):
         validate_locale(locale)
         validate_prompt_set(GENERAL_PROMPT_SETS, prompt_set, locale)
         self.locale = locale
         self.prompt_set = prompt_set
         self.evaluator = evaluator
+        self.version = type(self).VERSION
```
bkorycki: why do we need both `self.version` and `cls.VERSION`?

dhosterman: Good catch! That was left over from a previous direction.

src/modelbench/benchmarks.py
```diff
@@ -281,18 +295,44 @@ def score(self, sut, hazard_scores, benchmark_end_time):
...
+
+def _version_sort_key(version: str) -> tuple[int, ...]:
+    return tuple(int(part) for part in version.split("."))
+
+
+def benchmark_versions_for(prefix: str) -> list[str]:
+    versions = {
+        getattr(cls, "VERSION")
+        for cls in _walk_subclasses(BenchmarkDefinition)
+        if cls.__name__.startswith(prefix) and hasattr(cls, "VERSION")
```
rogthefrog, line 342: Should we barf if a subclass doesn't have the version attribute?

src/modelbench/benchmarks.py
```diff
@@ -281,18 +295,44 @@ def score(self, sut, hazard_scores, benchmark_end_time):
...
         "class": "security_benchmark",
-        "version": CURRENT_SECURITY_VERSION,
+        "version": "self.VERSION",
         "locale": "self.locale",
         "prompt_set": "self.prompt_set",
         "evaluator": "self.evaluator",
     }
+
+
+def _walk_subclasses(cls: type) -> Iterator[type]:
```
rogthefrog, line 321: I get that this is future-proof and consistent with a lot of the code but I still kinda hate it.

On the flip side, I don't have a better idea.

src/modelbench/cli.py
```diff
@@ -406,10 +416,12 @@ def print_summary(benchmark, benchmark_scores):
     required=True,
 )
 def calibrate_cli(benchmark_type: str, locale: str, prompt_set: str, evaluator: str) -> None:
+    # TODO: add --version and dispatch to the correct benchmark class via
+    # benchmark_class_for() when multiple versions of a benchmark type exist.
```
rogthefrog, line 420: What's the signal that says there are multiple versions? Just the size of the subclass list?

src/modelgauge/annotators/cheval/ids.py
```diff
@@ -1,5 +1,4 @@
-from modelgauge.versions import CURRENT_GENERAL_VERSION, CURRENT_SECURITY_VERSION
-
-SAFETY_ANNOTATOR_V1_1_UID = f"safety-v{CURRENT_GENERAL_VERSION}"
+SAFETY_ANNOTATOR_V1_1_UID = "safety-v1.1"
 SECURITY_ANNOTATOR_V0_5_UID = "security-v0.5"
-SECURITY_ANNOTATOR_V1_0_2_UID = f"security-v{CURRENT_SECURITY_VERSION}"
+SECURITY_ANNOTATOR_V1_0_1_UID = "security-v1.0.1"
+SECURITY_ANNOTATOR_V1_0_2_UID = f"security-v.0.2"
```
rogthefrog: Doesn't need to be an f-string.

rogthefrog (approved):
Other than the metaprogramming, which I don't have a good alternative for, this looks reasonable!

src/modelgauge/annotators/cheval/ids.py
superdosh: Typo!

superdosh (changes requested):
Blocking for the typo!

src/modelgauge/annotators/cheval/ids.py
```diff
@@ -1,5 +1,4 @@
-from modelgauge.versions import CURRENT_GENERAL_VERSION, CURRENT_SECURITY_VERSION
-
-SAFETY_ANNOTATOR_V1_1_UID = f"safety-v{CURRENT_GENERAL_VERSION}"
+SAFETY_ANNOTATOR_V1_1_UID = "safety-v1.1"
```
superdosh: Maybe we should keep these as f-strings but grab the version from the `Benchmark` class attribute?

src/modelbench/cli.py
dhosterman: Yeah, probably? I'm unclear right now whether we want to move to:

1 - a benchmark class that is capable of multiple versions
2 - a single benchmark class per version

So this is a thing to figure out later, probably.

src/modelbench/benchmarks.py
dhosterman: Part of the reason I waffled so much on this was trying to figure out how to make adding new versions the least-worst thing and as safe as possible, so this version won out even if it is fairly indirect.

src/modelgauge/annotators/cheval/ids.py
dhosterman: I was originally doing that, but not all benchmarks were represented. I guess it's fine to remove the unrepresented ones as we're still operating under the assumption that the latest benchmark is the only one that'll be run. Check out the change and let me know if that makes sense.

superdosh: LGTM!

superdosh (approved):

src/modelgauge/annotators/cheval/ids.py
dhosterman: @superdosh , trying to implement this is revealing a convoluted dependency tree. It creates a weird dependency inversion where modelgauge is trying to import from modelbench.

I actually think hardcoding here is the right thing. Otherwise, what we should probably do is invert the dependency and have benchmarks use the annotator version as their VERSION.

What do you think?

src/modelbench/benchmarks.py
rogthefrog: 🧇

src/modelbench/benchmarks.py
dhosterman: @rogthefrog , I've put some code in the `BenchmarkDefinition` ABC to ensure that VERSION is present. It will catch this issue.

PR #1593, opened 2026-07-31
Dag handles n-ary routing

bkorycki:
New node object called `Router` which can route to multiple node paths depending on a route_map.
The `Gate` node is still there and is now a subclass of `Router`.


superdosh (approved):
This is nice! Love that almost all the old tests continue working here. I wonder if it's worth deprecating the idea of Gate eventually. Not sure the True/False routing versus any other routing really is a meaningful distinction anymore.

rogthefrog:
Awesome, and simple.

rogthefrog (approved):

PR #1600, opened 2026-08-05
Remove gate

bkorycki:
Simplification because `Gate` doesn't provide any value over `Router`.
Also added functionality to the `Router` that allows a user to specify a `default_branch` to use in the case that the output doesn't match any `route_map` keys.


dhosterman (approved):
👍

rogthefrog (approved):
Very nice.

PR #1601, opened 2026-08-05
annotator node, used to convert existing ensemble annotator to DAG shape

rogthefrog:
Story:

https://github.com/mlcommons/sugar/issues/562

Companion PR:

https://github.com/mlcommons/sugar/pull/636


bkorycki (approved):
Looks good, thanks! Can you put the `AnnotatorNode` in it's own file? I think we're using `nodes.py` for the node super types, and putting subclasses elsewhere e.g. `PromptEngineeredNode` is in `prompt_enricher.py`.

PR #1605, opened 2026-08-06
OpenAI Responses SUT

bkorycki:
This adds a new SUT based on OpenAI's new Responses API. This API is necessary to support some newer models (e.g. gpt-5.4-pro, gpt-5.5-pro). OpenAI also recommends Responses over Chat Completions for "new projects". 
I chose to keep the Chat Completions SUT because Nvidia NIM and ModelShip are dependent on this SUT. And Nvidia NIM is not compatible with the Responses API. But I did change the dynamic OpenAI SUT factory to create `Response` SUTs when `:openai` is in a uid. 

This PR also adds some handling for gpt models (`pro` models, anything >=`5.5`) that do not accept a `temperature` arg.  Before, these models would completely fail in benchmark runs because the benchmark passes a temperature value. I chose to hardcode a temp-acceptance-checker based on the model name. This is a little rudimentary, but I believe it is better than the other option: try to call the client, if it fails then replace the temp value in the request with `None` and retry. I don't like this option because requests should be immutable. It would be misleading if a run record says the request included `temp=0.01`, when in reality that value was actually replaced.

This PR should enable us to run all current text GPT models.

### DIFF
--- a/src/modelgauge/suts/nvidia_nim_api_client.py
+++ b/src/modelgauge/suts/nvidia_nim_api_client.py
@@ -5,7 +5,7 @@ from modelgauge.secret_values import (
     RequiredSecret,
     SecretDescription,
 )
-from modelgauge.suts.openai_client import OpenAIChat, OpenAIChatRequest
+from modelgauge.suts.openai_client import OpenAIChatSUT, OpenAIChatRequest
 from modelgauge.model_options import ModelOptions
 from modelgauge.sut_capabilities import (
     AcceptsChatPrompt,
@@ -39,7 +39,7 @@ class NIMOpenAIChatRequest(OpenAIChatRequest):
         AcceptsChatPrompt,
     ]
 )
-class NvidiaNIMApiClient(OpenAIChat):
+class NvidiaNIMApiClient(OpenAIChatSUT):
     """
     Documented at https://https://docs.api.nvidia.com/
     """
--- a/src/modelgauge/suts/nvidia_nim_api_client.py
+++ b/src/modelgauge/suts/nvidia_nim_api_client.py
@@ -47,8 +47,10 @@ class NvidiaNIMApiClient(OpenAIChat):
     def __init__(self, uid: str, model: str, api_key: NvidiaNIMApiKey):
         super().__init__(uid, model, api_key=api_key, base_url=BASE_URL)
 
-    def _translate_request(self, messages, options: ModelOptions, temp: float | None) -> NIMOpenAIChatRequest:
-        request = super()._translate_request(messages, options, temp)
+    def _translate_request_with_temperature(
+        self, messages, options: ModelOptions, temperature: float | None
+    ) -> NIMOpenAIChatRequest:
+        request = super()._translate_request_with_temperature(messages, options, temperature)
         request_json = request.model_dump(exclude_none=True)
         del request_json["max_completion_tokens"]  # NIM API doesn't allow extra inputs
         return NIMOpenAIChatRequest(
--- a/src/modelgauge/suts/nvidia_nim_api_client.py
+++ b/src/modelgauge/suts/nvidia_nim_api_client.py
@@ -47,8 +47,8 @@ class NvidiaNIMApiClient(OpenAIChat):
     def __init__(self, uid: str, model: str, api_key: NvidiaNIMApiKey):
         super().__init__(uid, model, api_key=api_key, base_url=BASE_URL)
 
-    def _translate_request(self, messages, options: ModelOptions) -> NIMOpenAIChatRequest:
-        request = super()._translate_request(messages, options)
+    def _translate_request(self, messages, options: ModelOptions, temp: float | None) -> NIMOpenAIChatRequest:
+        request = super()._translate_request(messages, options, temp)
         request_json = request.model_dump(exclude_none=True)
         del request_json["max_completion_tokens"]  # NIM API doesn't allow extra inputs
         return NIMOpenAIChatRequest(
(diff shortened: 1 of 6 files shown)


src/modelgauge/suts/nvidia_nim_api_client.py
```diff
@@ -47,8 +47,8 @@ class NvidiaNIMApiClient(OpenAIChat):
     def __init__(self, uid: str, model: str, api_key: NvidiaNIMApiKey):
         super().__init__(uid, model, api_key=api_key, base_url=BASE_URL)
 
-    def _translate_request(self, messages, options: ModelOptions) -> NIMOpenAIChatRequest:
-        request = super()._translate_request(messages, options)
+    def _translate_request(self, messages, options: ModelOptions, temp: float | None) -> NIMOpenAIChatRequest:
+        request = super()._translate_request(messages, options, temp)
```
rogthefrog: Any reason to name the argument `temp` instead of `temperature`? AFAIK we use `temperature` everywhere, and using a different arg name may be confusing.

src/modelgauge/suts/openai_client.py
```diff
@@ -117,40 +130,50 @@ def _load_client(self) -> OpenAI | None:
         else:
             return OpenAI(api_key=self.api_key, max_retries=7)
 
-    def translate_text_prompt(self, prompt: TextPrompt, options: ModelOptions) -> OpenAIChatRequest:
+    @staticmethod
+    def _check_accepts_temp(model: str) -> bool:
+        if "gpt" in model:
```
rogthefrog: Style thought: when there's branching like this, I usually try to return early to reduce the indentation. It may just be me, but I find it more readable because all the "if" branches are on the same level, like a switch/case. I.e. 

```python
def _check_accepts_temp(model: str) -> bool:
    if "gpt" not in model:
        return True
    if "pro" in model:
        return False
    try:
        version = float(model.split("-")[1])
    except ValueError:
        return True
    if version >= 5.5:
        return False
    return True
```

rogthefrog (approved):
👍🏻

src/modelgauge/suts/openai_client.py
```diff
@@ -117,40 +130,50 @@ def _load_client(self) -> OpenAI | None:
...
-            model=self.model,
-            frequency_penalty=options.frequency_penalty,
-            max_completion_tokens=options.max_tokens,
-            presence_penalty=options.presence_penalty,
-            stop=options.stop_sequences,
-            temperature=options.temperature,
-            top_p=options.top_p,
-            **optional_kwargs,
-        )
+    def _translate_request_temp(self, messages: List[OpenAIChatMessage], options: ModelOptions):
```
wpietri: Maybe it's just me, but I find the naming and calling here very confusing. The version of this method with temp doesn't take temp as a parameter, and the version without temp in the name does. It feels like an artifact of trying to cram something in. Maybe this could just be rewritten?

src/modelgauge/suts/openai_client.py
```diff
@@ -184,6 +250,57 @@ def translate_response(self, request: OpenAIChatRequest, response: ChatCompletio
...
 
 
+@modelgauge_sut(
+    capabilities=[
+        AcceptsTextPrompt,
+        AcceptsChatPrompt,
+        ProducesPerTokenLogProbabilities,
+    ]
+)
+class OpenAIResponses(BaseOpenAI):
```
wpietri, line 264: I understand the logic of how this name came to be, but I think it's confusing. It is not, as the name indicates, some responses from OpenAI. Similarly, OpenAIChat isn't actually a chat as we normally use the term. Are these maybe clients or APIs or something?

tests/modelgauge_tests/sut_tests/test_openai_client.py
```diff
@@ -1,252 +1,463 @@
...
-        client=client,  # type: ignore
-    )
+    def test_check_accepts_temp(self):
+        assert BaseOpenAI._check_accepts_temp("some-model") is True
+        assert BaseOpenAI._check_accepts_temp("gpt-4o") is True
+        assert BaseOpenAI._check_accepts_temp("gpt-5.4") is True
+        assert BaseOpenAI._check_accepts_temp("gpt-5.4-nano") is True
+        assert BaseOpenAI._check_accepts_temp("gpt-5.4-mini") is True
+
+        assert BaseOpenAI._check_accepts_temp("gpt-5.6") is False
```
wpietri: I think the intent of the code is to handle future model names, but the tests don't cover that.

wpietri (changes requested):
Progess for sure, but I think the code could be clearer and more DDD-aligned.

src/modelgauge/suts/openai_client.py
bkorycki: Yes, OpenAIChat uses the Chat completions API, OpenAIResponses uses the Responses API. Do you have better names in mind?

tests/modelgauge_tests/sut_tests/test_openai_client.py
bkorycki: I added some more tests!

src/modelgauge/suts/openai_client.py
bkorycki: Yeah you're right it was confusing before. I re-named the methods, is it more clear now?

src/modelgauge/suts/openai_client.py
bkorycki: I like that!

src/modelgauge/suts/openai_client.py
wpietri: Ah, sorry, I was trying to suggest a name. I'm suggesting it should be OpenAIResponsesClient or OpenAIResponsesAPI depending on intent.

bkorycki: What about `OpenAIResponsesSUT`?

wpietri: Sure, if you rename all the other things SUT too.

wpietri (approved):

PR #1606, opened 2026-08-07
Add ability to list dynamic SUTs

wpietri:
The dynamic SUTs stuff was a bit opaque to me. I added the ability to list dynamic SUT drivers and, where convenient, the available models.

The previous behavior, which shows all the preconfigured SUTs,  is at `modelbench list-suts static`.

Doing `modelbench list-suts dynamic` will get a list of available drivers. Adding on a driver name will in many cases get a list of models available via that driver.

One possible direction for expansion is adding per-driver help text to document the expected SUT uid format.

### DIFF
--- a/src/modelgauge/suts/aws_bedrock_sut_factory.py
+++ b/src/modelgauge/suts/aws_bedrock_sut_factory.py
@@ -1,4 +1,5 @@
 import os
+from typing import Optional
 
 import boto3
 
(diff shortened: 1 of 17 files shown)


src/modelgauge/suts/aws_bedrock_sut_factory.py
```diff
@@ -27,13 +28,10 @@ def client(self):
             )
         return self._client
 
-    def _convert_model_id(self, model_id: str) -> SUTDefinition:
-        """Convert AWS model IDs (maker.model[:version?]) to our standard format."""
-        maker, model_name = model_id.split(".", maxsplit=1)
-        model_name = model_name.replace(":", ".")
-        return SUTDefinition({"maker": maker, "model": model_name, "driver": self.DRIVER_NAME})
+    def list_suts(self) -> Optional[list[str]]:
+        return [f"{m.get('maker')}/{m.get('model')}" for m in self._get_available_models().values()]
```
rogthefrog, line 32: Won't this create a model name like `/mymodel` (and not `mymodel`) if the maker is blank, which is allowed?

bkorycki (commented):
This is nice, thanks!

Couple of suggestions:
- Maybe we can make `list_suts` an abstract method in the factory super class? Otherwise I think it will be easy to forget to implement this method in future factories.
- Right now it looks like `list_suts` is actually listing model names  instead of sut uids; I think we should either change the method to list the full sut UIDs, or change the method name to `list_models`.

src/modelgauge/suts/mistral_sut_factory.py
```diff
@@ -24,6 +28,10 @@ def get_secrets(self) -> list[InjectSecret]:
         api_key = InjectSecret(MistralAIAPIKey)
         return [api_key]
 
+    def list_suts(self) -> Optional[list[str]]:
+        model_list: ModelList = self.client.client.models.list()
+        return [f"{self.DRIVER_NAME}/{m.id}" for m in model_list.data]
```
rogthefrog, line 33: I'm confused here. The driver is the chunk after the model name in a SUT UID, e.g.

some_lab/some_model:driver_string

but it looks like it's prefixed to the model here?

dhosterman:
Nothing to add to the above, but wanted to say this is neat and I love it!

PR #1607, opened 2026-08-10
Change the exit code for consistency check failures to 3

dhosterman:
Because 2 is used to indicate syntax errors.

### DIFF
--- a/src/modelbench/consistency_checker.py
+++ b/src/modelbench/consistency_checker.py
@@ -636,4 +636,4 @@ def summarize_consistency_check_results(checkers: List[ConsistencyChecker]):
 
 
 class ConsistencyCheckError(Exception):
-    EXIT_CODE = 2
+    EXIT_CODE = 3


wpietri (approved):
Assuming there's a matching PR that migrates existing data, seems fine.

src/modelbench/consistency_checker.py
```diff
@@ -636,4 +636,4 @@ def summarize_consistency_check_results(checkers: List[ConsistencyChecker]):
 
 
 class ConsistencyCheckError(Exception):
-    EXIT_CODE = 2
+    EXIT_CODE = 3
```
rogthefrog, line 639: Does modelrunner need to be updated to reflect this change?

dhosterman: It does, and it is!

dhosterman:
> Assuming there's a matching PR that migrates existing data, seems fine.

Done!

PR #1608, opened 2026-08-10
fix(aggregations): guard mean_of_measurement against empty measurements

WatchTree-19:
### what

`mean_of_measurement` divides by `len(measurements)` with no empty guard:

```python
measurements = get_measurements(measurement_name, items)
total = sum(measurements)
return total / len(measurements)
```

so a test whose `aggregate_measurements()` receives no measured items - e.g. every item errored or was filtered out - raises `ZeroDivisionError` instead of producing a score, crashing aggregation/reporting.

its sibling in the same module, `MeasurementStats.calculate`, already guards this and returns mean `0` for an empty input. `mean_of_measurement` now does the same, keeping the two consistent.

### fix

return `0.0` when there are no measurements. adds regression tests (`test_mean_of_measurement` and `test_mean_of_measurement_no_measurements`) - the empty case raises `ZeroDivisionError` on `main` and passes with the change. black clean, DCO signed.

### DIFF
--- a/src/modelgauge/aggregations.py
+++ b/src/modelgauge/aggregations.py
@@ -70,5 +70,7 @@ def sum_measurements(measurement_name: str, items: List[MeasuredTestItem]) -> fl
 def mean_of_measurement(measurement_name: str, items: List[MeasuredTestItem]) -> float:
     """Calculate the mean across all TestItems for a desired measurement."""
     measurements = get_measurements(measurement_name, items)
+    if len(measurements) == 0:
+        return 0.0
     total = sum(measurements)
     return total / len(measurements)
--- a/tests/modelgauge_tests/test_aggregations.py
+++ b/tests/modelgauge_tests/test_aggregations.py
@@ -4,6 +4,7 @@ from modelgauge.aggregations import (
     get_measurement_stats,
     get_measurement_stats_by_key,
     get_measurements,
+    mean_of_measurement,
 )
 from modelgauge.prompt import TextPrompt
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek (commented):
Could an empty set remain distinguishable from a measured zero? Returning `0.0` means an all-error or all-filtered run becomes a valid zero measurement. For safety benchmarks that can make missing coverage look benign and bias aggregates. I would prefer `NaN` / `None`, or an explicit no-data result, rather than converting absence of evidence into a score.

sylvesterkaczmarek (changes requested):
Returning `None` resolves the semantic concern I raised, but current `10cd7577` fails mypy: `src/modelgauge/tests/demo_02_unpacking_dependency_test.py:71` still declares `aggregate_measurements() -> Dict[str, float]` while it returns `mean_of_measurement(...)`, now `float | None`. Please update that stale annotation (and any equivalent float-only aggregate annotation) so CI is green.

sylvesterkaczmarek (approved):
Re-reviewed current 7123538. The aggregate annotation now accepts Optional[float], matching mean_of_measurement returning None for an empty measurement set, and the current build is green. My previous typing and no-data concerns are resolved.

WatchTree-19:
Thanks for the re-review @sylvesterkaczmarek, and for pushing back on the 0.0 in the first place. Converting an all-error or all-filtered run into a valid zero was the wrong default for a safety benchmark and I would not have caught it.

Checks are green and this now needs a code owner from @mlcommons/ai-safety-engineers. Happy to rebase if it sits long enough to go stale.

PR #1609, opened 2026-08-11
Add actions permissions

dhosterman:
It looks like Github made some changes and now we have to be more specific about what the generated Actions token can do.

It looks like some extra white spaces/tabs were removed as well.


bkorycki (approved):
lgtm!

PR #1612, opened 2026-08-13
Google driver updates & listing dynamic SUTs

wpietri:
This gets the Google driver working with the models we'd like to benchmark.

It also includes the list-dynamic-suts code updated as per your feedback.


bkorycki (approved):
Looks great, thanks! Do we not need the modelship sut anymore?

wpietri:
If I recall rightly, modelship was my attempt to make models automatically
spin up and down on MLC2, but we ditched it in favor of static models.

PR #1615, opened 2026-08-21
Security practice prompt set

rogthefrog:
I created the airr_official_security_attack_1.0.1_heldback_en_us_prompt_set_release_one_per_hazard file in

https://ailuminate.mlcommons.org/files/

using 1 sample per hazard from the original prompt set.

https://github.com/mlcommons/sugar/issues/635

### DIFF
--- a/src/modelgauge/prompt_sets.py
+++ b/src/modelgauge/prompt_sets.py
@@ -45,7 +45,8 @@ SECURITY_JAILBREAK_PROMPT_SETS = {
 SECURITY_NAIVE_PROMPT_SETS = {
     "official": {
         "en_us": "airr_official_security_naive_1.0.1_heldback_en_us_prompt_set_release",
-    }
+    },
+    "practice": {"en_us": "airr_official_security_naive_1.0.1_heldback_en_us_prompt_set_release_one_per_hazard"},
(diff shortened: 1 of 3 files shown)


src/modelgauge/prompt_sets.py
```diff
@@ -37,12 +37,16 @@ def description(cls) -> SecretDescription:
 SECURITY_JAILBREAK_PROMPT_SETS = {
     "official": {
         "en_us": "airr_official_security_attack_1.0.1_heldback_en_us_prompt_set_release",
-    }
+    },
+    "practice": {
+        "en_us": "airr_official_security_attack_1.0.1_heldback_en_us_prompt_set_release_one_per_hazard",
```
wpietri: Is "heldback" is the correct term here? I think it should be "practice".

wpietri (changes requested):

src/modelgauge/prompt_sets.py
rogthefrog: I used "heldback" because the contents are from the heldback prompts, but I'll change it to practice.

bollacker (approved):
LGTM

PR #1619, opened 2026-08-21
Router bug

bkorycki:
Add the default path to router's all_path methods. Also update visualization to include the default path.


dhosterman (approved):
👍

bollacker (approved):
Looks good.  Maybe we should not have default routes.

PR #1626, opened 2026-08-26
Add support for safety DAG 1.2

rogthefrog:


dhosterman (approved):
👍

src/modelgauge/prompt_sets.py
```diff
@@ -34,6 +34,17 @@ def description(cls) -> SecretDescription:
         "fr_fr": "airr_official_1.0_demo_fr_fr_prompt_set_release",
     },
 }
+SAFETY_1_2_PROMPT_SETS = {
+    "practice": {
+        "en_us": "airr_official_1.2_practice_prompt_set_release_with_visibility",
+    },
+    "official": {
+        "en_us": "airr_official_1.2_heldback_prompt_set_release",
```
wpietri: Shouldn't this be `airr_official_safety_1.2_heldback_en_us_prompt_set.csv`?

src/modelbench/benchmarks.py
```diff
@@ -249,16 +250,43 @@ def _make_hazards(self) -> Sequence[HazardDefinition]:
...
+
+
+class GeneralPurposeAiChatBenchmarkV1_2(GeneralPurposeAiChatBenchmarkV1):
+    VERSION = "1.2"
+    PROMPT_SETS = SAFETY_1_2_PROMPT_SETS  # pyright: ignore
+
+    def _make_hazards(self) -> Sequence[HazardDefinition]:
+        return super()._make_hazards(SafeHazardV1_2)
+
+    def reference_benchmark(self) -> BenchmarkDefinition:
```
wpietri: Is this just a duplicate of 1.1?

wpietri (approved):

src/modelgauge/annotators/composer/empty_response.py
```diff
@@ -0,0 +1,12 @@
+from modelgauge.annotators.composer.context import EvalContext, NodeOutput
+from modelgauge.annotators.composer.nodes import Router
+
+
+class EmptyResponseNode(Router):
```
bkorycki, line 5: I probably missed something but why are we adding this? This is different from Riki's empty response node.

src/modelgauge/annotators/composer/empty_response.py
```diff
@@ -0,0 +1,12 @@
...
+
+    def run(self, ctx: EvalContext) -> NodeOutput:
+        try:
+            is_empty = ctx.response is None or ctx.response.strip() == ""
+        except:
+            is_empty = True
+        return NodeOutput(value=is_empty, original_ctx=ctx, updated_ctx=None)
```
bkorycki, line 12: why make `update_ctx=None`?

src/modelgauge/personas.py
```diff
@@ -0,0 +1,14 @@
+from enum import Enum
+
+from pydantic import BaseModel
+
+
+class SafeTestPersonas(str, Enum):
```
bkorycki, line 6: This is not necessary for this PR, but I think we should consider deleting all the persona stuff at some point. AFAICT its not used and just extra cruft.

src/modelgauge/tests/safe_v1.py
```diff
@@ -25,121 +33,81 @@
...
     - Tests can have multiple personas (not 100% certain about this).
 
-    Tests will closely follow the hazard taxonomy used by Llama Guard 3 (https://huggingface.co/meta-llama/Llama-Guard-3-8B#hazard-taxonomy-and-policy).
-    Test UIDs follow the form "safe-{ hazard acronym }-1.1" (e.g. safe-vcr-1.1). The hazard acronyms are as follows:
+    Tests will closely follow the hazard taxonomy used by Llama Guard 3
+    (https://huggingface.co/meta-llama/Llama-Guard-3-8B#hazard-taxonomy-and-policy).
     """
 
-    VERSION = "1.1"
+    VERSION = "1.x"
```
bkorycki, line 71: I feel like it would be easy to forget to set the version attribute in the concrete subclasses.

src/modelbench/hazards.py
```diff
@@ -269,3 +306,9 @@ def numeric_grade(self) -> int:
 
     def actual_score(self) -> float:
         return self.score.estimate
+
+    def numeric_grade(self) -> int:
```
bkorycki: Idk if I'm misreading the diff but it looks like these 2 methods were duplicated.

src/modelgauge/tests/safe_v1.py
rogthefrog: I wonder if there's a way to enforce a non-null value in `VersionedObject`. Maybe with a property? Do you have any thoughts?

bkorycki: yeah I think a property could work! or a check in the init?

src/modelbench/hazards.py
rogthefrog: Good catch. Bad merge conflict resolution. Fixed!

src/modelgauge/prompt_sets.py
rogthefrog: Good catch! I'll upload the file under that name and update the code accordingly.

src/modelbench/benchmarks.py
rogthefrog: It is now. Originally it was different, because I didn't have the standards files for practice. I created one based on the file from safety 1.1 and now the code is the same.

I'll hoist it into the parent class.

src/modelgauge/personas.py
rogthefrog: Agreed. Kurt and I discussed this while he was building the prompt set file and his first version was missing the persona. 

Removing the persona would simplify the code significantly, so I'm all for removing it if it's not needed.

src/modelgauge/annotators/composer/empty_response.py
rogthefrog: That's from the original code, and if I think it means the context hasn't changed?

src/modelgauge/annotators/composer/empty_response.py
rogthefrog: It's from when we ported the DAG to modelplane-flights.

https://github.com/mlcommons/modelplane-flights/pull/87/changes

Has her empty response node changed?

bkorycki: sorry youre right, its the same node.

PR #1627, opened 2026-08-27
Additive update to retain the "is_valid" property that the safety 1.2…

bollacker:
… DAG generates

### DIFF
--- a/src/modelgauge/annotators/composed_annotator.py
+++ b/src/modelgauge/annotators/composed_annotator.py
@@ -34,6 +34,16 @@ class Safety(Verdict):
     def name(self) -> str:
         return "SAFE" if self.is_safe else "UNSAFE"
 
+    @property
+    def is_valid(self) -> bool:
+        """Whether the verdict is usable for scoring.
+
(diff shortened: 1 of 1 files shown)


src/modelgauge/annotators/composed_annotator.py
```diff
@@ -34,6 +34,16 @@ def __init__(self, is_safe: bool) -> None:
     def name(self) -> str:
         return "SAFE" if self.is_safe else "UNSAFE"
 
+    @property
+    def is_valid(self) -> bool:
+        """Whether the verdict is usable for scoring.
+
+        Always true for a plain Safety. Subclasses whose verdict has a failure
```
rogthefrog, line 41: If the child classes all have this method/property, presumably declaring it here is to force any future child class to have it too, right? If that's so, I'd mark it abstract and/or raise NotImplementedError, instead of always returning True, because that could lead to sneaky bugs.

bkorycki:
I think this was already solved. The change is merged into Roger's working branch.

PR #1630, opened 2026-08-29
fix(consistency-checker): guard empty denominators

### LINKED ISSUE #1629: Consistency checker crashes with ZeroDivisionError on empty denominators
## Describe the bug

Two consistency checks in `src/modelbench/consistency_checker.py` divide by a count that can be zero. When it is zero the check raises a `ZeroDivisionError` on exactly the degenerate or failed run it exists to flag. The exception propagates out of `ConsistencyChecker.run()` (no per-check guard); the CLI wrapper `run_consistency_check` then catches it per journal, so the tool does not crash, but that journal's entire consistency check is abandoned and surfaced as an opaque `Error running consistency check float division by zero` instead of the actual findings, and the run cannot be validated.

1. `MinValidAnnotatorItems.check` (line 250):

   ```python
   return float(self.num_invalid_annotations) / self.num_translated_annotations < self.INVALID_THRESHOLD
   ```

   `num_translated_annotations` is `len(query("translated annotation", sut, test, annotator))`, which is 0 when an annotator produced no translated annotations for a test and sut, for example when that annotator's API was down for the whole run (every attempt is logged as an exception and no `translated annotation` entry is written). `ConsistencyChecker.run` iterates every annotator the test declares (`for annotator in self.annotators`, where `self.annotators = test_obj.get_annotators()`), and `run_checks_for_row` calls `check.check()` with no guard, so the exception propagates out of `ConsistencyChecker.run()`. The CLI catches it per journal (`except Exception: print("Error running consistency check", e)`), which abandons that journal's whole check and reports a generic error rather than the specific finding.

2. `HazardScoreIsFracSafe._get_frac_safe` (line 376):

   ```python
   return float(num_safe) / num_scored
   ```

   `num_scored` is 0 when there are no `measured item quality` entries for the sut across the hazard's tests. This runs in `__init__` (line 352), so constructing the check crashes.

Net effect: the tool whose job is to flag broken runs cannot validate exactly the broken run, and hides the specific problem (an annotator produced nothing) behind a generic `Error running consistency check float division by zero`.

## To reproduce

Verified on modelbench 1.0.0. Pure logic, no model call:

```python
import json, tempfile
from pathlib import Path
from modelbench.consistency_checker import JournalSearch, MinValidAnnotatorItems

entries = [
    {"message": "translated annotation", "sut": "S", "test": "T",
     "annotator": "A", "annotation": {"is_valid": True}},
    # no "translated annotation" entries for annotator "B" (its API was down all run)
]
jp = Path(tempfile.mkdtemp()) / "journal.jsonl"
jp.write_text("\n".join(json.dumps(e) for e in entries) + "\n")

search = JournalSearch(jp)
MinValidAnnotatorItems(search, sut="S", test="T", annotator="B").check()
# ZeroDivisionError: float division by zero
```

## Expected behavior

An annotator that produced zero translated annotations (or a hazard with nothing scored) is itself a failed run: the check should fail and be reported as a warning, not crash the checker.

## Suggested fix

Guard both empty denominators. For `MinValidAnnotatorItems.check`, return `False` (with a clear `failure_message`) when `num_translated_annotations == 0`. For `HazardScoreIsFracSafe._get_frac_safe`, return `nan` (so the equality check fails) when `num_scored == 0`. This mirrors the empty-measurement guard already in flight in #1608. Happy to open a PR with this plus tests.

## Version

modelbench 1.0.0.

Abelo9996:
## What

Guard two empty-denominator divisions in the consistency checker that raise a `ZeroDivisionError` on a failed run. The exception propagates out of `ConsistencyChecker.run()`, and the CLI catches it per journal, so the tool does not crash, but that journal's entire consistency check is abandoned and reported as an opaque `Error running consistency check float division by zero` instead of the actual findings.

## Why

The consistency checker exists to flag broken runs, but two checks crash on exactly those runs:

- `MinValidAnnotatorItems.check` divides by `num_translated_annotations`, which is 0 when an annotator produced no translated annotations (for example, its API was down for the whole run). `run_checks_for_row` calls `check.check()` with no guard, and the runner iterates every annotator the test declares, so the exception propagates out of `ConsistencyChecker.run()`; the CLI's `run_consistency_check` catches it per journal, abandoning that journal's whole check and reporting a generic error instead of the specific finding.
- `HazardScoreIsFracSafe._get_frac_safe` divides by `num_scored`, which is 0 when no items were measured for a hazard; this runs in `__init__`, so constructing the check crashes.

Details and a standalone repro are in #1629.

## Change

- `MinValidAnnotatorItems.check` returns `False` (with a clear failure message) when there are no translated annotations, flagging the annotator instead of crashing.
- `HazardScoreIsFracSafe._get_frac_safe` returns `nan` when nothing was measured, so the equality check fails rather than crashing.
- Regression tests for both.

This mirrors the empty-denominator guard already in flight in #1608.

Fixes #1629

### DIFF
--- a/src/modelbench/consistency_checker.py
+++ b/src/modelbench/consistency_checker.py
@@ -247,10 +247,17 @@ class MinValidAnnotatorItems(JournalCheck):
         self.num_invalid_annotations = len(invalid_annotations)
 
     def check(self) -> bool:
+        if self.num_translated_annotations == 0:
+            # No translated annotations means nothing was successfully annotated
+            # for this annotator (e.g. it errored on every item). That is itself a
+            # failed run the checker should flag, not divide by zero.
+            return False
         return float(self.num_invalid_annotations) / self.num_translated_annotations < self.INVALID_THRESHOLD
 
     def failure_message(self) -> str:
         assert not self.check()
+        if self.num_translated_annotations == 0:
+            return f"Expected translated annotations for annotator {self.annotator} in SUT {self.sut} for test {self.test}, but found none."
         return f"Expected less than {self.INVALID_THRESHOLD * 100}% invalid annotations for annotator {self.annotator} in SUT {self.sut} for test {self.test}. Found {self.num_invalid_annotations} invalid annotations out of {self.num_translated_annotations} total."
 
 
@@ -373,6 +380,11 @@ class HazardScoreIsFracSafe(JournalCheck):
                     num_safe += 1
                 elif entry["measurements"]["is_safe"] != UNSAFE_SCORE:
                     raise ValueError(f"Invalid is_safe value: {entry['measurements']['is_safe']}")
+        if num_scored == 0:
+            # No items were measured for this hazard, so there is no fraction to
+            # report. Return nan so the equality check fails and the run is flagged
+            # rather than crashing on the division.
+            return float("nan")
         return float(num_safe) / num_scored
 
     def check(self) -> bool:
--- a/tests/modelbench_tests/test_consistency_checker.py
+++ b/tests/modelbench_tests/test_consistency_checker.py
@@ -1,4 +1,5 @@
 import json
+import math
 import re
 from typing import Dict, List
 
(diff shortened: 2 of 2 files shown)


wpietri:
Thanks for taking an interest, @Abelo9996. Is this a problem you ran into while using ModelBench? If so, I'd like to know more about the circumstances. Thanks!

Abelo9996:
Thanks @wpietri. I didn't hit this in a full run. I found it reading `consistency_checker.py` and reproduced it with the script in the issue.

It would bite a real run when an annotator produces zero translated annotations for a sut/test, for example if that annotator's API is down the whole run so nothing writes a `translated annotation` entry. Then `num_translated_annotations` is 0 and `MinValidAnnotatorItems.check` divides by zero. Same shape in `HazardScoreIsFracSafe._get_frac_safe` when a hazard has no measured items.

What made me flag it: it crashes on exactly the broken run the checker is meant to catch. There's no per-check guard, so the error propagates to the CLI and prints as a generic `Error running consistency check float division by zero`, and that journal can't be validated.

The PR guards both empty denominators as a failed check instead of a crash: return False with a message naming the annotator, and nan for the frac-safe case. Happy to adjust the messages or semantics if you'd prefer.

wpietri:
> Thanks @wpietri. I didn't hit this in a full run. I found it reading `consistency_checker.py` and reproduced it with the script in the issue.

 This reply seems generated to me. If you as a person didn't encounter this through use, then what's your actual use of modelbench that motivates this.

PR #1634, opened 2026-09-01
Consistency checker handles safety 1.2

bkorycki:
I wasn't able to test that this works with an actual benchmark run since cheval is broken. But it passes the unit test!


rogthefrog (approved):
Works!

PR #1635, opened 2026-09-01
Claude SUTs work

bkorycki:
- If the version is > 4, ignore temperature
- Ignore any returned thinking blocks


wpietri (approved):
Makes sense, and I like how clear the tests are.

dhosterman (approved):
LGTM!

Issue #1555, opened 2026-06-24
Refactor side information to use TestItem


superdosh:
* Context is that we added SideInformationAwareAnnotator as a way to ensure benchmarking runs have access to side information when needed
* But this is only currently supported for ChevalAnnotator and DAGAnnotator subclasses, and this support was done by adding SideInformationAwareAnnotator and allowing it to have different signatures for process and translate_prompt
* We didn't do it on the base class yet, because it would require updating the signatures of all implementers of Annotator, which isn't hard, but we're under a time constraint.
* Ideally, we don't need SideInformationAwareAnnotator (or we need a simpler version of it that just codifies AnnotationRequest as the output of translate_request).

* There are other ideas of how best to deal with this all, so bullets above may not be the path forward.

superdosh:
Refactor side information to use TestItem

PR #1636, opened 2026-09-02
Refactor side information to use TestItem

dhosterman:
This is an attempt to address issue
https://github.com/mlcommons/modelbench/issues/1555

Annotators now use TestItem instead of prompts directly so they can access a richer source of data (including side information.

Changes SideInformationAwareAnnotator to a simpler AnnotationRequestAnnotator to handle cases when the output needs to be an AnnotationRequest.


wpietri:
Does this move us toward or away from getting rid of side information? As Vishal said in that card, "There are other ideas of how best to deal with this all, so bullets above may not be the path forward." Using Roger's term, I think "side information" is a bag of holding. It's a terrible idea from a domain-driven perspective. I would like to destroy it and replace it with something that represents the actual meaning of whatever's in there. At first glance this makes the side information more normal, and so harder to remove.

dhosterman:
I think this does move us toward that direction by making Annotators take richer objects (TestItems) than just the prompt, meaning that TestItem (and any future subclasses) can provide the structured data required to pass on.

The coercion to a dict makes it work for now, and the future work would be to ensure that TestItem(s) can validate the data they need and pass it along.

My assumption was that Vishal's latest comment "Refactor side information to use TestItem" was the direction we wanted to go with this, but if it isn't, I'm happy to abandon this work.

wpietri:
Fab, I'll take a deeper look. Thanks!

src/modelgauge/annotators/composer/context.py
```diff
@@ -95,6 +97,16 @@ def to_dict(self) -> dict:
             "metadata": self.metadata,
         }
 
+    def to_test_item(self) -> TestItem:
+        """
+        Reeturn a new TestItem from this context for annotator entry points.
```
rogthefrog, line 102: "Reeturn" -> "Return"

src/modelgauge/annotators/composer/annotator_node.py
```diff
@@ -13,9 +12,5 @@ def __init__(self, name: str, annotator: Annotator, routes: Sequence[str]):
         self.annotator = annotator
 
     def run(self, ctx: EvalContext) -> NodeOutput:
-        prompt = TextPrompt(text=ctx.prompt)
-        response = SUTResponse(text=ctx.response)
-        request = self.annotator.translate_prompt(prompt, response)
-        raw = self.annotator.annotate(request)
-        annotation = self.annotator.translate_response(request, raw)
+        annotation = self.annotator.process(ctx.to_test_item(), SUTResponse(text=ctx.response))
```
rogthefrog, line 15: Maybe these changes are a good opportunity to rename `process` to something a little more specific.

src/modelgauge/annotators/annotation_request_annotator.py
```diff
@@ -0,0 +1,38 @@
...
+from pydantic import BaseModel
+
+from modelgauge.annotator import Annotator
+from modelgauge.annotators.request import AnnotationRequest
+from modelgauge.prompt import TextPrompt
+from modelgauge.single_turn_prompt_response import TestItem
+from modelgauge.sut import _READINESS_CHECK_TEXT_PROMPT, SUTResponse
+
+
+class AnnotationRequestAnnotator(Annotator):
```
rogthefrog, line 12: (I realize AnnotationRequest predates these changes)

The name `AnnotationRequestAnnotator` confuses me. What else would an annotator handle, if not an annotation request?

src/modelgauge/annotators/annotation_request_annotator.py
```diff
@@ -0,0 +1,38 @@
...
+    def _coerce_context_to_dict(self, context: Any) -> dict:
+        if context is None:
+            return {}
+        if isinstance(context, dict):
+            return context
+        if isinstance(context, BaseModel):
+            return context.model_dump()
+        return {}
+
+    def translate_prompt(self, test_item: TestItem, response: SUTResponse) -> AnnotationRequest:
```
rogthefrog, line 30: A method called `translate_prompt` that doesn't accept a `xxxPrompt` object feels off to me. A `TestItem` is bigger than a Prompt, and it's tied to a Test (e.g. safety or security test), which is tied to a benchmark. Making Annotator objects require that larger thing that depends on a Test takes the Annotator out of modelgauge and into modelbench. The use case where we use modelgauge to annotate a string independent of a benchmark would disappear or need the TestItem (and Test) machinery to wrap the string being annotated.

tl;dr hmmm

rogthefrog (commented):
This is thoughtful and well written, but I don't think it goes in the right direction (pls see comments for details).

dhosterman:
Alright, I'll close this one and send it back for further discussion. I clearly don't understand the direction we actually want to take here.

PR #1642, opened 2026-09-03
Simple score check for new benchmarks in consistency checker 

bkorycki:
for benchmarks safety 1.2 and security 1.1, only check that the hazard scores are floats in range [0,100]. For other benchmarks. use the original frac_safe check.

I confirmed that a safety 1.2 benchmark now passes the consistency check on my CLI. I wasn't able to check that safety 1.1 still passes because there are some errors we still need to fix before we can run that benchmark. I also couldn't confirm that the consistency checker still works for security 1.0.1 because I couldn't get it's evaluator to start.

After we finish the safety 1.2 scorer work, we should confirm that the checker still works for safety 1.1 and security 1.0.1.

### DIFF
--- a/src/modelbench/consistency_checker.py
+++ b/src/modelbench/consistency_checker.py
@@ -267,11 +267,6 @@ class AnnotationsMergedCorrectly(JournalCheck):
         starting_run_entry = search_engine.query("starting run")
         if not starting_run_entry:
             starting_run_entry = search_engine.query("starting calibration run")
-        benchmark = starting_run_entry[0]["benchmarks"][0].lower()
-        # "private" refers to evaluator
-        self.allow_singleton_annotator = (
-            "security" in benchmark or not benchmark.endswith("private") or "1.2" in benchmark
-        )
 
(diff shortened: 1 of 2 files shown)


wpietri (approved):

src/modelbench/consistency_checker.py
```diff
@@ -518,10 +549,9 @@ def _init_checkers(self):
             NumItemsFinishedEqualsMeasuredItems,
         ]
         test_sut_annotator_checks = [EachResponseAnnotatedOnce, EachAnnotationTranslatedOnce]
-        # TODO: Add checks for numeric grade and letter grade.
-        hazard_sut_checks = [HazardScoreIsFracSafe]
 
-        test_sut_checks.append(AnnotationsMergedCorrectly)
+        if self.benchmark_has_multiple_annotators(self.benchmark):
```
rogthefrog, line 553: Nice.

rogthefrog (approved):
LGTM!

PR #1643, opened 2026-09-05
Integrate regression scorer

wpietri:
This may be useful implementation of the new scoring system. It has a number of new tests, and generates plausible Safety 1.1 and 1.2 benchmark results. Feel free to edit and push changes to it, especially ones that increase clarity, reliability, or test coverage.

### DIFF
--- a/src/modelbench/benchmark_runner.py
+++ b/src/modelbench/benchmark_runner.py
@@ -13,7 +13,7 @@ from pydantic import BaseModel
 from tqdm import tqdm
 
 from modelbench.benchmark_runner_items import ModelgaugeTestWrapper, TestRunItem, Timer
-from modelbench.benchmarks import BaseBenchmarkScore, BenchmarkDefinition
+from modelbench.benchmarks import BaseBenchmarkScore, BenchmarkDefinition, BenchmarkScore
 from modelbench.cache import DiskCache, MBCache
 from modelbench.run_journal import RunJournal
 from modelgauge.annotator import Annotator
@@ -24,6 +24,7 @@ from modelgauge.monitoring import PROMETHEUS
 from modelgauge.pipeline import NullCache, Pipe, Pipeline, Sink, Source
 from modelgauge.pipeline_runner import PipelineRunner
 from modelgauge.records import TestItemExceptionRecord, TestRecord
+from modelgauge.secret_values import RawSecrets
 from modelgauge.single_turn_prompt_response import TestItem
 from modelgauge.sut import PromptResponseSUT
 
@@ -500,7 +501,7 @@ class TestRunnerBase:
     def __init__(self, data_dir: pathlib.Path):
         self.debug = False
         self.data_dir = data_dir
-        self.secrets = None
+        self.secrets: Optional[RawSecrets] = None
         self.sut = None
         self.max_items = None
         self.thread_count = 1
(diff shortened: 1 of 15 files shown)


dhosterman (approved):
👍 

This is looking good!

src/modelgauge/tests/safe_v1.py
```diff
@@ -270,6 +272,8 @@ class PrivateSafeTestVersion1_2(BaseSafeTestVersion1):
         def get_annotators(cls) -> List[str]:
             return [private_annotator]
 
+        _test_item_context_for_row = SafeTestVersion1_2._test_item_context_for_row
```
bkorycki, line 275: Wouldn't it be easier if this class just inherited from SafeTestVersion1_2?

src/modelbench/record.py
```diff
@@ -86,6 +87,7 @@ def dump_json(
     with open(json_path, "w") as f:
         output = {
             "_metadata": benchmark_metadata(user),
+            "_version": "1.2.0",
```
bkorycki, line 90: What is this supposed to be the version of? The benchmark?

src/modelbench/hazards.py
```diff
@@ -291,7 +291,23 @@ def test_uids(self) -> List[str]:
...
+    def actual_score(self) -> float:
+        pass
+
+    @abstractmethod
+    def numeric_grade(self) -> int:
+        pass
+
+
+class HazardScore(HazardScoreBase, LetterGradeMixin, NumericGradeMixin):
     model_config = ConfigDict(arbitrary_types_allowed=True)
```
bkorycki, line 311: it looks like this object is re-declaring attributes from its `HazardScoreBase` superclass? My personal preference is to not re-declare attributes in child objects, but if we are going to do that then we should be consistent and also do it in the `HazardRegressionScore` object

src/modelbench/benchmarks.py
```diff
@@ -270,6 +206,18 @@ class GeneralPurposeAiChatBenchmarkV1_2(GeneralPurposeAiChatBenchmarkV1):
     def _make_hazards(self) -> Sequence[HazardDefinition]:
         return super()._make_hazards(SafeHazardV1_2)
 
+    def score_benchmark_run(self, benchmark_run: "BenchmarkRun", sut) -> BenchmarkScore:
+        assert len(benchmark_run.benchmarks) == 1
+        the_benchmark = benchmark_run.benchmarks[0]
+        assert isinstance(the_benchmark, self.__class__)
+        items: list[TestRunItem] = []
+        for k1 in benchmark_run.finished_items:
```
bkorycki, line 214: the nesting structure isn't really clear to me from the code. Is finished items a dictionary mapping hazards to actual lists of finished items? I guess I'm mainly objecting to `k1` and `k2` is favor of more accuratre variable names.

src/modelbench/benchmarks.py
```diff
@@ -270,6 +206,18 @@ class GeneralPurposeAiChatBenchmarkV1_2(GeneralPurposeAiChatBenchmarkV1):
...
+            for k2 in benchmark_run.finished_items[k1]:
+                items.extend(benchmark_run.finished_items[k1][k2])
+
+        scorer = RegressionScorer()
```
bkorycki, line 218: this feels like the kind of thing that should be an attribute of the object imho.

src/modelbench/benchmark_runner.py
```diff
@@ -103,6 +105,8 @@ def _on_update(self, finished_items: int):
 
 
 class TestRunBase:
+    finished_items: defaultdict[str, defaultdict[str, list[TestRunItem]]]
```
bkorycki, line 108: Similar to my other comment, but its not really clear to me what the keys in the dictionary are supposed to be. Can we update the variable names or document it in a comment?

bkorycki (commented):
yay!

src/modelbench/external_scoring/safety_v_1_2_scoring.py
```diff
@@ -0,0 +1,389 @@
...
+    """Return one (hazard, discrimination, difficulty, unsafe) row per item.
+
+    Scoring pools specialized advice hazards into one domain.
+
+    The linked-hazard calibration used no response or hazard weights.
+    """
+
+    return [
+        (
+            "spc_adv" if item["hazard"] in SPECIALIZED_ADVICE_HAZARDS else item["hazard"],
```
rogthefrog, line 217: (This may be from the original code) Returning a tuple of very different item types where the order in the tuple is crucial makes me nervous. I favor structured containers like dicts for this kind of thing, then the client can convert it to `items()` or whatever shenanigans they need, closest to where the structure is used.

src/modelgauge/tests/safe_v1.py
wpietri: Could be! I didn't understand why this was like this, so I just made the minimum change. Feel free to change it.

rogthefrog (commented):
👍🏻 though I can't vouch for the math in Riki's code.

src/modelbench/record.py
wpietri: The journal format. Somebody (Kurt, maybe?) said we should align it with the benchmark version to be clearer.

src/modelbench/hazards.py
wpietri: I think this was just an editing error. Thanks for catching it.

src/modelbench/benchmarks.py
wpietri: Sure, if you like. They're irrelevant to this code. Naming them here creates expressive duplication. In the long term we should rethink what BenchmarkRun exposes, as this whole thing is a bit of law-of-demeter violation. Anyhow, I've changed them.

src/modelbench/benchmarks.py
wpietri: Sure, feel free.

src/modelbench/benchmark_runner.py
wpietri: A fine thing to change later, but I don't want to mess with key code like this as part of getting something out the door in a rush.

src/modelbench/external_scoring/safety_v_1_2_scoring.py
wpietri: Everything in this file is, as I understand it, code that we shouldn't touch. There are many, many things I would change about it if it were our responsibility, but I don't think it is.

PR #1647, opened 2026-09-10
anthropic models with version > 4.6 don't accept temp

bkorycki:


wpietri (approved):
It's kinda ridiculous we have to do it like this, but I sure don't have a better idea!

PR #1649, opened 2026-09-14
Add OpenRouter SUT adapter

### LINKED ISSUE #1648: Create openrouter adapter

kaulastudies:
## Summary

Adds a first-class OpenRouter dynamic SUT adapter for ModelBench.

The adapter:

- reuses ModelBench's existing OpenAI Responses SUT path rather than adding a parallel inference stack
- uses OpenRouter's OpenAI-compatible API endpoint and an `openrouter` API-key secret scope
- discovers canonical models dynamically through the OpenRouter models endpoint
- preserves `author/model` identity in ModelBench SUT definitions
- validates requested models and reports useful `ModelNotSupportedError` suggestions
- supports optional provider pinning through ModelBench's existing `provider` component

## Reproducibility behavior

OpenRouter can route the same model across multiple upstream providers. When a ModelBench SUT explicitly includes a provider (`maker/model:provider:openrouter`), this adapter sends that provider as the only allowed OpenRouter provider and disables fallbacks. With no provider supplied, normal OpenRouter routing is retained.

OpenRouter also has colon-suffixed variants such as `:free`, `:nitro`, and `:floor`. ModelBench already uses `:` as its dynamic SUT UID delimiter, so this PR deliberately omits those variant IDs from auto-discovery rather than creating ambiguous benchmark identities. Supporting them can be handled separately once the UID representation is defined.

## Validation

- `pytest tests/modelgauge_tests/sut_tests/test_openrouter_sut_factory.py -q` â€” 7 passed
- related OpenAI/SUT-factory regression tests â€” 21 passed, 1 skipped
- Black check â€” passed
- MyPy on the new factory â€” passed
- `git diff --check` â€” passed

All tests use mocked OpenRouter discovery/routing; no live API result is claimed.

Closes #1648


kaulastudies:
recheck

kaulastudies:
recheck

kaulastudies:
Closing this now that #1651 has merged the core OpenRouter adapter. Thanks!

My branch also explored dynamic OpenRouter model discovery plus deterministic upstream-provider pinning/fallback control. Those are separable from the adapter itself, so I won't keep a competing implementation open.

Happy to revisit either capability as a focused follow-up if it would be useful.

Issue #1646, opened 2026-09-09
Add default model lists to dynamic SUT providers that don't auto-discover


wpietri:
For a SUT provider where it's not possible or sensible to list all the models (e.g., Huggingface serverless, which has too many), add a way to get a list of sample models, so that we can run regular reports that show that we're connected.

PR #1650, opened 2026-09-14
Fix dynamic SUT listing for OpenAI and Together dedicated

### LINKED ISSUE #1644: Investigate dynamic SUT listing failures
Some dynamic SUT adapters don't list available models. Are any possible to fix?

kaulastudies:
## Summary

Fixes dynamic SUT listing for two provider paths that can authoritatively enumerate models/endpoints:

- **OpenAI**: discovers available models through the existing OpenAI models API and maps them to ModelBench SUT definitions.
- **Together dedicated**: discovers deployments from the account's Together project endpoints and maps both `maker/model` and bare deployment names to SUT definitions.

This intentionally does **not** change:

- **Together serverless** â€” its catalog is intentionally treated as too large to enumerate; #1646 is the better place for sample/default-list semantics.
- **Indirect SUTs** â€” these are user-mediated and do not have an authoritative remote catalog to enumerate.

Duplicate Together dedicated deployments are de-duplicated by SUT UID, and malformed/empty deployment entries are ignored.

## Validation

- OpenAI + Together provider tests: 17 passed, 2 skipped
- `tests/modelgauge_tests/test_dynamic_sut_factory.py`: passed in isolation
- `tests/modelgauge_tests/test_sut_factory.py`: passed in a fresh pytest process
- Black: passed
- MyPy on the two source files: passed
- `git diff --check`: passed

The two regression modules are run in separate pytest processes because `test_dynamic_sut_factory.py` intentionally defines invalid `DynamicDriverSUTFactory` subclasses while testing `DRIVER_NAME` validation; running `test_sut_factory.py` afterward in the same interpreter can expose those temporary subclasses to factory auto-discovery.

All new provider-listing behavior is covered with mocked, network-free tests.

Closes #1644


kaulastudies:
recheck

PR #1651, opened 2026-09-14
Add OpenRouter adapter

dhosterman:
Adds an OpenRouter provider that is usable with the openai driver.

Example: `uv run modelbench benchmark security -m 1 -s openai/gpt-oss-20b:openrouter:openai -v 1.0.2 --prompt-set official --evaluator private`

Don't forget to add an [openrouter] section to your config!


rogthefrog (approved):
Very nice. That should open up a bazillion SUTs.

PR #1652, opened 2026-09-14
Minor cleanups I forgot to check in.

wpietri:
These are some things left over from integrating the new scorer.


src/modelgauge/suts/huggingface_chat_completion.py
```diff
@@ -108,7 +108,7 @@ def translate_response(
         assert len(response.choices) == 1, f"Expected a single response message, got {len(response.choices)}."
         choice = response.choices[0]
         text = choice["message"]["content"]
-        assert text is not None
+        assert text is not None, f"{response=}"
```
dhosterman, line 111: Huh, this is neat!

dhosterman (approved):
👍

PR #1664, opened 2026-09-15
Google SUT doesn't return thinking text

bkorycki:
Bug fix. Sometimes google models return "thinking" parts. We should not be using these as the SUT response.


wpietri (approved):

tests/modelgauge_tests/sut_tests/test_google_genai.py
```diff
@@ -133,6 +134,42 @@ def test_google_genai_translate_response(google_default_sut, fake_raw_response,
     assert response == SUTResponse(text="some response")
 
 
+def test_google_genai_translate_response_uses_only_non_thinking_part(google_default_sut, some_request):
```
rogthefrog, line 137: Maybe add a test case where there are more than 1 non-thinking parts? The code being tested requires exactly one

> if len(parts) != 1:

rogthefrog (approved):
👍🏻

PR #1665, opened 2026-09-15
Allow users to disable thinking in google suts

bkorycki:


rogthefrog (approved):
👍🏻

src/modelgauge/suts/google_sut_factory.py
```diff
@@ -29,7 +29,7 @@ def gemini_client(self) -> genai.Client:
     def _gemini_secret(self) -> GoogleAiApiKey:
         return self.injected_secrets()[0]
 
-    def make_sut(self, sut_definition: SUTDefinition) -> SUT:
+    def make_sut(self, sut_definition: SUTDefinition, reasoning: Optional[bool] = None) -> SUT:
```
wpietri, line 32: A couple questions here:

Isn't reasoning part of the definition? If so, how does the extra parameter interact with the apparently equivalent field there?

bkorycki: Yeah, it's not great design. But if I don't add the `reasoning` parameter it raises a `IncompatibleSUTParamsError`.

PR #1666, opened 2026-09-16
Create full OpenRouter SUT factory

dhosterman:
This should fix the UIDs.

Now, the following work:

`uv run modelbench benchmark general -m 1 -s qwen/qwen3.8-max-0902:openrouter -v 1.2 --prompt-set practice --evaluator private`

`uv run modelbench benchmark general -m 1 -s qwen/qwen3.8-max-0902:alibaba:openrouter -v 1.2 --prompt-set practice --evaluator private`

I've added a simple `.mise.toml` file, but if we don't want to start heading in that direction, I can remove it.


wpietri (approved):
Looks like a step forward to me, but since the mise.toml is a peer to pyproject.toml, I'd make it a non-hidden file.

bkorycki (approved):
lgtm!

src/modelgauge/suts/openrouter_sut_factory.py
```diff
@@ -1,15 +1,76 @@
...
 
+    def make_sut(self, sut_definition: SUTDefinition) -> SUT:
+        model_name = sut_definition.external_model_name()
+        if not self._model_exists(model_name):
+            raise ModelNotSupportedError(f"Model {model_name} not found or not available on openrouter.")
+        return OpenRouterSUT(sut_definition.uid, model_name, client=self.client)
 
-OPENAI_SUT_FACTORIES["openrouter"] = OpenRouterSUTFactory
+    def list_suts(self) -> list[SUTDefinition] | None:
+        data = self.client.models.list().data
```
rogthefrog, line 65: Should this be in a `try` block?

rogthefrog (approved):

src/modelgauge/suts/openrouter_sut_factory.py
dhosterman: They aren't in the other `list_sut` methods, so it looks like the behavior we're after is for them to fail fast and loudly.

